PR #452, opened 2015-11-12
Dev exxon

weijjia:
1. support client tls for authenticating user
2. remove android:theme in manifest, as a library, we shouldn't provide it, this could potentially cause calling app building errors.

### DIFF
--- a/src/src/com/microsoft/aad/adal/AuthenticationActivity.java
+++ b/src/src/com/microsoft/aad/adal/AuthenticationActivity.java
@@ -28,8 +28,10 @@ import java.security.InvalidKeyException;
 import java.security.KeyStoreException;
 import java.security.NoSuchAlgorithmException;
 import java.security.NoSuchProviderException;
+import java.security.PrivateKey;
 import java.security.UnrecoverableEntryException;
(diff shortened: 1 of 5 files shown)


src/src/com/microsoft/aad/adal/AuthenticationActivity.java
```diff
@@ -642,6 +650,29 @@ public void setPKeyAuthStatus(boolean status) {
...
+                    try {
+                    	final X509Certificate[] certChain = KeyChain.getCertificateChain(
+                                getApplicationContext(), alias);
+                    	final PrivateKey privateKey = KeyChain.getPrivateKey(mCallingContext, alias);
+                        request.proceed(privateKey, certChain);
+                        
+                    } catch (KeyChainException e) {
+                        Log.e(TAG, "KeyChain exception", e);
+                    } catch (InterruptedException e) {
+                        Log.e(TAG, "KeyChain exception", e);
```
sdabbiru: Should be Interrupted exception instead of KeyChain exception?

sdabbiru:
Apart from above comment, I sign off.

PR #470, opened 2016-01-11
Feature: redirectUri checking

heidijinxujia:
Question: do we need to check if the input redirectUri is in the registered list?


src/src/com/microsoft/aad/adal/AuthenticationContext.java
```diff
@@ -1153,6 +1153,16 @@ private AuthenticationResult acquireTokenAfterValidation(CallbackHandler callbac
             AuthenticationResult result = null;
             request.setVersion(getVersionName());
             request.setBrokerAccountName(request.getLoginHint());
+            
+            //check the redirectUri before sending the request
+            //if the redirectUri from the client does not match the valid redirectUri
+            // the client app would not jump to the login page
+            if(!request.getRedirectUri().equals(this.getRedirectUriForBroker())){
+             String msg = "the redirectUri is invalid.";
```
weijjia: I expect the message to be a little bit more concrete. Can we check the following: 1) if it's starting with msauth:// 2) is the package name as expected 3) if the signature as expected? And then we can give more detailed messsage, telling which part of the redirect uri is wrong.

heidijinxujia: Detailed error messages added. Please have a look.

src/src/com/microsoft/aad/adal/AuthenticationContext.java
```diff
@@ -1138,6 +1138,36 @@ private AuthenticationResult acquireTokenLocalCall(final CallbackHandler callbac
...
+    //redirectUri format %PREFIX://%PACKAGE_NAME/%SIGNATURE   
+    private boolean checkRedictUri(CallbackHandler callbackHandle, final AuthenticationRequest request) throws UnsupportedEncodingException{
+    	
+    	String errMsg = new String();    	
+    	String inputUri = request.getRedirectUri();
+    	PackageHelper helper = new PackageHelper(mContext);
+    	
+    	if(inputUri == null){
+    		errMsg = "the redirectUri is null";    		
+    	}else if(!inputUri.split("://")[0].equals(AuthenticationConstants.Broker.REDIRECT_PREFIX)){
```
weijjia: You can use startwith("msauth://"), this looks simpler :)

src/src/com/microsoft/aad/adal/AuthenticationContext.java
```diff
@@ -1138,6 +1138,36 @@ private AuthenticationResult acquireTokenLocalCall(final CallbackHandler callbac
...
+    		errMsg = "the redirectUri is not starting with the valid redict prefix " + AuthenticationConstants.Broker.REDIRECT_PREFIX;    		
+    	}else if((inputUri.split("://").length > 1 && !inputUri.split("://")[1].split("/")[0].equals(URLEncoder.encode(mContext.getPackageName(), AuthenticationConstants.ENCODING_UTF8))) || inputUri.split("://").length == 1 ){
```
weijjia: i guess we can simply check if the given redirect uri contains the expected package name and signature. Can you also log out the expected package name and signature when it's wrong.

heidijinxujia: And "msauth" is defined in AuthenticationConstants.Broker.REDIRECT_PREFIX.  Will it be better to avoid the usage of hard coding parameters?

weijjia:
This is failing on travis, we need to update the android sdk version on .travis.yml. 
Also, could you add unit tests for the redirect uri check?

heidijinxujia:
I've edited the .travis.yml file. The travis still got failed. Any suggestions?

weijjia:
From the console output, it fails at the emulator startup.

heidijinxujia:
I've edited the android file in .travis. Still failed. Any suggestions? Thanks!

heidijinxujia:
As to the JUnit test, how should I mock a CallbackHandler for testing?

src/src/com/microsoft/aad/adal/AuthenticationContext.java
```diff
@@ -1138,6 +1138,36 @@ private AuthenticationResult acquireTokenLocalCall(final CallbackHandler callbac
...
+    //if the redirectUri from the client does not match the valid redirectUri
+    // the client app would not jump to the login page
+    //redirectUri format %PREFIX://%PACKAGE_NAME/%SIGNATURE   
+    private boolean checkRedictUri(CallbackHandler callbackHandle, final AuthenticationRequest request) throws UnsupportedEncodingException{
+    	
+    	String errMsg = new String();    	
+    	String inputUri = request.getRedirectUri();
+    	PackageHelper helper = new PackageHelper(mContext);
+    	
+    	if(inputUri == null){
```
weijjia: you also need to verify if inputUri is empty. Use StringExtensions.isNullorBlank

PR #471, opened 2016-01-12
Exceptions Review

heidijinxujia:
Question,
In which situation, the catch of checked exceptions need to call returnToCaller()? (Compare Line 213 with Line 446)
Thanks, :)！

### DIFF
--- a/src/src/com/microsoft/aad/adal/AuthenticationActivity.java
+++ b/src/src/com/microsoft/aad/adal/AuthenticationActivity.java
@@ -127,7 +127,7 @@ public class AuthenticationActivity extends Activity {
             Logger.v(TAG, "ActivityBroadcastReceiver onReceive");
 
             if (intent.getAction().equalsIgnoreCase(AuthenticationConstants.Browser.ACTION_CANCEL)) {
-                try {
+            //  try {
                     Logger.v(TAG,
                             "ActivityBroadcastReceiver onReceive action is for cancelling Authentication Activity");
 
@@ -141,15 +141,11 @@ public class AuthenticationActivity extends Activity {
                         // cancelled
                         // and callback will be called after this request.
                     }
-                } catch (Exception ex) {
-                	//we need to re-throw runtime exceptions after logging
-                    //if(e instanceof NullPointerException){ throw ex; }
-                    //@heidi 
-                    //@Dec 15 2015
+                /*} catch (Exception ex) {
                 	Logger.e(TAG, "ActivityBroadcastReceiver onReceive exception",
-                            ExceptionExtensions.getExceptionMessage(ex),
-                            ADALError.BROADCAST_RECEIVER_ERROR);
-                }
+                			ExceptionExtensions.getExceptionMessage(ex),
+                			ADALError.BROADCAST_RECEIVER_ERROR);
+                }*/
             }
         }
     }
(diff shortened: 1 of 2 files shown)


src/src/com/microsoft/aad/adal/AuthenticationActivity.java
```diff
@@ -127,7 +127,7 @@ public void onReceive(Context context, Intent intent) {
             Logger.v(TAG, "ActivityBroadcastReceiver onReceive");
 
             if (intent.getAction().equalsIgnoreCase(AuthenticationConstants.Browser.ACTION_CANCEL)) {
-                try {
+            //  try {
```
RandalliLama: I assume you are going to remove these lines.  Would you mind just doing that.  It clutters code reviews to see things that are in progress.

src/src/com/microsoft/aad/adal/AuthenticationContext.java
```diff
@@ -1153,6 +1153,16 @@ private AuthenticationResult acquireTokenAfterValidation(CallbackHandler callbac
             AuthenticationResult result = null;
             request.setVersion(getVersionName());
             request.setBrokerAccountName(request.getLoginHint());
+            
+            //check the redirectUri before sending the request
+            //if the redirectUri from the client does not match the valid redirectUri
+            // the client app would not jump to the login page
```
RandalliLama: We only want to do this is skipBroker is false.

weijjia: This is under the case canSwithToBroker=true. But this code looks different compared to another PR for checking redirect uri @heidijinxujia .

weijjia:
Can we get the travis issue resolved?

src/src/com/microsoft/aad/adal/AuthenticationActivity.java
```diff
@@ -657,9 +658,10 @@ public void alias(String alias) {
                         request.proceed(privateKey, certChain);
                         
                     } catch (KeyChainException e) {
-                        Log.e(TAG, "KeyChain exception", e);
+                        Log.e(TAG + ".onReceivedClientCertRequest()", "The alias of the desired certificate chain is valid, but have problems on accessing it.", e);
                     } catch (InterruptedException e) {
-                        Log.e(TAG, "InterruptedException exception", e);
+                    	//KeyChain.bindAsUser() throws InterruptedException
```
weijjia: make sure we get all the commented code removed.

src/src/com/microsoft/aad/adal/AuthenticationActivity.java
```diff
@@ -215,7 +211,10 @@ protected void onCreate(Bundle savedInstanceState) {
             mStartUrl = oauth.getCodeRequestUrl();
             mQueryParameters = oauth.getAuthorizationEndpointQueryParameters();
         } catch (UnsupportedEncodingException e) {
-            Log.d(TAG, e.getMessage());
+        	//checked exceptions
```
weijjia: Make sure you get this removed. 
For this case, AuthenticationActivity is called in AuthenticationContext, we need to make sure we return the error back to AuthenticationContext.

weijjia:
Make sure you get the conflicts resolved. Also, we should have more exception cases need to be covered, are you going to send them in multiple reviews?

PR #476, opened 2016-01-19
should continue with regular sign-in when no cert found for client TL…

weijjia:
When webview receives the client TLS request, if there is no cert found, and user clicks on cancel, we should continue with the regular sign-in. 

also update the travis fail since updating the android version to 21


kpanwar:
:shipit:

PR #477, opened 2016-01-20
Including fragment example in the README file

murki:
An example for how to ask for a Token if using a Fragment was missing from the documentation, so I'm adding a code example along with the logic to wrap said Fragment on a IWindowComponent


weijjia:
Thanks for the contribution.

Issue #479, opened 2016-01-22
AuthenticationCallback passed in acquireToken() can reference an invalid UI causing memory leaks and NullPointerExceptions


murki:
Since the Activity/Fragment can be destroyed at anytime by Android there is a big risk of memory leaks and the app crashing when implementing AuthenticationCallback's `onSuccess()` and `onError()` methods.

The easiest way to see this is by enabling the "Don't keep actives" option on Android's Developer options. After that methods like Fragment's `isAdded()` and `getActivity()` would report an invalid state when used inside the AuthenticationCallback.

``` java
@Override
public void onCreate(Bundle savedInstanceState) {
  //...
  AuthenticationContext adalAuthContext = new AuthenticationContext(getActivity().getApplicationContext(), AUTHORITY_URL, true);
  //…
}

public void loginClick() {
  adalAuthContext.acquireToken(wrapFragment(this), RESOURCE_ID, CLIENT_ID, REDIRECT_URL, emailTextView.getText().toString(), PromptBehavior.Always, "", new AuthenticationCallback<AuthenticationResult>() { // <-- This callback holds a reference to the "original" fragment (the one used to call `startActivityForResult()`)
    @Override
    public void onSuccess(AuthenticationResult authenticationResult) {
      String token = authenticationResult.getAccessToken(); // <-- Valid token is returned, but can't be used
      Boolean isFragmentInAValidState = isAdded() && getActivity() != null; // <— This is false, which means the callback is referencing an old fragment instance effectively leaking it, and also causing all kinds of NPEs
    }

    @Override
    public void onError(Exception e) {
      //...
    }
  });
}

@Override
public void onActivityResult(int requestCode, int resultCode, Intent data) {
  // At this point, this is a new Fragment that has been created by Android
  super.onActivityResult(requestCode, resultCode, data);
  adalAuthContext.onActivityResult(requestCode, resultCode, data);
}
```

This issue has been discussed with @weijjia and she is looking into it.

RandalliLama:
@murki @weijjia If we fix this will there be backward compatibility issues with apps that did not have bugs?

murki:
@RandalliLama I think all apps are exposed to this bug (again the easy way to see it, is by enabling the dev option). 

My naive workaround would not break compatibility, it just adds an extra method overload to an existing method. 

[...]

PR #480, opened 2016-01-22
Naive work-around for #479

murki:
Potential workaround for #479. 

Here I'm adding an overload for `AuthenticationContext.onActivityResult()` that will optionally receive a callback and replace its internal reference with the one provided. This would solve the issue of using a callback with a reference to an invalid Activity/Fragment since when Android calls the `onActivityResult()` method, it happens on the new valid Activity/Fragment.


weijjia:
@murki Close it for now. We're going to look into the case when android kills the activity and get back to the issue you already filed.

Issue #426, opened 2015-09-17
When a redirect URI is passed ADAL should respect it and not generate a different one.


RandalliLama:
In an OOB conversation @kpanwar mentioned that ADAL Android will try to intelligently generate an Authenticator compatible redirect uri and use that instead of the developer provided URI in some cases.  This will cause all kinds of problems.  We should revert this behavior.

Issue #350, opened 2015-02-26
Throw exception instead of silent switch from Broker mode


afshins:
In method CanSwitchToBroker, instead of checking all the conditions and silently switching to not use broker in one of them fails, ADAL should throw exception if any condition other than developer's explicit intent (which is via skipBroker flag) is not met.

addev-ashish:
Fixed in PR#483, PR#473, PR#475

PR #483, opened 2016-01-28
Implemented the redirectUri checking function.

heidijinxujia:
Implemented the redirectUri checking function.The redirectUri is checked before sending the request. If the redirectUri from the client does not match the valid redirectUri, we will log the error and the client app will not jump to the login page.


heidijinxujia:
need unit testing for the method (In progress)

src/src/com/microsoft/aad/adal/AuthenticationContext.java
```diff
@@ -1132,6 +1133,51 @@ private AuthenticationResult acquireTokenLocalCall(final CallbackHandler callbac
...
+     * if the redirectUri from the client does not match the valid redirectUri
+     * the client app would not jump to the login page
+     * redirectUri format %PREFIX://%PACKAGE_NAME/%SIGNATURE
+     * 
+     * @param callbackHandle
+     * @param request
+     * @throws UnsupportedEncodingException
+     * @return true if the RedictUri is valid
+     */
+    private boolean checkRedirectUri(CallbackHandler callbackHandle, final AuthenticationRequest request) throws UnsupportedEncodingException{
```
weijjia: can we change the name to verifyBrokerRedirectUri?

src/src/com/microsoft/aad/adal/AuthenticationContext.java
```diff
@@ -1132,6 +1133,51 @@ private AuthenticationResult acquireTokenLocalCall(final CallbackHandler callbac
...
+        
+        String errMsg = new String();        
+        String inputUri = request.getRedirectUri();
+        PackageHelper helper = new PackageHelper(mContext);
+        
+        if(StringExtensions.IsNullOrBlank(inputUri))
+        {
+            errMsg = "the redirectUri is null";
```
weijjia: nit: is null or blank

src/src/com/microsoft/aad/adal/AuthenticationContext.java
```diff
@@ -1147,6 +1193,19 @@ private AuthenticationResult acquireTokenAfterValidation(CallbackHandler callbac
             AuthenticationResult result = null;
             request.setVersion(getVersionName());
             request.setBrokerAccountName(request.getLoginHint());
+            
+            //check if the redirectUri is valid
+            try
```
weijjia: i think we should move the try/catch inside the redirect uri check method itself.

src/src/com/microsoft/aad/adal/AuthenticationContext.java
```diff
@@ -1132,6 +1133,51 @@ private AuthenticationResult acquireTokenLocalCall(final CallbackHandler callbac
...
+        }
+        else if((inputUri.split("://").length == 2 && !inputUri.split("://")[1].split("/")[0].equals(URLEncoder.encode(mContext.getPackageName(), AuthenticationConstants.ENCODING_UTF8))) || inputUri.split("://").length == 1 )
+        {
+            errMsg = "the package name of redirectUri is not as expected. The expected package name is" + mContext.getPackageName();
+        }
+        else if((inputUri.split("://")[1].split("/").length == 2 && !inputUri.split("://")[1].split("/")[1].equals(URLEncoder.encode(helper.getCurrentSignatureForPackage(mContext.getPackageName()), AuthenticationConstants.ENCODING_UTF8)))||inputUri.split("://")[1].split("/").length == 1)
+        {
+            errMsg = "the signature of redirectUri is not as expected. The expected signiture is " + URLEncoder.encode(helper.getCurrentSignatureForPackage(mContext.getPackageName()), AuthenticationConstants.ENCODING_UTF8);
+        }
+        if(errMsg.isEmpty())
```
weijjia: I prefer to see the early return for error case, and return true at the end.

src/src/com/microsoft/aad/adal/AuthenticationContext.java
```diff
@@ -1132,6 +1133,51 @@ private AuthenticationResult acquireTokenLocalCall(final CallbackHandler callbac
...
+     * @param callbackHandle
+     * @param request
+     * @throws UnsupportedEncodingException
+     * @return true if the RedictUri is valid
+     */
+    private boolean checkRedirectUri(CallbackHandler callbackHandle, final AuthenticationRequest request) throws UnsupportedEncodingException{
+        
+        String errMsg = new String();        
+        String inputUri = request.getRedirectUri();
+        PackageHelper helper = new PackageHelper(mContext);
```
weijjia: I will name it packageHelper, it is easy for ppl to read the code.

src/src/com/microsoft/aad/adal/AuthenticationContext.java
```diff
@@ -1132,6 +1133,51 @@ private AuthenticationResult acquireTokenLocalCall(final CallbackHandler callbac
...
+        
+        if(StringExtensions.IsNullOrBlank(inputUri))
+        {
+            errMsg = "the redirectUri is null";            
+        }
+        else if(!inputUri.split("://")[0].equals(AuthenticationConstants.Broker.REDIRECT_PREFIX))
```
weijjia: All the checks are correct, but I think we're making it too complicated.

heidijinxujia:
I've fixed the bugs in the review. However, currently I have problems in the unit test for mocking the argument CallbackHandler.

src/src/com/microsoft/aad/adal/AuthenticationContext.java
```diff
@@ -1132,10 +1133,62 @@ private AuthenticationResult acquireTokenLocalCall(final CallbackHandler callbac
...
+        String inputUri = request.getRedirectUri();
+        PackageHelper helper = new PackageHelper(mContext);
+        
+        if(StringExtensions.IsNullOrBlank(inputUri))
+        {
+            errMsg = "the redirectUri is null or blank";            
+        }
+        else if(!inputUri.startsWith(AuthenticationConstants.Broker.REDIRECT_PREFIX + "://"))
+        {
+            errMsg = "the redirectUri is not starting with the valid redict prefix " + AuthenticationConstants.Broker.REDIRECT_PREFIX;
```
RandalliLama: change to "redirectUri does not start with"

src/src/com/microsoft/aad/adal/AuthenticationContext.java
```diff
@@ -1132,10 +1133,62 @@ private AuthenticationResult acquireTokenLocalCall(final CallbackHandler callbac
...
+        if(StringExtensions.IsNullOrBlank(inputUri))
+        {
+            errMsg = "the redirectUri is null or blank";            
+        }
+        else if(!inputUri.startsWith(AuthenticationConstants.Broker.REDIRECT_PREFIX + "://"))
+        {
+            errMsg = "the redirectUri is not starting with the valid redict prefix " + AuthenticationConstants.Broker.REDIRECT_PREFIX;            
+        } else
+			try {
+				if(!inputUri.startsWith(AuthenticationConstants.Broker.REDIRECT_PREFIX + "://" + URLEncoder.encode(mContext.getPackageName(), AuthenticationConstants.ENCODING_UTF8) + "/"))
```
RandalliLama: Consider putting this line of code into a method of it's own:

URLEncoder.encode(mContext.getPackageName(), AuthenticationConstants.ENCODING_UTF8)

and in that method do the try catch.

RandalliLama: Consider doing this code once and assigning to a variable that can be used in two places:

URLEncoder.encode(mContext.getPackageName(), AuthenticationConstants.ENCODING_UTF8)

src/src/com/microsoft/aad/adal/AuthenticationContext.java
```diff
@@ -1132,10 +1133,62 @@ private AuthenticationResult acquireTokenLocalCall(final CallbackHandler callbac
...
+        
+        if(StringExtensions.IsNullOrBlank(inputUri))
+        {
+            errMsg = "the redirectUri is null or blank";            
+        }
+        else if(!inputUri.startsWith(AuthenticationConstants.Broker.REDIRECT_PREFIX + "://"))
+        {
+            errMsg = "the redirectUri is not starting with the valid redict prefix " + AuthenticationConstants.Broker.REDIRECT_PREFIX;            
+        } else
+			try {
```
RandalliLama: Add the curly brace after this.  All if/elses and other clauses should be wrapped in curly braces.

src/src/com/microsoft/aad/adal/AuthenticationContext.java
```diff
@@ -1132,10 +1133,62 @@ private AuthenticationResult acquireTokenLocalCall(final CallbackHandler callbac
...
+				}
+				else if(!inputUri.equalsIgnoreCase(getRedirectUriForBroker()))
+				{
+				    errMsg = "the signature of redirectUri is not as expected. The expected signiture is " + URLEncoder.encode(helper.getCurrentSignatureForPackage(mContext.getPackageName()), AuthenticationConstants.ENCODING_UTF8);
+				}
+			} catch (UnsupportedEncodingException e) {
+				Logger.e(TAG + methodName, "Digest error", "", ADALError.ENCODING_IS_NOT_SUPPORTED, e);
+				callbackHandle.onError(new AuthenticationException(ADALError.ENCODING_IS_NOT_SUPPORTED, "Digest error"));
+			}
+        if(!errMsg.isEmpty())
```
RandalliLama: Add a space before this line to make it clear that it is not part of the previous set of clauses.

src/src/com/microsoft/aad/adal/AuthenticationContext.java
```diff
@@ -1132,10 +1133,62 @@ private AuthenticationResult acquireTokenLocalCall(final CallbackHandler callbac
...
+        {
+        	Logger.e(TAG + methodName, errMsg , "", ADALError.DEVELOPER_REDIRECTURI_INVALID);            
+            callbackHandle.onError(new AuthenticationException(ADALError.DEVELOPER_REDIRECTURI_INVALID, errMsg));
```
RandalliLama: This is indented differently for some reason.  Make sure that all indentation is being done with spaces rather than tabs.

src/src/com/microsoft/aad/adal/AuthenticationContext.java
```diff
@@ -1132,10 +1133,62 @@ private AuthenticationResult acquireTokenLocalCall(final CallbackHandler callbac
...
+				{
+				    errMsg = "the package name of redirectUri is not as expected. The expected package name is" + URLEncoder.encode(mContext.getPackageName(), AuthenticationConstants.ENCODING_UTF8);
+				}
+				else if(!inputUri.equalsIgnoreCase(getRedirectUriForBroker()))
+				{
+				    errMsg = "the signature of redirectUri is not as expected. The expected signiture is " + URLEncoder.encode(helper.getCurrentSignatureForPackage(mContext.getPackageName()), AuthenticationConstants.ENCODING_UTF8);
+				}
+			} catch (UnsupportedEncodingException e) {
+				Logger.e(TAG + methodName, "Digest error", "", ADALError.ENCODING_IS_NOT_SUPPORTED, e);
+				callbackHandle.onError(new AuthenticationException(ADALError.ENCODING_IS_NOT_SUPPORTED, "Digest error"));
```
RandalliLama: It looks like it might be possible to call onError more than once.  It would be better to have this on exactly one line of code.  The other way to do it might be to put a return statement after this, but I would prefer a single place where we exit the method on error.

src/src/com/microsoft/aad/adal/AuthenticationContext.java
```diff
@@ -1132,10 +1133,62 @@ private AuthenticationResult acquireTokenLocalCall(final CallbackHandler callbac
...
+        {
+        	Logger.v(TAG + methodName, "The broker redirect URI is valid.");
+        	return true;
+        }        
+    }
 
     private AuthenticationResult acquireTokenAfterValidation(CallbackHandler callbackHandle,
             final IWindowComponent activity, final boolean useDialog,
             final AuthenticationRequest request) {
+    	final String methodName = ":acquireTokenAfterValidation";
```
RandalliLama: Update the rest of the log statements in this method to use the constant as well.

heidijinxujia:
@weijjia Ready for code review. Thanks~!

heidijinxujia:
@kpanwar As well as this pull request?

src/src/com/microsoft/aad/adal/AuthenticationContext.java
```diff
@@ -1132,10 +1133,81 @@ private AuthenticationResult acquireTokenLocalCall(final CallbackHandler callbac
...
+     * @param request
+     * @throws UnsupportedEncodingException
+     * @return true if the RedictUri is valid
+     */
+    private boolean verifyBrokerRedirectUri(CallbackHandler callbackHandle, final AuthenticationRequest request){   
+        final String methodName = ":verifyBrokerRedirectUri";
+        String errMsg = new String();        
+        String inputUri = request.getRedirectUri();
+        PackageHelper helper = new PackageHelper(mContext);
+        AuthenticationException excep = null;
```
weijjia: nit: i want this to be renamed to be exception.

src/src/com/microsoft/aad/adal/AuthenticationContext.java
```diff
@@ -1132,10 +1133,81 @@ private AuthenticationResult acquireTokenLocalCall(final CallbackHandler callbac
...
+     * redirectUri format %PREFIX://%PACKAGE_NAME/%SIGNATURE
+     * 
+     * @param callbackHandle
+     * @param request
+     * @throws UnsupportedEncodingException
+     * @return true if the RedictUri is valid
+     */
+    private boolean verifyBrokerRedirectUri(CallbackHandler callbackHandle, final AuthenticationRequest request){   
+        final String methodName = ":verifyBrokerRedirectUri";
+        String errMsg = new String();
```
weijjia: Does it really make sense to make use the constructor here? Can't you just do String errMsg = "";

src/src/com/microsoft/aad/adal/AuthenticationContext.java
```diff
@@ -1132,10 +1133,81 @@ private AuthenticationResult acquireTokenLocalCall(final CallbackHandler callbac
...
+        String inputUri = request.getRedirectUri();
+        PackageHelper helper = new PackageHelper(mContext);
+        AuthenticationException excep = null;
+        
+        if(StringExtensions.IsNullOrBlank(inputUri))
+        {
+            errMsg = "the redirectUri is null or blank";   
+            excep = new AuthenticationException(ADALError.DEVELOPER_REDIRECTURI_INVALID, errMsg);
+            Logger.e(TAG + methodName, errMsg , "", ADALError.DEVELOPER_REDIRECTURI_INVALID);     
+
```
weijjia: nit: extra line

src/src/com/microsoft/aad/adal/AuthenticationContext.java
```diff
@@ -1132,10 +1133,81 @@ private AuthenticationResult acquireTokenLocalCall(final CallbackHandler callbac
...
+     * 
+     * @param callbackHandle
+     * @param request
+     * @throws UnsupportedEncodingException
+     * @return true if the RedictUri is valid
+     */
+    private boolean verifyBrokerRedirectUri(CallbackHandler callbackHandle, final AuthenticationRequest request){   
+        final String methodName = ":verifyBrokerRedirectUri";
+        String errMsg = new String();        
+        String inputUri = request.getRedirectUri();
```
weijjia: nit: Are you going to change inputUri? If not, it could be final.

src/src/com/microsoft/aad/adal/AuthenticationContext.java
```diff
@@ -1132,10 +1133,81 @@ private AuthenticationResult acquireTokenLocalCall(final CallbackHandler callbac
...
+                {
+                    errMsg = "the signature of redirectUri is not as expected. The expected signature is " + urlEncodedSignature;
+                    excep = new AuthenticationException(ADALError.DEVELOPER_REDIRECTURI_INVALID, errMsg);
+                    Logger.e(TAG + methodName, errMsg , "", ADALError.DEVELOPER_REDIRECTURI_INVALID);     
+                   
+                }
+            } 
+            catch (UnsupportedEncodingException e) 
+            {
+                excep = new AuthenticationException(ADALError.ENCODING_IS_NOT_SUPPORTED, "Digest error");
```
weijjia: why the exception message is digest error? Can you put the real exception message here? using e.getMessage

src/src/com/microsoft/aad/adal/AuthenticationContext.java
heidijinxujia: The string is initialised with no characters, or "" internally. It's the same.

public String() {
    this.offset = 0;
    this.count = 0;
    this.value = new char[0];
}

src/src/com/microsoft/aad/adal/AuthenticationContext.java
```diff
@@ -1132,10 +1133,81 @@ private AuthenticationResult acquireTokenLocalCall(final CallbackHandler callbac
...
+     * @param callbackHandle
+     * @param request
+     * @throws UnsupportedEncodingException
+     * @return true if the RedictUri is valid
+     */
+    private boolean verifyBrokerRedirectUri(CallbackHandler callbackHandle, final AuthenticationRequest request){   
+        final String methodName = ":verifyBrokerRedirectUri";
+        String errMsg = new String();        
+        String inputUri = request.getRedirectUri();
+        PackageHelper helper = new PackageHelper(mContext);
```
weijjia: nit: i want this to be renamed as packageHelper. packageHelper would be easy for people reading the code know what it is. Also, you only use it in line 1173, you can declare it where you use it.

weijjia:
:shipit:

src/src/com/microsoft/aad/adal/AuthenticationContext.java
```diff
@@ -1132,10 +1133,80 @@ private AuthenticationResult acquireTokenLocalCall(final CallbackHandler callbac
...
+            exception = new AuthenticationException(ADALError.DEVELOPER_REDIRECTURI_INVALID, errMsg);
+            Logger.e(TAG + methodName, errMsg , "", ADALError.DEVELOPER_REDIRECTURI_INVALID);     
+        } 
+        else
+        {
+            try 
+            {
+                PackageHelper packageHelper = new PackageHelper(mContext);
+                String urlEncodedPackagename = URLEncoder.encode(mContext.getPackageName(), AuthenticationConstants.ENCODING_UTF8);
+                String urlEncodedSignature = URLEncoder.encode(packageHelper.getCurrentSignatureForPackage(mContext.getPackageName()), AuthenticationConstants.ENCODING_UTF8);
```
kpanwar: nitpick: this is a base64 url encoded value and should be represented as that in variable name and URLEncoder method name. Please file a lo-pri bug to eventually make this synthetic change.

src/src/com/microsoft/aad/adal/AuthenticationContext.java
```diff
@@ -1132,10 +1133,80 @@ private AuthenticationResult acquireTokenLocalCall(final CallbackHandler callbac
...
+        }
+        
+        if(exception != null)
+        { 
+            callbackHandle.onError(exception);
+            return false;
+        }
+        else
+        {
+            Logger.v(TAG + methodName, "The broker redirect URI is valid.");
```
kpanwar: We should log the actual URI value here if we are not doing it anywhere else.

kpanwar:
:shipit: with 1 minor comment and 1 logging change.

src/src/com/microsoft/aad/adal/AuthenticationContext.java
heidijinxujia: Is it OK to change it to "urlEncodedPackagename_UTF8" or you are expected another format? @kpanwar

heidijinxujia:
@weijjia ready for code review~

tests/Functional/src/com/microsoft/aad/adal/test/AuthenticationContextTest.java
```diff
@@ -1906,6 +1906,70 @@ private AuthenticationContext getAuthenticationContext(Context mockContext, Stri
...
+        Method m = ReflectionUtils.getTestMethod(authContext, "verifyBrokerRedirectUri", c);
+
+        //test@case valid redirect uri
+        String testRedirectUri = authContext.getRedirectUriForBroker();
+        Object authRequest = AuthenticationContextTest.createAuthenticationRequest(VALID_AUTHORITY, 
+                "resource", "clientid", testRedirectUri, "loginHint");
+        Boolean testResult = (Boolean)m.invoke(authContext, authRequest);
+        assertTrue(testResult);
+
+        //test@case broker redirect uri with invalid prefix
```
weijjia: Again, like the comment in the other code review, i want each individual test is only responsible for one thing, it makes it so clear to know what the test is testing for.

tests/Functional/src/com/microsoft/aad/adal/test/AuthenticationContextTest.java
```diff
@@ -1906,6 +1906,70 @@ private AuthenticationContext getAuthenticationContext(Context mockContext, Stri
 
         return context;
     }
+
```
weijjia: May be picky, but can we have a component test? pass in the client id, redirect uri and all the other required parameters, and then call acquiretoken, make sure if it went through the broker flow and we get the exception with the correct message back when the redirect uri is not right.

src/src/com/microsoft/aad/adal/AuthenticationContext.java
RandalliLama: @kpanwar nitpick: What is a "synthetic change"?

If we are just talking about renaming something then let's not add yet another issue that will never get done.  Let's just do it.  Unless I am missing something and it is a harder change than it sounds.

src/src/com/microsoft/aad/adal/AuthenticationContext.java
```diff
@@ -1132,10 +1133,79 @@ private AuthenticationResult acquireTokenLocalCall(final CallbackHandler callbac
...
+        if(StringExtensions.IsNullOrBlank(inputUri))
+        {
+            errMsg = "the redirectUri is null or blank";
+            exception = new AuthenticationException(ADALError.DEVELOPER_REDIRECTURI_INVALID, errMsg);
+            Logger.e(TAG + methodName, errMsg , "", ADALError.DEVELOPER_REDIRECTURI_INVALID);     
+        }
+        else if(!inputUri.startsWith(AuthenticationConstants.Broker.REDIRECT_PREFIX + "://"))
+        {
+            errMsg = "redirectUri does not start with the valid redict prefix " + AuthenticationConstants.Broker.REDIRECT_PREFIX; 
+            exception = new AuthenticationException(ADALError.DEVELOPER_REDIRECTURI_INVALID, errMsg);
```
RandalliLama: Please change "redict" to "redirect".

src/src/com/microsoft/aad/adal/AuthenticationContext.java
```diff
@@ -1132,10 +1133,79 @@ private AuthenticationResult acquireTokenLocalCall(final CallbackHandler callbac
...
+        
+        if(StringExtensions.IsNullOrBlank(inputUri))
+        {
+            errMsg = "the redirectUri is null or blank";
+            exception = new AuthenticationException(ADALError.DEVELOPER_REDIRECTURI_INVALID, errMsg);
+            Logger.e(TAG + methodName, errMsg , "", ADALError.DEVELOPER_REDIRECTURI_INVALID);     
+        }
+        else if(!inputUri.startsWith(AuthenticationConstants.Broker.REDIRECT_PREFIX + "://"))
+        {
+            errMsg = "redirectUri does not start with the valid redict prefix " + AuthenticationConstants.Broker.REDIRECT_PREFIX;
```
RandalliLama: Please change "redict" to "redirect"

src/src/com/microsoft/aad/adal/AuthenticationContext.java
```diff
@@ -1132,10 +1133,79 @@ private AuthenticationResult acquireTokenLocalCall(final CallbackHandler callbac
...
+        else
+        {
+            try 
+            {
+                PackageHelper packageHelper = new PackageHelper(mContext);
+                String base64URLEncodePackagename = URLEncoder.encode(mContext.getPackageName(), AuthenticationConstants.ENCODING_UTF8);
+                String base64URLEncodeSignature = URLEncoder.encode(packageHelper.getCurrentSignatureForPackage(mContext.getPackageName()), AuthenticationConstants.ENCODING_UTF8);
+                if(!inputUri.startsWith(AuthenticationConstants.Broker.REDIRECT_PREFIX + "://" + base64URLEncodePackagename + "/"))
+                {
+                    errMsg = "the package name of redirectUri is not as expected. The expected package name is " + base64URLEncodePackagename;
```
RandalliLama: Let's change this message to "The base64 url encoded package name component of the redirect uri does not match the expected value.  This apps package name is: <package_name> so the redirect uri is expected to be: <calculated_redirect_uri>"

We should try to make error messages as clear and actionable as possible.  If it's possible to create a message that tells the developer exactly what they need to do to fix the problem, then we should do so.

src/src/com/microsoft/aad/adal/AuthenticationContext.java
RandalliLama: We need to make this message more prescriptive.  I think it should direct the developer to the readme, and we need to add a section to the readme that explains the issues around invoking the broker.

Please file an issue to track the need to scrub exceptions and make them prescriptive and point to documentation that can help the developer figure out what they need to change.

src/src/com/microsoft/aad/adal/AuthenticationContext.java
```diff
@@ -1132,10 +1133,87 @@ private AuthenticationResult acquireTokenLocalCall(final CallbackHandler callbac
...
+            catch (UnsupportedEncodingException e) 
+            {
+                exception = new AuthenticationException(ADALError.ENCODING_IS_NOT_SUPPORTED, "The base64 url encoding is not supported."); 
+                Logger.e(TAG + methodName, e.getMessage(), "", ADALError.ENCODING_IS_NOT_SUPPORTED, e);
+            }
+        }
+        
+        if(exception != null)
+        { 
+            throw exception;
```
RandalliLama: Why not just re-throw the exception from inside the catch block above.  Then you don't need this check.

src/src/com/microsoft/aad/adal/AuthenticationContext.java
```diff
@@ -1132,10 +1133,87 @@ private AuthenticationResult acquireTokenLocalCall(final CallbackHandler callbac
...
+     * if the redirectUri from the client does not match the valid redirectUri
+     * the client app would not jump to the login page
+     * redirectUri format %PREFIX://%PACKAGE_NAME/%SIGNATURE
+     * 
+     * @param callbackHandle
+     * @param request
+     * @throws UnsupportedEncodingException
+     * @return true if the redirectUri is valid
+     */
+    private boolean verifyBrokerRedirectUri(final AuthenticationRequest request){
```
RandalliLama: Put a space between the parenthesis and the opening curly brace.

src/src/com/microsoft/aad/adal/AuthenticationContext.java
```diff
@@ -1132,10 +1133,87 @@ private AuthenticationResult acquireTokenLocalCall(final CallbackHandler callbac
...
+                            + "This apps signature is: " + base64URLEncodeSignature
+                            + "so the redirect uri is expected to be:" + actualUri;
+                    exception = new AuthenticationException(ADALError.DEVELOPER_REDIRECTURI_INVALID, errMsg);
+                    Logger.e(TAG + methodName, errMsg , "", ADALError.DEVELOPER_REDIRECTURI_INVALID);     
+                   
+                }
+            } 
+            catch (UnsupportedEncodingException e) 
+            {
+                exception = new AuthenticationException(ADALError.ENCODING_IS_NOT_SUPPORTED, "The base64 url encoding is not supported.");
```
RandalliLama: Add the Unsupported exception that was caught as the "cause" property of the AuthenticationException.

RandalliLama:
Fix those last two things and I'll be able to sign-off.

src/src/com/microsoft/aad/adal/AuthenticationContext.java
heidijinxujia: This try-catch only deal with the UnsupportedEncodingException in the latest two checking(pmackagenmae and  signature endcoding).  I can refactor it, if you think it is better to put all catch in one single block.

RandalliLama: Is there a reason that we don't just throw the other exceptions as soon as they are constructed?  Why do we save them in a variable and throw them later?

RandalliLama:
:shipit:

RandalliLama:
@wei are you ok with the change at this point?

weijjia:
@RandalliLama Yup, i'm fine with it.

addev-ashish:
#350

addev-ashish:
#426

PR #486, opened 2016-01-30
not repond to client tls if issuer contains CN=MS-Organization-Access

weijjia:
1. when webview receives client TLS request, if acceptable issuer contains CN=MS-Organization-Access, then do not respond to the request. 
2. Workarround for responding to device auth challenge returned from ADFS on token endpoint. ADFS is returning an empty cert thumbprint, while ADAL asks either cert thumbprint or cert authority be to present. Fix for this is to check if the device is already workplace joined before checking for the presence of cert thumbprint or authority. 
3. fix some logs. 
4. add unit test for 2

### DIFF
--- a/src/src/com/microsoft/aad/adal/AuthenticationActivity.java
+++ b/src/src/com/microsoft/aad/adal/AuthenticationActivity.java
@@ -28,6 +28,7 @@ import java.security.InvalidKeyException;
 import java.security.KeyStoreException;
 import java.security.NoSuchAlgorithmException;
 import java.security.NoSuchProviderException;
+import java.security.Principal;
 import java.security.PrivateKey;
 import java.security.UnrecoverableEntryException;
 import java.security.cert.CertificateException;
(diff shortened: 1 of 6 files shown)


weijjia:
@sdabbiru Can you also take a look at this PR?

src/src/com/microsoft/aad/adal/AuthenticationActivity.java
```diff
@@ -640,8 +646,23 @@ public void postRunnable(Runnable item) {
         
         @TargetApi(Build.VERSION_CODES.LOLLIPOP)
         @Override
-        public void onReceivedClientCertRequest (WebView view, final ClientCertRequest request){
-            Logger.v(TAG + "onReceivedClientCertRequest", "onReceivedClientCertRequest");
+        public void onReceivedClientCertRequest (WebView view, final ClientCertRequest request)
+        {
+            final String methodName = ":onReceivedClientCertRequest";
+            Logger.v(TAG + methodName, "Webview receives client TLS request.");
```
RPangrle: It might be better to grab the calling method name from a stack trace in the logger (http://stackoverflow.com/questions/5439406/java-equivalent-to-func) and not have to worry about copy/paste errors.

RPangrle:
:shipit:

src/src/com/microsoft/aad/adal/AuthenticationActivity.java
weijjia: yup, but this will be runtime, and it will be costly.

heidijinxujia:
Em.. is there a conclusion that which format we should use in Logger? @weijjia

RPangrle:
I would leave that as a future work item. It's much more broadly scoped than this code review.

PR #494, opened 2016-02-10
For silent request throw exception when access token couldn't be obtained

imakhalova:
Recreated pull request against dev branch (instead of master)
Currently if server response was something that doesn't contain access token, it wipes refresh token from cache and tries to acqure token again.
 But if the request was silent we will fail with error "Prompt is not allowed and failed to get token", which doesn't give any info why token was wiped.
 I propose in case of silent request shouldn't even try to acquire token, instead throw exception with server response error. It will give info why token was wiped.


src/src/com/microsoft/aad/adal/AuthenticationContext.java
```diff
@@ -1625,7 +1625,11 @@ private AuthenticationResult refreshToken(final CallbackHandler callbackHandle,
                 // remove item from cache to avoid same usage of
                 // refresh token in next acquireToken call
                 removeItemFromCache(refreshItem);
-                return acquireTokenLocalCall(callbackHandle, activity, useDialog, request);
+                if (!request.isSilent() && callbackHandle.callback != null && activity != null) {
+                    return acquireTokenLocalCall(callbackHandle, activity, useDialog, request);
+                } else {
+                    throw new AuthenticationException(ADALError.AUTH_FAILED_SERVER_ERROR, request.getLogInfo() + errLogInfo);
```
weijjia: Why not using callback to send the error back to caller instead of throwing AuthenticationException?

imakhalova: You're actually right, much cleaner way to handle it.

src/src/com/microsoft/aad/adal/AuthenticationContext.java
```diff
@@ -1625,7 +1625,13 @@ private AuthenticationResult refreshToken(final CallbackHandler callbackHandle,
                 // remove item from cache to avoid same usage of
                 // refresh token in next acquireToken call
                 removeItemFromCache(refreshItem);
-                return acquireTokenLocalCall(callbackHandle, activity, useDialog, request);
+                if (!request.isSilent() && callbackHandle.callback != null && activity != null) {
+                    return acquireTokenLocalCall(callbackHandle, activity, useDialog, request);
+                } else {
+                    callbackHandle.onError(
+                            new AuthenticationException(ADALError.AUTH_FAILED_SERVER_ERROR, request.getLogInfo() + errLogInfo));
```
weijjia: Can we return the AUTH_REFRESH_FAILED_PROMPT_NOT_ALLOWED error code? This is the error code we advertise to all the adal users for silent request fails, and you need a interactive flow to start again. Changing the error code will break the rule we already set.

imakhalova: Ok, it makes sense to be backward compat. But we'd like to distinguish between AUTH_REFRESH_FAILED_PROMPT_NOT_ALLOWED from aquireTokenProvider and at this place. So let's extend Exception class.

RandalliLama:
I don't think this is the right way to approach the problem that we are attempting solve.  The name AccessTokenException is too broad.  It could apply to almost anything.  Further, the name does not tell the developer anything about what to do when they get this exception.  To me, the overriding design of the exception OM should be that each exception represents a class or errors that have the same or similar remediation.  We have established a pattern across all ADALs that in cases where the refresh token doesn't exist, or cannot be used, we return an error that indicates that user interaction is required.  That's the remediation for this class of errors.

I think that the right way to handle the loss of information in this case is a combination of more specific error enums, and attaching a cause exception if that makes sense.

RandalliLama:
@imakhalova When collecting telemetry do you also collect the error messages in the exceptions?

imakhalova:
@RandalliLama we do collect the error codes and error messages from AuthenticationException

imakhalova:
I'm fine with passing exception cause within AuthenticationException

src/src/com/microsoft/aad/adal/AuthenticationContext.java
```diff
@@ -1625,7 +1625,16 @@ private AuthenticationResult refreshToken(final CallbackHandler callbackHandle,
                 // remove item from cache to avoid same usage of
                 // refresh token in next acquireToken call
                 removeItemFromCache(refreshItem);
-                return acquireTokenLocalCall(callbackHandle, activity, useDialog, request);
+                if (!request.isSilent() && callbackHandle.callback != null && activity != null) {
+                    return acquireTokenLocalCall(callbackHandle, activity, useDialog, request);
+                } else {
+                    // cause exception added to indicate that refresh token was wiped
+                    callbackHandle.onError(new AuthenticationException(
```
weijjia: With the refactor I'm doing with another PR, i don't think this is necessary anymore.

imakhalova: First, I think this change contains extra data that would be nice to have on client side (exception and server logs), and second it would be nice to see your change against this change, that way is easier to see that your refactoring doesn't break this logic

weijjia: Well, I agree. I'll have your exception change out before the FoCI and refactor change.

PR #495, opened 2016-02-10
remove catching generic Exception and catch only expected ones

imakhalova:
+bump up target version to support multi catch statement
Recreated pull request against dev branch (instead of master).
Currently all Runtime exceptions are swallowed and if they happen it's hard to trace the reason why token is not obtained.
 Instead it's better to catch only excepted exceptions, because if unexpected exception happens mobile app should crash and give a stack trace what went wrong, that way it's easy to fix the problem.
 Sample: You have button but when you click nothing happens. Or button that crashes you app and you see NPE in you logs...From user stand point - nothing changes because button doesn't work, from dev standpoint - much more info from second case.

### DIFF
--- a/src/pom.xml
+++ b/src/pom.xml
@@ -179,8 +179,8 @@
         <artifactId>maven-compiler-plugin</artifactId>
         <version>3.1</version>
         <configuration>
-          <source>1.6</source>
-          <target>1.6</target>
+          <source>1.7</source>
+          <target>1.7</target>
         </configuration>
       </plugin>
       <plugin>
--- a/src/src/com/microsoft/aad/adal/AuthenticationActivity.java
+++ b/src/src/com/microsoft/aad/adal/AuthenticationActivity.java
@@ -814,7 +814,7 @@ public class AuthenticationActivity extends Activity {
             try {
                 result.taskResult = oauthRequest.getToken(urlItems[0]);
                 Logger.v(TAG, "TokenTask processed the result. " + mRequest.getLogInfo());
-            } catch (IOException exc) {
+            } catch (IOException | UnexpectedServerResponseException exc) {
                 Logger.e(TAG, "Error in processing code to get a token. " + mRequest.getLogInfo(),
                         "Request url:" + urlItems[0],
                         ADALError.AUTHORIZATION_CODE_NOT_EXCHANGED_FOR_TOKEN, exc);
--- a/src/src/com/microsoft/aad/adal/AuthenticationActivity.java
+++ b/src/src/com/microsoft/aad/adal/AuthenticationActivity.java
@@ -23,6 +23,7 @@ import java.io.Serializable;
 import java.io.UnsupportedEncodingException;
 import java.net.URLEncoder;
 import java.security.DigestException;
+import java.security.GeneralSecurityException;
 import java.security.InvalidAlgorithmParameterException;
 import java.security.InvalidKeyException;
 import java.security.KeyStoreException;
@@ -127,24 +128,18 @@ public class AuthenticationActivity extends Activity {
             Logger.v(TAG, "ActivityBroadcastReceiver onReceive");
 
             if (intent.getAction().equalsIgnoreCase(AuthenticationConstants.Browser.ACTION_CANCEL)) {
-                try {
-                    Logger.v(TAG,
-                            "ActivityBroadcastReceiver onReceive action is for cancelling Authentication Activity");
+                Logger.v(TAG,
+                        "ActivityBroadcastReceiver onReceive action is for cancelling Authentication Activity");
 
-                    int cancelRequestId = intent.getIntExtra(
-                            AuthenticationConstants.Browser.REQUEST_ID, 0);
+                int cancelRequestId = intent.getIntExtra(
+                        AuthenticationConstants.Browser.REQUEST_ID, 0);
 
-                    if (cancelRequestId == mWaitingRequestId) {
-                        Logger.v(TAG, "Waiting requestId is same and cancelling this activity");
-                        AuthenticationActivity.this.finish();
-                        // no need to send result back to activity. It is
-                        // cancelled
-                        // and callback will be called after this request.
-                    }
-                } catch (Exception ex) {
-                    Logger.e(TAG, "ActivityBroadcastReceiver onReceive exception",
-                            ExceptionExtensions.getExceptionMessage(ex),
-                            ADALError.BROADCAST_RECEIVER_ERROR);
+                if (cancelRequestId == mWaitingRequestId) {
+                    Logger.v(TAG, "Waiting requestId is same and cancelling this activity");
+                    AuthenticationActivity.this.finish();
+                    // no need to send result back to activity. It is
+                    // cancelled
+                    // and callback will be called after this request.
                 }
             }
         }
@@ -776,7 +771,7 @@ public class AuthenticationActivity extends Activity {
             try {
                 result.taskResult = oauthRequest.getToken(urlItems[0]);
                 Logger.v(TAG, "TokenTask processed the result. " + mRequest.getLogInfo());
-            } catch (Exception exc) {
+            } catch (IOException exc) {
                 Logger.e(TAG, "Error in processing code to get a token. " + mRequest.getLogInfo(),
                         "Request url:" + urlItems[0],
                         ADALError.AUTHORIZATION_CODE_NOT_EXCHANGED_FOR_TOKEN, exc);
@@ -790,7 +785,7 @@ public class AuthenticationActivity extends Activity {
                 // Record account in the AccountManager service
                 try {
                     setAccount(result);
-                } catch (Exception exc) {
+                } catch (GeneralSecurityException | IOException exc) {
                     Logger.e(TAG, "Error in setting the account" + mRequest.getLogInfo(), "",
                             ADALError.BROKER_ACCOUNT_SAVE_FAILED, exc);
                     result.taskException = exc;
@@ -826,7 +821,7 @@ public class AuthenticationActivity extends Activity {
             } else {
                 try {
                     appIdList = cryptoHelper.decrypt(appIdList);
-                } catch (Exception ex) {
+                } catch (GeneralSecurityException | IOException ex) {
                     Logger.e(TAG, "appUIDList failed to decrypt", "appIdList:" + appIdList,
                             ADALError.ENCRYPTION_FAILED, ex);
                     appIdList = "";
--- a/src/src/com/microsoft/aad/adal/AuthenticationActivity.java
+++ b/src/src/com/microsoft/aad/adal/AuthenticationActivity.java
@@ -814,7 +814,7 @@ public class AuthenticationActivity extends Activity {
             try {
                 result.taskResult = oauthRequest.getToken(urlItems[0]);
                 Logger.v(TAG, "TokenTask processed the result. " + mRequest.getLogInfo());
-            } catch (IOException | UnexpectedServerResponseException exc) {
+            } catch (IOException | AuthenticationServerProtocolException exc) {
                 Logger.e(TAG, "Error in processing code to get a token. " + mRequest.getLogInfo(),
                         "Request url:" + urlItems[0],
                         ADALError.AUTHORIZATION_CODE_NOT_EXCHANGED_FOR_TOKEN, exc);
--- a/src/src/com/microsoft/aad/adal/AuthenticationContext.java
+++ b/src/src/com/microsoft/aad/adal/AuthenticationContext.java
@@ -793,7 +793,7 @@ public class AuthenticationContext {
                                     result = oauthRequest.getToken(endingUrl);
                                     Logger.v(TAG, "OnActivityResult processed the result. "
                                             + authenticationRequest.getLogInfo());
-                                } catch (IOException | UnexpectedServerResponseException exc) {
+                                } catch (IOException | AuthenticationServerProtocolException exc) {
                                     String msg = "Error in processing code to get token. "
                                             + authenticationRequest.getLogInfo() + correlationInfo;
                                     Logger.e(TAG, msg,
@@ -1595,7 +1595,7 @@ public class AuthenticationContext {
                 Logger.v(TAG, "Refresh token is not returned or empty");
                 result.setRefreshToken(refreshItem.mRefreshToken);
             }
-        } catch (IOException | UnexpectedServerResponseException exc) {
+        } catch (IOException | AuthenticationServerProtocolException exc) {
             // Server side error or similar
             Logger.e(TAG, "Error in refresh token for request:" + request.getLogInfo(),
                     ExceptionExtensions.getExceptionMessage(exc), ADALError.AUTH_FAILED_NO_TOKEN,
--- a/src/src/com/microsoft/aad/adal/AuthenticationContext.java
+++ b/src/src/com/microsoft/aad/adal/AuthenticationContext.java
@@ -793,7 +793,7 @@ public class AuthenticationContext {
                                     result = oauthRequest.getToken(endingUrl);
                                     Logger.v(TAG, "OnActivityResult processed the result. "
                                             + authenticationRequest.getLogInfo());
-                                } catch (Exception exc) {
+                                } catch (IOException exc) {
                                     String msg = "Error in processing code to get token. "
                                             + authenticationRequest.getLogInfo() + correlationInfo;
                                     Logger.e(TAG, msg,
@@ -1100,25 +1100,17 @@ public class AuthenticationContext {
         }
 
         if (mValidateAuthority && !mAuthorityValidated) {
-            try {
-                final URL authorityUrlInCallback = authorityUrl;
-                // Discovery call creates an Async Task to send
-                // Web Requests
-                // using a handler
-                boolean result = validateAuthority(authorityUrl);
-                if (result) {
-                    mAuthorityValidated = true;
-                    Logger.v(TAG, "Authority is validated: " + authorityUrlInCallback.toString());
-                } else {
-                    Logger.v(TAG, "Call external callback since instance is invalid"
-                            + authorityUrlInCallback.toString());
-                    callbackHandle.onError(new AuthenticationException(
-                            ADALError.DEVELOPER_AUTHORITY_IS_NOT_VALID_INSTANCE));
-                    return null;
-                }
-            } catch (Exception exc) {
-                Logger.e(TAG, "Authority validation has an error.", "",
-                        ADALError.DEVELOPER_AUTHORITY_IS_NOT_VALID_INSTANCE, exc);
+            final URL authorityUrlInCallback = authorityUrl;
+            // Discovery call creates an Async Task to send
+            // Web Requests
+            // using a handler
+            boolean result = validateAuthority(authorityUrl);
+            if (result) {
+                mAuthorityValidated = true;
+                Logger.v(TAG, "Authority is validated: " + authorityUrlInCallback.toString());
+            } else {
+                Logger.v(TAG, "Call external callback since instance is invalid"
+                        + authorityUrlInCallback.toString());
                 callbackHandle.onError(new AuthenticationException(
                         ADALError.DEVELOPER_AUTHORITY_IS_NOT_VALID_INSTANCE));
                 return null;
@@ -1603,7 +1595,7 @@ public class AuthenticationContext {
                 Logger.v(TAG, "Refresh token is not returned or empty");
                 result.setRefreshToken(refreshItem.mRefreshToken);
             }
-        } catch (Exception exc) {
+        } catch (IOException exc) {
             // Server side error or similar
             Logger.e(TAG, "Error in refresh token for request:" + request.getLogInfo(),
                     ExceptionExtensions.getExceptionMessage(exc), ADALError.AUTH_FAILED_NO_TOKEN,
@@ -1672,15 +1664,9 @@ public class AuthenticationContext {
 
             // Set CorrelationId for Instance Discovery
             mDiscovery.setCorrelationId(getRequestCorrelationId());
-            try {
-                boolean result = mDiscovery.isValidAuthority(authorityUrl);
-                Logger.v(TAG, "Finish validating authority:" + authorityUrl + " result:" + result);
-                return result;
-            } catch (Exception exc) {
-                Logger.e(TAG, "Instance validation returned error", "",
-                        ADALError.DEVELOPER_AUTHORITY_CAN_NOT_BE_VALIDED, exc);
-
-            }
+            boolean result = mDiscovery.isValidAuthority(authorityUrl);
+            Logger.v(TAG, "Finish validating authority:" + authorityUrl + " result:" + result);
+            return result;
         }
         return false;
     }
@@ -1828,26 +1814,15 @@ public class AuthenticationContext {
                 if (mValidateAuthority) {
                     Logger.v(TAG, "Validating authority");
 
-                    try {
-                        if (validateAuthority(authorityUrl)) {
-                            Logger.v(TAG, "Authority is validated" + authorityUrl.toString());
-                        } else {
-                            Logger.v(
-                                    TAG,
-                                    "Call callback since instance is invalid:"
-                                            + authorityUrl.toString());
-                            callbackHandle.onError(new AuthenticationException(
-                                    ADALError.DEVELOPER_AUTHORITY_IS_NOT_VALID_INSTANCE));
-                            return;
-                        }
-                    } catch (Exception exc) {
-                        Logger.e(TAG, "Authority validation is failed",
-                                ExceptionExtensions.getExceptionMessage(exc),
-                                ADALError.SERVER_INVALID_REQUEST, exc);
-                        callbackHandle
-                                .onError(new AuthenticationException(
-                                        ADALError.SERVER_INVALID_REQUEST,
-                                        "Authority validation is failed"));
+                    if (validateAuthority(authorityUrl)) {
+                        Logger.v(TAG, "Authority is validated" + authorityUrl.toString());
+                    } else {
+                        Logger.v(
+                                TAG,
+                                "Call callback since instance is invalid:"
+                                        + authorityUrl.toString());
+                        callbackHandle.onError(new AuthenticationException(
+                                ADALError.DEVELOPER_AUTHORITY_IS_NOT_VALID_INSTANCE));
                         return;
                     }
                 }
--- a/src/src/com/microsoft/aad/adal/AuthenticationContext.java
+++ b/src/src/com/microsoft/aad/adal/AuthenticationContext.java
@@ -793,7 +793,7 @@ public class AuthenticationContext {
                                     result = oauthRequest.getToken(endingUrl);
                                     Logger.v(TAG, "OnActivityResult processed the result. "
                                             + authenticationRequest.getLogInfo());
-                                } catch (IOException exc) {
+                                } catch (IOException | UnexpectedServerResponseException exc) {
                                     String msg = "Error in processing code to get token. "
                                             + authenticationRequest.getLogInfo() + correlationInfo;
                                     Logger.e(TAG, msg,
@@ -1595,7 +1595,7 @@ public class AuthenticationContext {
                 Logger.v(TAG, "Refresh token is not returned or empty");
                 result.setRefreshToken(refreshItem.mRefreshToken);
             }
-        } catch (IOException exc) {
+        } catch (IOException | UnexpectedServerResponseException exc) {
             // Server side error or similar
             Logger.e(TAG, "Error in refresh token for request:" + request.getLogInfo(),
                     ExceptionExtensions.getExceptionMessage(exc), ADALError.AUTH_FAILED_NO_TOKEN,
--- a/src/src/com/microsoft/aad/adal/BasicWebViewClient.java
+++ b/src/src/com/microsoft/aad/adal/BasicWebViewClient.java
@@ -183,24 +183,6 @@ abstract class BasicWebViewClient extends WebViewClient {
                                 view.loadUrl(loadUrl, headers);
                             }
                         });
-                    } catch (IllegalArgumentException e) {
-                        Logger.e(TAG, "Argument exception", e.getMessage(),
-                                ADALError.ARGUMENT_EXCEPTION, e);
-                        // It should return error code and finish the
-                        // activity, so that onActivityResult implementation
-                        // returns errors to callback.
-                        Intent resultIntent = new Intent();
-                        resultIntent.putExtra(
-                                        AuthenticationConstants.Browser.RESPONSE_AUTHENTICATION_EXCEPTION,
-                                        e);
-                        if (mRequest != null) {
-                            resultIntent.putExtra(
-                                    AuthenticationConstants.Browser.RESPONSE_REQUEST_INFO,
-                                    mRequest);
-                        }
-                        sendResponse(
-                                AuthenticationConstants.UIResponse.BROWSER_CODE_AUTHENTICATION_EXCEPTION,
-                                resultIntent);
                     } catch (AuthenticationException e) {
                         Logger.e(TAG, "It is failed to create device certificate response",
                                 e.getMessage(), ADALError.DEVICE_CERTIFICATE_RESPONSE_FAILED, e);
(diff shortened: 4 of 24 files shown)


src/src/com/microsoft/aad/adal/AuthenticationActivity.java
```diff
@@ -790,7 +785,7 @@ protected TokenTaskResult doInBackground(String... urlItems) {
                 // Record account in the AccountManager service
                 try {
                     setAccount(result);
-                } catch (Exception exc) {
+                } catch (GeneralSecurityException | IOException exc) {
```
weijjia: You should have this in different catches, and should we log out the exception we got in the log.

src/src/com/microsoft/aad/adal/AuthenticationActivity.java
```diff
@@ -826,7 +821,7 @@ private void appendAppUIDToAccount(StorageHelper cryptoHelper, Account account)
             } else {
                 try {
                     appIdList = cryptoHelper.decrypt(appIdList);
-                } catch (Exception ex) {
+                } catch (GeneralSecurityException | IOException ex) {
```
weijjia: same as above.

src/src/com/microsoft/aad/adal/BasicWebViewClient.java
```diff
@@ -183,24 +183,6 @@ public void run() {
...
-                        Intent resultIntent = new Intent();
-                        resultIntent.putExtra(
-                                        AuthenticationConstants.Browser.RESPONSE_AUTHENTICATION_EXCEPTION,
-                                        e);
-                        if (mRequest != null) {
-                            resultIntent.putExtra(
-                                    AuthenticationConstants.Browser.RESPONSE_REQUEST_INFO,
-                                    mRequest);
-                        }
-                        sendResponse(
```
weijjia: The removal here I think it would potential cause activity leak problem. And I don't think the catch for exception is a problem here. We return the exception back to caller. it's not swallowing any errors.

src/src/com/microsoft/aad/adal/BasicWebViewClient.java
```diff
@@ -219,20 +201,7 @@ public void run() {
...
-                        Intent resultIntent = new Intent();
-                        resultIntent.putExtra(
-                                        AuthenticationConstants.Browser.RESPONSE_AUTHENTICATION_EXCEPTION,
-                                        e);
-                        if (mRequest != null) {
-                            resultIntent.putExtra(
-                                    AuthenticationConstants.Browser.RESPONSE_REQUEST_INFO,
-                                    mRequest);
                         }
-                        sendResponse(
```
weijjia: Same as above.

src/src/com/microsoft/aad/adal/AuthenticationActivity.java
imakhalova: Why do you need different catches and duplicated code if java supports multi-catch?

src/src/com/microsoft/aad/adal/BasicWebViewClient.java
imakhalova: Do you know which method should throw this IllegalArgumentException ?
What user should do when he gets it?
Crash and let the developer to fix his request

src/src/com/microsoft/aad/adal/BasicWebViewClient.java
imakhalova: same as above

src/src/com/microsoft/aad/adal/BasicWebViewClient.java
agamemnum: Catching IllegalArgumentException is generally should be avoided - we are aiming for a crash in here... So instead of the leak we would see app crash now.

The Java Programming Language, by Gosling, Arnold, and Holmes: "Unchecked run-time exceptions represent conditions that, generally speaking, reflect errors in your program's logic and cannot be reasonably recovered from at run time."

weijjia: This illegalArgumentException is thrown when we process the challenge returned from server. I understand we should generally avoid catching illegalArgumentException especially in the case when checking for passed in arguments. But here is the case we hit conditional access flow, server send back the challenge, and adal need to process the challenge. We need to make sure the coming 401 challenge contains "headervalue". This has nothing to do with the client request, could potentially be the server error, do you still think we should crash and let developer fix his request? If we send back the error, then we know what's going wrong and will have the correlation id, and may know what's going wrong with the broker challenge. I agree the thrown exception could be changed though, it doesn't make a lot of sense to throw illegalArgument exception here.

src/src/com/microsoft/aad/adal/AuthenticationActivity.java
weijjia: I know java supports multi-catch, but I also want it to be super clear in the log that we can tell what is the exception thrown. I don't mind having a couple of more lines by just logging 'GeneralSecurityException is thrown'

src/src/com/microsoft/aad/adal/BasicWebViewClient.java
agamemnum: What about adding a new exception type - `IllegalServiceResponseException`?
And we would basically update below to use it?
   `  {
            throw new IllegalArgumentException("headerValue");
        }
`

weijjia: That works for me. @RandalliLama @heidijinxujia Do you have any concern on this?

src/src/com/microsoft/aad/adal/AuthenticationActivity.java
agamemnum: But we pass exception object into the logger, where we would print everything we need. We should be able to get exact exception, message and so on in the log. Does this address your concern?

weijjia: Well this is true if Logger.getInstance.setAndroidLogEnabled is not set to be false. Also, If externalLogger is used, we're losing the stacktrace.

agamemnum: Sorry I'm not sure I understand. Do you prefer to write a message that clearly indicates a type of error in the syslog directly? I thought the reason why Logger was created is to make all tracing logic centralized (especially around exceptions and errors)... If we need to improve Logger we can do it in separate pull request. Can you please provide an example of desired code and some explanation on how it would help to debug an issue?

weijjia: Here is our current way to log errors
```
    public void error(String tag, String message, String additionalMessage, ADALError errorCode) {
        if (mAndroidLogEnabled) {
            Log.e(tag, getLogMessage(message, additionalMessage, errorCode));
        }

        logCommon(tag, message, additionalMessage, LogLevel.Error, errorCode);
    }
```
if mAndroidLogEnabled is set to be false, Log.e won't be called. We won't know the exact exception thrown. i.e For the following decrption of appIdList, we're catching the GeneralSecurityException and IOException, and we log like "appUIDList failed to decrypt" and then set appIdList to be empty. This is the broker flow, and when we next time trying to find a matching cache item from broker, this could cause failure of uid check in broker. This is basically why i also just want us the log the exception name and exception message in the log message itself. 

I also agree we should improve Logger, we already passed in the exception today, Logger should be improved to handle this.

weijjia:
@heidijinxujia This is the exception catching change @imakhalova made, i thought you also make similar changes, have you already done with that?

src/src/com/microsoft/aad/adal/AuthenticationActivity.java
agamemnum: So if in the code below:
````
   public void error(String tag, String message, String additionalMessage, ADALError errorCode,
            Throwable err) {
        if (mAndroidLogEnabled) {
            Log.e(tag, getLogMessage(message, additionalMessage, errorCode), err);
        }

        logCommon(tag, message, additionalMessage, LogLevel.Error, errorCode);
    }

```
we append exception information to  `additionalMessage` would it resolve your concerns?

weijjia: yup, that will be nice. Thanks.

imakhalova: {AncestorPath = 53355657}
It actually used another error method which accepts exceptions as last argument and appends it to log line.

src/src/com/microsoft/aad/adal/AuthenticationActivity.java
```diff
@@ -776,7 +771,7 @@ protected TokenTaskResult doInBackground(String... urlItems) {
             try {
                 result.taskResult = oauthRequest.getToken(urlItems[0]);
                 Logger.v(TAG, "TokenTask processed the result. " + mRequest.getLogInfo());
-            } catch (Exception exc) {
+            } catch (IOException exc) {
```
heidijinxujia: Should it also catch the IllegalArgumentException and AuthenticationException throwed by getToken()?

src/src/com/microsoft/aad/adal/AuthenticationActivity.java
heidijinxujia: I think we should log the specific exception out of GeneralSecurityException. @weijjia

src/src/com/microsoft/aad/adal/AuthenticationContext.java
```diff
@@ -793,7 +793,7 @@ public void run() {
                                     result = oauthRequest.getToken(endingUrl);
                                     Logger.v(TAG, "OnActivityResult processed the result. "
                                             + authenticationRequest.getLogInfo());
-                                } catch (Exception exc) {
+                                } catch (IOException exc) {
```
heidijinxujia: Same here. Should it also catch the IllegalArgumentException and AuthenticationException throwed by getToken()? @weijjia

src/src/com/microsoft/aad/adal/AuthenticationContext.java
```diff
@@ -1603,7 +1595,7 @@ private AuthenticationResult refreshToken(final CallbackHandler callbackHandle,
                 Logger.v(TAG, "Refresh token is not returned or empty");
                 result.setRefreshToken(refreshItem.mRefreshToken);
             }
-        } catch (Exception exc) {
+        } catch (IOException exc) {
```
heidijinxujia: + IllegalArgumentException ?

src/src/com/microsoft/aad/adal/AuthenticationActivity.java
agamemnum: @heidijinxujia What do you like to log for `GeneralSecurityException` in there? Will changes in the `Logger.error` would address your concerns here too?

src/src/com/microsoft/aad/adal/AuthenticationContext.java
```diff
@@ -1672,15 +1664,9 @@ private boolean validateAuthority(final URL authorityUrl) {
...
-            try {
-                boolean result = mDiscovery.isValidAuthority(authorityUrl);
-                Logger.v(TAG, "Finish validating authority:" + authorityUrl + " result:" + result);
-                return result;
-            } catch (Exception exc) {
-                Logger.e(TAG, "Instance validation returned error", "",
-                        ADALError.DEVELOPER_AUTHORITY_CAN_NOT_BE_VALIDED, exc);
-
-            }
+            boolean result = mDiscovery.isValidAuthority(authorityUrl);
```
heidijinxujia: isValidAuthority might throw AuthenticationException.

src/src/com/microsoft/aad/adal/AuthenticationActivity.java
imakhalova: I see what you mean, but I don't want to cover it in this PR

src/pom.xml
```diff
@@ -179,8 +179,8 @@
         <artifactId>maven-compiler-plugin</artifactId>
         <version>3.1</version>
         <configuration>
-          <source>1.6</source>
-          <target>1.6</target>
+          <source>1.7</source>
+          <target>1.7</target>
```
imakhalova: test comment from Codeflow

src/src/com/microsoft/aad/adal/AuthenticationActivity.java
imakhalova: Nope, let these type of exceptions fall through, because they don't depend on user and if they happen should be fixed by dev

src/src/com/microsoft/aad/adal/AuthenticationContext.java
agamemnum: @heidijinxujia we should avoid catching `IllegalArgumentException` that's why we introduced new error type for server errors to avoid any catching of this type.

src/src/com/microsoft/aad/adal/AuthenticationActivity.java
heidijinxujia: OK. Thanks. :)

src/src/com/microsoft/aad/adal/StorageHelper.java
```diff
@@ -466,7 +466,7 @@ final synchronized private SecretKey getSecretKeyFromAndroidKeyStore() throws IO
             final byte[] encryptedKey = readKeyData(keyFile);
             sSecretKeyFromAndroidKeyStore = unwrap(wrapCipher, encryptedKey);
             Logger.v(TAG, "Finished reading SecretKey");
-        } catch (Exception ex) {
+        } catch (GeneralSecurityException ex) {
```
weijjia: readKeyData would throw out the IOException, I think we should also catch the exception here.

src/src/com/microsoft/aad/adal/StorageHelper.java
```diff
@@ -466,7 +466,7 @@ final synchronized private SecretKey getSecretKeyFromAndroidKeyStore() throws IO
...
             // Reset KeyPair info so that new request will generate correct KeyPairs.
             // All tokens with previous SecretKey are not possible to decrypt.
             Logger.e(TAG, "Unwrap failed for AndroidKeyStore", "", ADALError.ANDROIDKEYSTORE_FAILED);
```
weijjia: @imakhalova I saw your email mentioning there are many decryption failure from android keystore, and i saw the unwrap failure you mentioned. We don't even have the exception passed into the logger, then there is no way we know the exception stacktrace. You don't need to worry about it in the PR, I'm going through the whole encryption/decryption logic. I will make the change and send another PR.

src/src/com/microsoft/aad/adal/StorageHelper.java
imakhalova: Right, let's add that

src/src/com/microsoft/aad/adal/StorageHelper.java
imakhalova: I'll add exception object as argument to logger

src/src/com/microsoft/aad/adal/AuthenticationContext.java
imakhalova: This is kind of IlligalArgumentException for dev, let app fail with it

src/src/com/microsoft/aad/adal/AuthenticationActivity.java
RandalliLama: Do we already have an issue filed to address this?

agamemnum: Done and merged already.

src/src/com/microsoft/aad/adal/AuthenticationContext.java
RandalliLama: @imakhalova  This is not the case.  There is no reason that a dev could not choose to use different authorities on every call.  It's hard to say where they may get those authorities, but the fact that one of them may be invalid should not cause the app to fall over.

RandalliLama: What that means is the app should be able to catch and recover if it so chooses.  That implies that the exception that ADAL lets through should indicate in a meaningful way the cause of the problem.  If that is what is going on in this change then I am fine with it.

agamemnum: @imakhalova let's make more concrete catch in here, instead removing it completely. That can be addressed later.

RandalliLama: Why not address it now?

agamemnum: @RandalliLama what would you like us to do in there? we can go and fix isValidAuthority to return a boolean instead of throw. Would this work? I basically propose to replace `catch(Exception)` with `catch(AuthenticationException)` for now.

RandalliLama: @agamemnum Sorry, I think that I misunderstood you.  Catching the AuthenticationException and returning false in those cases is the correct thing to do.  If that's what you are proposing that we do in this PR then I am in agreement.

src/src/com/microsoft/aad/adal/UnexpectedServerResponseException.java
```diff
@@ -0,0 +1,14 @@
+package com.microsoft.aad.adal;
+
+import java.io.IOException;
+
+/**
+ * Created by imakhal on 2/17/2016.
+ */
+public class UnexpectedServerResponseException extends IOException {
```
RandalliLama: The definition of IOException is "This class is the general class of exceptions produced by failed or interrupted I/O operations."  That's not how UnexpectedServerResponseException is being used above.  It is being used at least in some cases where we simply got values back from the server that the code doesn't know how to handle.  That's not an I/O problem.  I'll have to think about this some more this evening to come up with a better suggestion.

agamemnum: @imakhalova it's a good point. Why did you inherited it from the `IOException`?

src/src/com/microsoft/aad/adal/AuthenticationContext.java
imakhalova: @RandalliLama Sure I can return false and log an error, but I see that passed null authority will throw IlligalArgumantException...

src/src/com/microsoft/aad/adal/UnexpectedServerResponseException.java
imakhalova: My logic was that this is exception that sent from network calls and indicates type of error during network call. And I think this is actually failure of IO operation/network call.
But I don't mind to inherit from Exception.

RandalliLama: I propose that this exception be renamed to ResourceAuthenticationChallengeException.  The goal is to make the name of the exception abstract enough that it could potentially be used for multiple causes, but specific enough that the developer can reasonably apply the same remediation in all cases that the exception is thrown.  The name UnexpectedServerResponseException achieves the first goal, but I don't think that it achieves the second goal.

I would also make the ResourceAuthenticationChallengeException extend the AuthenticationException class instead of the IOException class.  I'd like most, if not all, exceptions thrown by ADAL to be ADAL defined exceptions.  This will make it simpler to maintain a consistent way of collecting telemetry on errors when we introduce a more consistent way of handling sub-errors.  For instance, all Authentication exceptions have an ADALError typed property that can  contain very specific information about the exact cause of the exception in an easily collected enum.

RandalliLama: I just noticed that this exception is being used for at least two different types of servers and their responses.  For the device challenge response I think we need a separate exception that is also derived from AuthenticationException.  I would call it AuthenticationServerProtocolException.  In a subsequent PR all OAuth and other protocol level errors should use this new exception.

src/src/com/microsoft/aad/adal/AuthenticationContext.java
RandalliLama: We should only be throwing IllegalArgumentException when doing parameter validation in public methods.

Catching AuthenticationException makes sense if it needs to be transferred to another thread.

src/src/com/microsoft/aad/adal/BasicWebViewClient.java
```diff
@@ -184,7 +185,7 @@ public void run() {
                                 view.loadUrl(loadUrl, headers);
                             }
                         });
-                    } catch (IllegalArgumentException e) {
+                    } catch (UnexpectedServerResponseException e) {
```
RandalliLama: Should be AuthenticationServerProtocolException

src/src/com/microsoft/aad/adal/ChallangeResponseBuilder.java
```diff
@@ -85,13 +85,14 @@ public String getAuthorizationHeaderValue() {
...
-    public ChallangeResponse getChallangeResponseFromUri(final String redirectUri) {
+    public ChallangeResponse getChallangeResponseFromUri(final String redirectUri)
+            throws UnexpectedServerResponseException {
         ChallangeRequest request = getChallangeRequest(redirectUri);
         return getDeviceCertResponse(request);
     }
 
     public ChallangeResponse getChallangeResponseFromHeader(final String challangeHeaderValue,
-            final String endpoint) throws UnsupportedEncodingException {
+            final String endpoint) throws UnsupportedEncodingException, UnexpectedServerResponseException {
```
RandalliLama: Should be AuthenticationServerProtocolException

src/src/com/microsoft/aad/adal/HttpWebRequest.java
```diff
@@ -138,7 +138,7 @@ public HttpWebResponse send() {
                 mConnection.setUseCaches(mUseCaches);
                 mConnection.setRequestMethod(mRequestMethod);
                 mConnection.setDoInput(true); // it will at least read status
-                                              // code. Default is true.
+                // code. Default is true.
```
RandalliLama: Did you intentionally change the whitespace here?

src/src/com/microsoft/aad/adal/HttpWebRequest.java
```diff
@@ -182,8 +182,10 @@ public HttpWebResponse send() {
                 Logger.v(TAG, "Response is received");
                 response.setBody(responseBody);
                 response.setResponseHeaders(mConnection.getHeaderFields());
-            } catch (Exception e) {
-                Logger.e(TAG, "Exception:" + e.getMessage(), " Method:" + mRequestMethod,
+            } catch (InterruptedException ignore) {
```
RandalliLama: While we are here, let's change the name of this variable from 'ignore' to 'e'.  There's not reason for it to be different from the other exception variable names and the difference seems to imply there is funny going on in this catch block.

src/src/com/microsoft/aad/adal/Oauth2.java
```diff
@@ -516,12 +517,12 @@ private AuthenticationResult postMessage(String requestMessage, HashMap<String,
 
                 Logger.v(TAG, "Server error message:" + errMessage);
                 if (response.getResponseException() != null) {
-                    throw response.getResponseException();
+                    throw new IOException(response.getResponseException());
```
RandalliLama: Should be AuthenticationServerProtocolException.  There is no reason for the app to crash in this case.

src/src/com/microsoft/aad/adal/Oauth2.java
```diff
@@ -516,12 +517,12 @@ private AuthenticationResult postMessage(String requestMessage, HashMap<String,
...
                 }
             } else {
                 ClientMetrics.INSTANCE.setLastErrorCodes(result.getErrorCodes());
             }
-        } catch (IllegalArgumentException e) {
+        } catch (UnexpectedServerResponseException e) {
```
RandalliLama: Should be AuthenticationServerProtocolException

src/src/com/microsoft/aad/adal/UnexpectedServerResponseException.java
imakhalova: What I don't like in AuthenticationException is that it's derived from RuntimeException and it's really hard to track where this Exception been thrown and where it needs try-catch block.

src/src/com/microsoft/aad/adal/Oauth2.java
```diff
@@ -614,18 +615,13 @@ private AuthenticationResult processTokenResponse(HttpWebResponse webResponse) {
...
-                if (!correlation.equals(mRequest.getCorrelationId())) {
-                    Logger.w(TAG, "CorrelationId is not matching", "",
-                            ADALError.CORRELATION_ID_NOT_MATCHING_REQUEST_RESPONSE);
-                }
-
-                Logger.v(TAG, "Response correlationId:" + correlationIdInHeader);
-            } catch (Exception ex) {
-                Logger.e(TAG, "Wrong format of the correlation ID:" + correlationIdInHeader, "",
-                        ADALError.CORRELATION_ID_FORMAT, ex);
+            UUID correlation = UUID.fromString(correlationIdInHeader);
```
RandalliLama: This shouldn't be fatal.  In fact, it should only be logged as a warning.  We should reverse the logic.

1. mRequest.getCorrelationId() should be converted to a string.
2. That string should be compared to the string correlation id returned from the server.
3. If they are not equal then we can log a warning that contains the correlation id returned from the server.

The correlation id is pure diagnostics and the fact that something unexpected might happen with the correlation id should not affect the success or failure of a request in any way.  Changing the logic to the above will ensure that no exceptions are thrown unless it is truly an exceptional case.  If it were me, I would also move this code out to a separate method named something like 'warnOnCorrelationIdMismatch', but I'm not particularly passionate about it.

src/src/com/microsoft/aad/adal/HttpWebRequest.java
imakhalova: Nope, android studio did it for me, I'll revert it

src/src/com/microsoft/aad/adal/AuthenticationContext.java
RandalliLama: @imakhalova I think that throwing IllegalArgumentException is appropriate from a public developer facing method is the appropriate thing to do when the argument is clearly Illegal.  I think the fact that the passed in authority is null or not a URI should be separated from the case where it is legally formed but not an a trusted authority, which is really what this method is trying to determine.  The fact that is null or not a uri should have been checked at that point that it was passed in originally though.  If that is being re-checked here, we should just remove that redundant check.

src/src/com/microsoft/aad/adal/UnexpectedServerResponseException.java
RandalliLama: I agree completely.  That was a complete f$%^ up.  I think it would be worth changing this.  It is a breaking change, but it is a compile time breaking change rather than a runtime breaking change so it shouldn't change the behavior of existing apps.

src/src/com/microsoft/aad/adal/Oauth2.java
imakhalova: I'm not sure about your concern. The comparison and warning are stayed at their place. I just remove try-catch block since nothing actually throws any excepted exception here

src/src/com/microsoft/aad/adal/UnexpectedServerResponseException.java
imakhalova: @RandalliLama So are we agreed on swiching AuthenticationException to extend Exception instead of RuntimeException?

RandalliLama: We are agreed.  Please proceed to switch the inheritance hierarchy of AuthenticationException.

imakhalova: Ok, I'll submit new PR, this one got pretty heavy

src/src/com/microsoft/aad/adal/Oauth2.java
RandalliLama: UUID.fromString can throw IllegalArgumentException.  There's no reason that we should be attempting to convert the correlation id from the server since there's no reason that it has to be a string.

That's no a bug that you introduced, but removing the catch all exceptions clause will allow a new, potentially app crashing failure mode that didn't exist anymore.  You change is correct.  But now we should deal with the failure mode that shouldn't have been there in the first place.

RandalliLama: If you just want a quicker fix, then catching the IllegalArgumentException would work as well.

imakhalova: Will this approach work? Or do we really want to see the error that format is wrong?
            if (!correlationIdInHeader.equals(mRequest.getCorrelationId())) {
                Logger.w(TAG, "CorrelationId is not matching",
                        "response = " + correlationIdInHeader + " request = " + mRequest.getCorrelationId(),
                        ADALError.CORRELATION_ID_NOT_MATCHING_REQUEST_RESPONSE);
            }
...

imakhalova: Anyway I noticed that we implicitly catch IllegalArgumentException  in other places where we parse this GUID, so let's keep it consistent.
I don't want to do refactoring within this PR.

RandalliLama: Does that mean you are going to revert this particular change?  If so I will file an issue to track the need to remove any exceptions throwing if the returned correlation id is not a guid.

imakhalova: Yes I will revert, but catching only IllegalArgumentException instead of generic Exception

RandalliLama: Sounds good.

RandalliLama: Added #511

src/src/com/microsoft/aad/adal/HttpWebRequest.java
imakhalova: Ok, I'll rename it to e, but this Exception needs to be ignores. it only happens if Thread.sleep got interruption and we don't do anything about it and this method is only for debug purposes there (simulate slow networks)

src/src/com/microsoft/aad/adal/Oauth2.java
imakhalova: This won't crash, it's not RuntimeException. And the only exception from response is actually IOException (see HttpWebRequest.send). Instead of throwing any exception it's better to be specific

src/src/com/microsoft/aad/adal/Oauth2.java
imakhalova: renamed

imakhalova: This catch can actually be removed

src/src/com/microsoft/aad/adal/ChallangeResponseBuilder.java
imakhalova: then I don't see where should be ResourceAuthenticationChallengeException then?

src/src/com/microsoft/aad/adal/Oauth2.java
RandalliLama: Even better.

src/src/com/microsoft/aad/adal/ChallangeResponseBuilder.java
RandalliLama: ResourceAuthenticationChallengeException should be thrown from the AuthenticationParameters class.  That class is about parsing challenges that come from resources.  This class is about parsing device auth challenges that come from the authentication server, not the resource.

src/src/com/microsoft/aad/adal/Oauth2.java
RandalliLama: If the only exceptions that are going to be returned from getResponseException are IOExceptions, why wrap it in another IOException before rethrowing it?

imakhalova: True, but it unsafe just to convert (in case if some other developer starts to set some other exceptions to this variable), let's make this exception implicitly IOException

src/src/com/microsoft/aad/adal/AuthenticationContext.java
imakhalova: Ok, changed Discovery.isValidAuthority to log and return false instead of throwing and catching exception

RandalliLama:
I'm good when @weijjia is good.  :shipit: 

Thank you.

weijjia:
I'm fine with those changes. Thank you @imakhalova

PR #499, opened 2016-02-12
update for null acceptable issuers for client tls request

weijjia:
Should check for null acceptable issuers, if it's null continue with the request, and prompt user for choosing a cert.


RPangrle:
:shipit:

Issue #496, opened 2016-02-11
Support Family of Client IDs


weijjia:
Server will allow client IDs to be grouped on the server side in a way where a RT for one client ID(first part apps) can be redeemed for a AT and RT for a different client ID as long as they're in the same group. This will move us closer to being able to provide SSO-like functionality between apps without requiring the broker (or workplace join).

PR #503, opened 2016-02-17
Add support for family client id #496

weijjia:
1. Add support for family client id #496 
2. Add unit test for family client id support
3. code refactor

### DIFF
--- a/src/src/com/microsoft/aad/adal/AuthenticationContext.java
+++ b/src/src/com/microsoft/aad/adal/AuthenticationContext.java
@@ -237,10 +237,11 @@ public class AuthenticationContext {
                 }
 
                 @Override
-                public Iterator<TokenCacheItem> getAll() {
+                public Iterator<TokenCacheItem> getAll() 
+                {
                     throw new UnsupportedOperationException(
                             "Broker cache does not support direct getAll operation");
-                    }
+                }
             };
         }
         return mTokenCacheStore;
--- a/src/src/com/microsoft/aad/adal/AuthenticationContext.java
+++ b/src/src/com/microsoft/aad/adal/AuthenticationContext.java
@@ -1270,7 +1270,7 @@ public class AuthenticationContext {
 
         // Trying to find refresh token first, if existed, will try to use the refresh token. 
         Logger.v(TAG, "Checking refresh tokens");
-        RefreshItem refreshItem = getRefreshItem(request);
+        RefreshItem refreshItem = getRefreshToken(request);
         AuthenticationResult authResult = null;
         if (!promptUser(request.getPrompt()) && refreshItem != null
                 && !StringExtensions.IsNullOrBlank(refreshItem.mRefreshToken)) {
@@ -1454,7 +1454,7 @@ public class AuthenticationContext {
         }
     }
 
-    private RefreshItem getRefreshItem(final AuthenticationRequest request) 
+    private RefreshItem getRefreshToken(final AuthenticationRequest request) 
     {
         final String methodName = ":getRefreshToken";
         RefreshItem refreshItem = null;
--- a/src/src/com/microsoft/aad/adal/AuthenticationContext.java
+++ b/src/src/com/microsoft/aad/adal/AuthenticationContext.java
@@ -1275,7 +1275,16 @@ public class AuthenticationContext {
         if (!promptUser(request.getPrompt()) && refreshItem != null
                 && !StringExtensions.IsNullOrBlank(refreshItem.mRefreshToken)) {
             Logger.v(TAG, "Refresh token is available and it will attempt to refresh token");
-            authResult = getTokenWithRefreshToken(callbackHandle, activity, useDialog, request, refreshItem, true);
+            try
+            {
+                authResult = getTokenWithRefreshToken(activity, useDialog, request, refreshItem, true);
+            }
+            catch (AuthenticationException authenticationException)
+            {
+                callbackHandle.onError(authenticationException);
+                return null;
+            }
+            
             if (authResult != null && !StringExtensions.IsNullOrBlank(authResult.getAccessToken()))
             {
                 callbackHandle.onSuccess(authResult);
@@ -1679,11 +1688,10 @@ public class AuthenticationContext {
      * @param externalCallback
      * @return
      */
-    private AuthenticationResult getTokenWithRefreshToken(final CallbackHandler callbackHandle,
-            final IWindowComponent activity, final boolean useDialog,
-            final AuthenticationRequest request, final RefreshItem refreshItem,
-            final boolean useCache) {
-
+    private AuthenticationResult getTokenWithRefreshToken(final IWindowComponent activity, final boolean useDialog,
+            final AuthenticationRequest request, final RefreshItem refreshItem, final boolean useCache) 
+            throws AuthenticationException 
+    {
         Logger.v(TAG, "Process refreshToken for " + request.getLogInfo() + " refreshTokenId:"
                 + getTokenHash(refreshItem.mRefreshToken));
 
@@ -1692,13 +1700,13 @@ public class AuthenticationContext {
         // state to not remove refresh token if user turned Airplane mode or
         // similar.
         if (!mConnectionService.isConnectionAvailable()) {
-            AuthenticationException exc = new AuthenticationException(
+            AuthenticationException authenticationException = new AuthenticationException(
                     ADALError.DEVICE_CONNECTION_IS_NOT_AVAILABLE,
                     "Connection is not available to refresh token");
             Logger.w(TAG, "Connection is not available to refresh token", request.getLogInfo(),
                     ADALError.DEVICE_CONNECTION_IS_NOT_AVAILABLE);
-            callbackHandle.onError(exc);
-            return null;
+            
+            throw authenticationException;
         }
 
         AuthenticationResult result = null;
@@ -1718,8 +1726,7 @@ public class AuthenticationContext {
             AuthenticationException authException = new AuthenticationException(
                     ADALError.AUTH_FAILED_NO_TOKEN, ExceptionExtensions.getExceptionMessage(exc),
                     exc);
-            callbackHandle.onError(authException);
-            return null;
+            throw authException;
         }
 
         // useCache will be false when calling from acquireTokenUsingRefreshToken, and if false, need to return
@@ -1764,10 +1771,6 @@ public class AuthenticationContext {
             // error code and error description in
             // Authentication result for Oauth errors
             Logger.v(TAG, "Cache is not used for Request:" + request.getLogInfo());
-            if (callbackHandle.callback != null) 
-            {
-                callbackHandle.onSuccess(result);
-            }
             return result;
         }
     }
@@ -1963,7 +1966,18 @@ public class AuthenticationContext {
 
                 // Follow refresh logic now. Authority is valid or
                 // skipped validation
-                getTokenWithRefreshToken(callbackHandle, null, false, request, refreshItem, false);
+                final AuthenticationResult authResult;
+                try
+                {
+                    authResult = getTokenWithRefreshToken(null, false, request, refreshItem, false);
+                }
+                catch (final AuthenticationException authException)
+                {
+                    callbackHandle.onError(authException);
+                    return;
+                }
+                
+                callbackHandle.onSuccess(authResult);
             }
         });
     }
--- a/src/src/com/microsoft/aad/adal/AuthenticationContext.java
+++ b/src/src/com/microsoft/aad/adal/AuthenticationContext.java
@@ -1288,7 +1288,7 @@ public class AuthenticationContext {
         // if it's silent request, will return the AUTH_REFRESH_FAILED_PROMPT_NOT_ALLOWED.
         if (refreshItem == null || authResult == null || (authResult != null && StringExtensions.IsNullOrBlank(authResult.getAccessToken()))) 
         {
-            Logger.v(TAG, "Refresh token is not available");
+            Logger.v(TAG, "Refresh token is not available or refresh token request failed to return token.");
             if (!request.isSilent() && callbackHandle.callback != null
                     && (activity != null || useDialog)) 
             {
@@ -1481,10 +1481,18 @@ public class AuthenticationContext {
             }
             
             // If still cannot find an item, try with family clientid
+            // Disable this feature for ADFS authority
             if (item == null || StringExtensions.IsNullOrBlank(item.getRefreshToken()))
             {
-                Logger.v(TAG + methodName, "No refresh token found, trying to find family client id item.");
-                item = findFamilyItemForUser(request.getUserIdentifierType(), userId);
+                if (!UrlExtensions.isADFSAuthority(StringExtensions.getUrl(mAuthority)))
+                {
+                    Logger.v(TAG + methodName, "No refresh token found, trying to find family client id item.");
+                    item = findFamilyItemForUser(request.getUserIdentifierType(), userId);
+                }
+                else 
+                {
+                    Logger.v(TAG + methodName, "No refresh token found, skip the family client id item lookup for ADFS authority.");
+                }
             }
 
             if (item != null && !StringExtensions.IsNullOrBlank(item.getRefreshToken())) {
@@ -1529,7 +1537,7 @@ public class AuthenticationContext {
                     if (!StringExtensions.IsNullOrBlank(tokenCacheItem.getFamilyClientId())
                             && !StringExtensions.IsNullOrBlank(tokenCacheItem.getRefreshToken()))
                     {
-                        Logger.v(TAG + methodName, "Found family item matching the given user id.");
+                        Logger.v(TAG + methodName, "Found family item matching the given user id, and the clientId selected for FoCI is: " + tokenCacheItem.getClientId());
                         return tokenCacheItem;
                     }
                 }
@@ -1548,7 +1556,7 @@ public class AuthenticationContext {
                     if (!StringExtensions.IsNullOrBlank(tokenCacheItem.getFamilyClientId())
                             && !StringExtensions.IsNullOrBlank(tokenCacheItem.getRefreshToken()))
                     {
-                        Logger.v(TAG + methodName, "Found family item matching the given user id.");
+                        Logger.v(TAG + methodName, "Found family item matching the given user id, and the clientId selected for FoCI is: " + tokenCacheItem.getClientId());
                         return tokenCacheItem;
                     }
                 }
--- a/src/src/com/microsoft/aad/adal/AuthenticationContext.java
+++ b/src/src/com/microsoft/aad/adal/AuthenticationContext.java
@@ -237,8 +237,7 @@ public class AuthenticationContext {
                 }
 
                 @Override
-                public Iterator<TokenCacheItem> getAll() 
-                {
+                public Iterator<TokenCacheItem> getAll() {
                     throw new UnsupportedOperationException(
                             "Broker cache does not support direct getAll operation");
                 }
@@ -1274,21 +1273,16 @@ public class AuthenticationContext {
         RefreshItem refreshItem = getRefreshToken(request);
         AuthenticationResult authResult = null;
         if (!promptUser(request.getPrompt()) && refreshItem != null
-                && !StringExtensions.IsNullOrBlank(refreshItem.mRefreshToken)) 
-        {
+                && !StringExtensions.IsNullOrBlank(refreshItem.mRefreshToken)) {
             Logger.v(TAG, "Refresh token is available and it will attempt to refresh token");
-            try
-            {
+            try {
                 authResult = getTokenWithRefreshToken(activity, useDialog, request, refreshItem, true);
-            }
-            catch (AuthenticationException authenticationException)
-            {
+            } catch (AuthenticationException authenticationException) {
                 callbackHandle.onError(authenticationException);
                 return null;
             }
             
-            if (authResult != null && !StringExtensions.IsNullOrBlank(authResult.getAccessToken()))
-            {
+            if (authResult != null && !StringExtensions.IsNullOrBlank(authResult.getAccessToken())) {
                 callbackHandle.onSuccess(authResult);
                 return authResult;
             }
@@ -1298,16 +1292,12 @@ public class AuthenticationContext {
         // If it's non-silent request, will try to acquire token interactively. 
         // if it's silent request, will return the AUTH_REFRESH_FAILED_PROMPT_NOT_ALLOWED.
         if (refreshItem == null || authResult == null 
-                || (authResult != null && StringExtensions.IsNullOrBlank(authResult.getAccessToken()))) 
-        {
+                || (authResult != null && StringExtensions.IsNullOrBlank(authResult.getAccessToken()))) {
             Logger.v(TAG, "Refresh token is not available or refresh token request failed to return token.");
             if (!request.isSilent() && callbackHandle.callback != null
-                    && (activity != null || useDialog)) 
-            {
+                    && (activity != null || useDialog)) {
                 acquireTokenInteractively(callbackHandle, activity, request, useDialog);
-            } 
-            else 
-            {
+            } else {
                 final String errorInfo = authResult == null? "" : authResult.getErrorLogInfo();
                 // User does not want to launch activity
                 Logger.e(TAG, "Prompt is not allowed and failed to get token:", request.getLogInfo() + " " + errorInfo,
@@ -1321,11 +1311,9 @@ public class AuthenticationContext {
     }
 
     private void acquireTokenInteractively(final CallbackHandler callbackHandle, final IWindowComponent activity, 
-            final AuthenticationRequest request, final boolean useDialog)
-    {
+            final AuthenticationRequest request, final boolean useDialog) {
         //Check if there is network connection
-        if (!mConnectionService.isConnectionAvailable()) 
-        {
+        if (!mConnectionService.isConnectionAvailable()) {
             AuthenticationException exc = new AuthenticationException(
                     ADALError.DEVICE_CONNECTION_IS_NOT_AVAILABLE,
                     "Connection is not available to request token");
@@ -1347,17 +1335,13 @@ public class AuthenticationContext {
                 new AuthenticationRequestState(callbackHandle.callback.hashCode(), request,
                         callbackHandle.callback));
 
-        if (useDialog) 
-        {
+        if (useDialog) {
             AuthenticationDialog dialog = new AuthenticationDialog(mHandler, mContext,
                     this, request);
             dialog.show();
-        } 
-        else 
-        {
+        } else {
             // onActivityResult will receive the response
-            if (!startAuthenticationActivity(activity, request)) 
-            {
+            if (!startAuthenticationActivity(activity, request)) {
                 callbackHandle.onError(new AuthenticationException(
                         ADALError.DEVELOPER_ACTIVITY_IS_NOT_RESOLVED));
             }
@@ -1472,12 +1456,10 @@ public class AuthenticationContext {
         }
     }
 
-    private RefreshItem getRefreshToken(final AuthenticationRequest request) 
-    {
+    private RefreshItem getRefreshToken(final AuthenticationRequest request) {
         final String methodName = ":getRefreshToken";
         RefreshItem refreshItem = null;
-        if (mTokenCacheStore != null) 
-        {
+        if (mTokenCacheStore != null) {
             boolean multiResource = false;
             // target refreshToken for this resource first. CacheKey will
             // include the resourceId in the cachekey
@@ -1500,15 +1482,11 @@ public class AuthenticationContext {
             
             // If still cannot find an item, try with family clientid
             // Disable this feature for ADFS authority
-            if (item == null || StringExtensions.IsNullOrBlank(item.getRefreshToken()))
-            {
-                if (!UrlExtensions.isADFSAuthority(StringExtensions.getUrl(mAuthority)))
-                {
+            if (item == null || StringExtensions.IsNullOrBlank(item.getRefreshToken())) {
+                if (!UrlExtensions.isADFSAuthority(StringExtensions.getUrl(mAuthority))) {
                     Logger.v(TAG + methodName, "No refresh token found, trying to find family client id item.");
                     item = findFamilyItemForUser(request.getUserIdentifierType(), userId);
-                }
-                else 
-                {
+                } else {
                     Logger.v(TAG + methodName, "No refresh token found, skip the family client id item lookup for ADFS authority.");
                 }
             }
@@ -1525,55 +1503,43 @@ public class AuthenticationContext {
         return refreshItem;
     }
 
-    private TokenCacheItem findFamilyItemForUser(final UserIdentifierType userIdentifierType, final String userId)
-    {
+    private TokenCacheItem findFamilyItemForUser(final UserIdentifierType userIdentifierType, final String userId) {
         final String methodName = ":findFamilyItemForUser";
             
-        if (StringExtensions.IsNullOrBlank(userId))
-        {
+        if (StringExtensions.IsNullOrBlank(userId)) {
             Logger.v(TAG + methodName, "User id is not provided, cannot continue with finding family item.");
             return null;
         }
 
         Iterator<TokenCacheItem> allItems = mTokenCacheStore.getAll();
-        if (allItems == null || !allItems.hasNext())
-        {
+        if (allItems == null || !allItems.hasNext()) {
             Logger.v(TAG + methodName, "No items in the cache, cannot continue with finding family item.");
         }
         
-        if (userIdentifierType == UserIdentifierType.UniqueId)
-        {
+        if (userIdentifierType == UserIdentifierType.UniqueId) {
             Logger.v(TAG + methodName, "UserIdentifier type is unique id, will look for the family item based on unique id.");
-            while (allItems.hasNext())
-            {
+            while (allItems.hasNext()) {
                 final TokenCacheItem tokenCacheItem = allItems.next();
                 final UserInfo tokenUserInfo = tokenCacheItem.getUserInfo();
                 if (tokenUserInfo!= null 
                     && tokenUserInfo.getUserId() != null 
-                    && tokenUserInfo.getUserId().equalsIgnoreCase(userId))
-                {
+                    && tokenUserInfo.getUserId().equalsIgnoreCase(userId)) {
                     if (!StringExtensions.IsNullOrBlank(tokenCacheItem.getFamilyClientId())
-                            && !StringExtensions.IsNullOrBlank(tokenCacheItem.getRefreshToken()))
-                    {
+                            && !StringExtensions.IsNullOrBlank(tokenCacheItem.getRefreshToken())) {
                         Logger.v(TAG + methodName, "Found family item matching the given user id, and the clientId selected for FoCI is: " + tokenCacheItem.getClientId());
                         return tokenCacheItem;
                     }
                 }
             }
-        }
-        else if (userIdentifierType == UserIdentifierType.LoginHint)
-        {
+        } else if (userIdentifierType == UserIdentifierType.LoginHint) {
             Logger.v(TAG + methodName, "UserIdentifier type is loginhint, will look for the family item based on login hint.");
-            while (allItems.hasNext())
-            {
+            while (allItems.hasNext()) {
                 final TokenCacheItem tokenCacheItem = allItems.next();
                 final UserInfo tokenUserInfo = tokenCacheItem.getUserInfo();
                 if (tokenUserInfo != null && tokenUserInfo.getDisplayableId() != null
-                    && tokenUserInfo.getDisplayableId().equalsIgnoreCase(userId))
-                {
+                    && tokenUserInfo.getDisplayableId().equalsIgnoreCase(userId)) {
                     if (!StringExtensions.IsNullOrBlank(tokenCacheItem.getFamilyClientId())
-                            && !StringExtensions.IsNullOrBlank(tokenCacheItem.getRefreshToken()))
-                    {
+                            && !StringExtensions.IsNullOrBlank(tokenCacheItem.getRefreshToken())) {
                         Logger.v(TAG + methodName, "Found family item matching the given user id, and the clientId selected for FoCI is: " + tokenCacheItem.getClientId());
                         return tokenCacheItem;
                     }
@@ -1581,7 +1547,7 @@ public class AuthenticationContext {
             }
         }
 
-    return null;
+        return null;
     }
     
     private void setItemToCache(final AuthenticationRequest request, AuthenticationResult result,
@@ -1699,8 +1665,7 @@ public class AuthenticationContext {
      */
     private AuthenticationResult getTokenWithRefreshToken(final IWindowComponent activity, final boolean useDialog,
             final AuthenticationRequest request, final RefreshItem refreshItem, final boolean useCache) 
-            throws AuthenticationException 
-    {
+            throws AuthenticationException {
         Logger.v(TAG, "Process refreshToken for " + request.getLogInfo() + " refreshTokenId:"
                 + getTokenHash(refreshItem.mRefreshToken));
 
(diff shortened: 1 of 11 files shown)


src/src/com/microsoft/aad/adal/AuthenticationContext.java
```diff
@@ -1261,63 +1268,84 @@ private AuthenticationResult localFlow(CallbackHandler callbackHandle,
...
+        // Trying to find refresh token first, if existed, will try to use the refresh token. 
         Logger.v(TAG, "Checking refresh tokens");
         RefreshItem refreshItem = getRefreshToken(request);
+        AuthenticationResult authResult = null;
         if (!promptUser(request.getPrompt()) && refreshItem != null
                 && !StringExtensions.IsNullOrBlank(refreshItem.mRefreshToken)) {
             Logger.v(TAG, "Refresh token is available and it will attempt to refresh token");
-            return refreshToken(callbackHandle, activity, useDialog, request, refreshItem, true);
-        } else {
+            authResult = refreshToken(callbackHandle, activity, useDialog, request, refreshItem, true);
```
kpanwar: Method name should contain verbs. Consider renaming to GetRefreshToken

src/src/com/microsoft/aad/adal/AuthenticationContext.java
```diff
@@ -1461,6 +1500,63 @@ private RefreshItem getRefreshToken(final AuthenticationRequest request) {
...
+        {
+            Logger.v(TAG + methodName, "UserIdentifier type is loginhint, will look for the family item based on login hint.");
+            while (allItems.hasNext())
+            {
+                final TokenCacheItem tokenCacheItem = allItems.next();
+                final UserInfo tokenUserInfo = tokenCacheItem.getUserInfo();
+                if (tokenUserInfo != null && tokenUserInfo.getDisplayableId() != null
+                    && tokenUserInfo.getDisplayableId().equalsIgnoreCase(userId))
+                {
+                    if (!StringExtensions.IsNullOrBlank(tokenCacheItem.getFamilyClientId()))
```
kpanwar: check for familyClientId can be merged with outer if.

src/src/com/microsoft/aad/adal/AuthenticationContext.java
```diff
@@ -1461,6 +1500,63 @@ private RefreshItem getRefreshToken(final AuthenticationRequest request) {
...
+            Logger.v(TAG + methodName, "UserIdentifier type is unique id, will look for the family item based on unique id.");
+            while (allItems.hasNext())
+            {
+                final TokenCacheItem tokenCacheItem = allItems.next();
+                final UserInfo tokenUserInfo = tokenCacheItem.getUserInfo();
+                if (tokenUserInfo!= null 
+                    && tokenUserInfo.getUserId() != null 
+                    && tokenUserInfo.getUserId().equalsIgnoreCase(userId))
+                {
+                    if (!StringExtensions.IsNullOrBlank(tokenCacheItem.getFamilyClientId()))
```
kpanwar: check for familyClientId can be merged with outer if.

weijjia:
@sdabbiru FYI

src/src/com/microsoft/aad/adal/AuthenticationContext.java
weijjia: i should also check refresh token is not empty.

src/src/com/microsoft/aad/adal/AuthenticationContext.java
```diff
@@ -1261,63 +1268,84 @@ private AuthenticationResult localFlow(CallbackHandler callbackHandle,
             return cachedItem;
         }
 
+        // Trying to find refresh token first, if existed, will try to use the refresh token. 
         Logger.v(TAG, "Checking refresh tokens");
-        RefreshItem refreshItem = getRefreshToken(request);
+        RefreshItem refreshItem = getRefreshItem(request);
+        AuthenticationResult authResult = null;
```
sdabbiru: NIT: Please reverse the order of lines 1273 and 1274.

src/src/com/microsoft/aad/adal/AuthenticationContext.java
```diff
@@ -1261,63 +1268,84 @@ private AuthenticationResult localFlow(CallbackHandler callbackHandle,
...
+                return authResult;
+            }
+        } 
+        
+        // refresh token does not exist or refresh token request failed to give back the token. 
+        // If it's non-silent request, will try to acquire token interactively. 
+        // if it's silent request, will return the AUTH_REFRESH_FAILED_PROMPT_NOT_ALLOWED.
+        if (refreshItem == null || authResult == null || (authResult != null && StringExtensions.IsNullOrBlank(authResult.getAccessToken()))) 
+        {
             Logger.v(TAG, "Refresh token is not available");
```
sdabbiru: The log message needs to be updated.

src/src/com/microsoft/aad/adal/AuthenticationContext.java
```diff
@@ -1448,6 +1479,13 @@ private RefreshItem getRefreshToken(final AuthenticationRequest request) {
                 item = mTokenCacheStore.getItem(keyUsed);
                 multiResource = true;
             }
+            
+            // If still cannot find an item, try with family clientid
+            if (item == null || StringExtensions.IsNullOrBlank(item.getRefreshToken()))
+            {
+                Logger.v(TAG + methodName, "No refresh token found, trying to find family client id item.");
```
kpanwar: This feature should not be enabled for ADFS authority.

src/src/com/microsoft/aad/adal/AuthenticationContext.java
```diff
@@ -1461,6 +1499,65 @@ private RefreshItem getRefreshToken(final AuthenticationRequest request) {
...
+                final TokenCacheItem tokenCacheItem = allItems.next();
+                final UserInfo tokenUserInfo = tokenCacheItem.getUserInfo();
+                if (tokenUserInfo!= null 
+                    && tokenUserInfo.getUserId() != null 
+                    && tokenUserInfo.getUserId().equalsIgnoreCase(userId))
+                {
+                    if (!StringExtensions.IsNullOrBlank(tokenCacheItem.getFamilyClientId())
+                            && !StringExtensions.IsNullOrBlank(tokenCacheItem.getRefreshToken()))
+                    {
+                        Logger.v(TAG + methodName, "Found family item matching the given user id.");
```
kpanwar: It will be helpful to publish what client ID RT is being selected for FOCI.

src/src/com/microsoft/aad/adal/AuthenticationContext.java
```diff
@@ -1461,6 +1499,65 @@ private RefreshItem getRefreshToken(final AuthenticationRequest request) {
...
+            {
+                final TokenCacheItem tokenCacheItem = allItems.next();
+                final UserInfo tokenUserInfo = tokenCacheItem.getUserInfo();
+                if (tokenUserInfo != null && tokenUserInfo.getDisplayableId() != null
+                    && tokenUserInfo.getDisplayableId().equalsIgnoreCase(userId))
+                {
+                    if (!StringExtensions.IsNullOrBlank(tokenCacheItem.getFamilyClientId())
+                            && !StringExtensions.IsNullOrBlank(tokenCacheItem.getRefreshToken()))
+                    {
+                        Logger.v(TAG + methodName, "Found family item matching the given user id.");
```
kpanwar: It will be helpful to publish what client ID RT is being selected for FOCI.

src/src/com/microsoft/aad/adal/AuthenticationContext.java
weijjia: why? we try to find a refresh token first.

imakhalova: getRefreshToken is more accurate for this function, because what it really searches for is refreshToken and not item, even if it returns refreshtoken within item

weijjia: yup, it makes sense. I'm thinking why i change it to getRefreshItem..:(

src/src/com/microsoft/aad/adal/AuthenticationContext.java
```diff
@@ -1461,6 +1507,65 @@ private RefreshItem getRefreshToken(final AuthenticationRequest request) {
...
+            return null;
+        }
+
+        Iterator<TokenCacheItem> allItems = mTokenCacheStore.getAll();
+        if (allItems == null || !allItems.hasNext())
+        {
+            Logger.v(TAG + methodName, "No items in the cache, cannot continue with finding family item.");
+        }
+        
+        if (userIdentifierType == UserIdentifierType.UniqueId)
```
kpanwar: this check will only work when there exists a token for a user with the same tenant. You should search with unique id first (hoping the user signed into the same tenant in another app), if no match is found then get the displayable id of the user with passed in unique id and search again.

kpanwar:
:shipit:

src/src/com/microsoft/aad/adal/AuthenticationContext.java
```diff
@@ -1261,63 +1268,84 @@ private AuthenticationResult localFlow(CallbackHandler callbackHandle,
...
         if (!promptUser(request.getPrompt()) && refreshItem != null
                 && !StringExtensions.IsNullOrBlank(refreshItem.mRefreshToken)) {
             Logger.v(TAG, "Refresh token is available and it will attempt to refresh token");
-            return refreshToken(callbackHandle, activity, useDialog, request, refreshItem, true);
-        } else {
-            Logger.v(TAG, "Refresh token is not available");
+            authResult = getTokenWithRefreshToken(callbackHandle, activity, useDialog, request, refreshItem, true);
+            if (authResult != null && !StringExtensions.IsNullOrBlank(authResult.getAccessToken()))
+            {
+                callbackHandle.onSuccess(authResult);
```
imakhalova: You will invoke callback twice which is really bad thing

src/src/com/microsoft/aad/adal/AuthenticationContext.java
```diff
@@ -1261,63 +1268,84 @@ private AuthenticationResult localFlow(CallbackHandler callbackHandle,
...
-                    // onActivityResult will receive the response
-                    if (!startAuthenticationActivity(activity, request)) {
-                        callbackHandle.onError(new AuthenticationException(
-                                ADALError.DEVELOPER_ACTIVITY_IS_NOT_RESOLVED));
-                    }
-                }
-            } else {
-
+                    && (activity != null || useDialog)) 
+            {
```
imakhalova: which environment are you using, that puts { to the next string...it's quite disturbing that you change code style of library

src/src/com/microsoft/aad/adal/AuthenticationContext.java
```diff
@@ -1261,63 +1268,84 @@ private AuthenticationResult localFlow(CallbackHandler callbackHandle,
...
+                callbackHandle.onSuccess(authResult);
+                return authResult;
+            }
+        } 
+        
+        // refresh token does not exist or refresh token request failed to give back the token. 
+        // If it's non-silent request, will try to acquire token interactively. 
+        // if it's silent request, will return the AUTH_REFRESH_FAILED_PROMPT_NOT_ALLOWED.
+        if (refreshItem == null || authResult == null || (authResult != null && StringExtensions.IsNullOrBlank(authResult.getAccessToken())))
```
imakhalova: please split it into 2 lines, it doesn't fit in screen

src/src/com/microsoft/aad/adal/AuthenticationContext.java
```diff
@@ -1644,19 +1753,19 @@ private AuthenticationResult refreshToken(final CallbackHandler callbackHandle,
...
 
             // User is not using cache and explicitly
             // calling with refresh token. User should received
             // error code and error description in
             // Authentication result for Oauth errors
             Logger.v(TAG, "Cache is not used for Request:" + request.getLogInfo());
-            if (callbackHandle.callback != null) {
+            if (callbackHandle.callback != null) 
+            {
                 callbackHandle.onSuccess(result);
```
imakhalova: I think callback handle checks whether callback is null, no need to double check it

src/src/com/microsoft/aad/adal/AuthenticationContext.java
imakhalova: switch() ?

src/src/com/microsoft/aad/adal/AuthenticationContext.java
```diff
@@ -1461,6 +1507,65 @@ private RefreshItem getRefreshToken(final AuthenticationRequest request) {
...
+            }
+        }
+        else if (userIdentifierType == UserIdentifierType.LoginHint)
+        {
+            Logger.v(TAG + methodName, "UserIdentifier type is loginhint, will look for the family item based on login hint.");
+            while (allItems.hasNext())
+            {
+                final TokenCacheItem tokenCacheItem = allItems.next();
+                final UserInfo tokenUserInfo = tokenCacheItem.getUserInfo();
+                if (tokenUserInfo != null && tokenUserInfo.getDisplayableId() != null
```
imakhalova: Can tokenUserInfo  really be null? do we store null UserInfo in cachestore?

src/src/com/microsoft/aad/adal/AuthenticationContext.java
```diff
@@ -1461,6 +1507,65 @@ private RefreshItem getRefreshToken(final AuthenticationRequest request) {
...
+            Logger.v(TAG + methodName, "UserIdentifier type is unique id, will look for the family item based on unique id.");
+            while (allItems.hasNext())
+            {
+                final TokenCacheItem tokenCacheItem = allItems.next();
+                final UserInfo tokenUserInfo = tokenCacheItem.getUserInfo();
+                if (tokenUserInfo!= null 
+                    && tokenUserInfo.getUserId() != null 
+                    && tokenUserInfo.getUserId().equalsIgnoreCase(userId))
+                {
+                    if (!StringExtensions.IsNullOrBlank(tokenCacheItem.getFamilyClientId())
```
imakhalova: Where do you check if it's the same family as in request?
Client can be from different family

imakhalova:
Please don't merge it yet, I think it's not quite ready for shipping.
@weijjia why did you decide to include refactoring within this PR?
Can you please split the feature and refactoring?

src/src/com/microsoft/aad/adal/AuthenticationContext.java
weijjia: As client, we don't know which family it belongs. FoCI now is only open to the first part apps, we only try to find a token cache item with the FoCI flag, and try with the refresh token.

weijjia:
@imakhalova In today's logic, i.e. if i found a token cache with FoCI flag, i'll use it for refresh token request(we always find the first token cache item). If the refresh token request fails, we go back to the acquiretokenlocal call again.But when we go to the acquiretokenlcoalcall again, we should just return the error back or going through a interactive flow again. If not making any refactor with the current code, i'll need to add a new flag showing, oh, i don't find any FoCI refresh token, return error or going through interactive flow. Does it make sense?

src/src/com/microsoft/aad/adal/AuthenticationContext.java
weijjia: if idtoken is not returned, then userinfo won't exist.

src/src/com/microsoft/aad/adal/AuthenticationContext.java
weijjia: I agree with you. If there is exception thrown, and it's non-silent request, the callback will potentially be called twice. Thanks for catching it.

src/src/com/microsoft/aad/adal/AuthenticationContext.java
weijjia: I prefer to follow this style：
1. 
```
if ()
{
    ....
}
else if ()
{
    ....
}
else
{
    ....
}
```
2. no tabs in the code(tab should be 4 spaces)
3. put a space for all the brackets after i.e. if, else if. 
This makes code so much easier to read. I'm currently changing it for all the code I touch. If I get time, i'll change across files, and make sure the whole library follow the style.

imakhalova: I understand that you like it, but there is java code convention, It is about 20 years old. And this library supports it. So I really wish we keep that.
You can read it on official oracle web site (If-else statements section) 
http://www.oracle.com/technetwork/java/codeconventions-150003.pdf

src/src/com/microsoft/aad/adal/AuthenticationContext.java
```diff
@@ -1261,63 +1269,100 @@ private AuthenticationResult localFlow(CallbackHandler callbackHandle,
...
                 callbackHandle.onError(new AuthenticationException(
-                        ADALError.AUTH_REFRESH_FAILED_PROMPT_NOT_ALLOWED));
+                        ADALError.AUTH_REFRESH_FAILED_PROMPT_NOT_ALLOWED, request.getLogInfo() + " " + errorInfo));
             }
         }
 
         return null;
     }
 
+    private void acquireTokenInteractively(final CallbackHandler callbackHandle, final IWindowComponent activity, final AuthenticationRequest request, final boolean useDialog)
```
imakhalova: If line doesn't fit this view, better to split it

src/src/com/microsoft/aad/adal/AuthenticationContext.java
RandalliLama: @imakhalova Coding conventions are not the central issue of this PR.  Coding conventions are a large enough discussion that I think we should separate it from this particular PR.  Let's open an issue. We can use that issue to come to a conclusion about coding conventions.

src/src/com/microsoft/aad/adal/AuthenticationResult.java
```diff
@@ -306,4 +323,14 @@ void setTenantId(String tenantid) {
     void setRefreshToken(String refreshToken){
         mRefreshToken = refreshToken;
     }
+    
+    final String getFamilyClientId()
+    {
+        return mFamilyClientId;
+    }
+    
+    final void setFamilyClientId (final String familyClientId)
```
imakhalova: why do you make this methods final?

src/src/com/microsoft/aad/adal/AuthenticationContext.java
imakhalova: Then why do we need to check it? why it can be not found by regular getRefreshToken in cache? Authority and clientId are the same? As I understand cacheKey doesn't contain familyId?

src/src/com/microsoft/aad/adal/AuthenticationContext.java
imakhalova: @RandalliLama ,the problem isn't really code convention here, but that this convention change and refactoring that increases PR...And I've got merge failure once just because of this.

src/src/com/microsoft/aad/adal/AuthenticationContext.java
imakhalova: So actually we can use refreshToken even if we're not within this family?
I'm not sure I follow the idea.

src/src/com/microsoft/aad/adal/AuthenticationContext.java
```diff
@@ -1854,7 +1980,18 @@ public void run() {
...
+                {
+                    authResult = getTokenWithRefreshToken(null, false, request, refreshItem, false);
+                }
+                catch (final AuthenticationException authException)
+                {
+                    callbackHandle.onError(authException);
+                    return;
+                }
+                
+                callbackHandle.onSuccess(authResult);
```
imakhalova: why did you move the callback call out of the method and produced duplicated code?...

src/src/com/microsoft/aad/adal/AuthenticationContext.java
```diff
@@ -1261,63 +1269,100 @@ private AuthenticationResult localFlow(CallbackHandler callbackHandle,
...
+            {
+                authResult = getTokenWithRefreshToken(activity, useDialog, request, refreshItem, true);
+            }
+            catch (AuthenticationException authenticationException)
+            {
+                callbackHandle.onError(authenticationException);
+                return null;
+            }
+            
+            if (authResult != null && !StringExtensions.IsNullOrBlank(authResult.getAccessToken()))
```
imakhalova: Can authResult  be null?

src/src/com/microsoft/aad/adal/AuthenticationContext.java
```diff
@@ -1625,10 +1753,13 @@ private AuthenticationResult refreshToken(final CallbackHandler callbackHandle,
                 // remove item from cache to avoid same usage of
                 // refresh token in next acquireToken call
                 removeItemFromCache(refreshItem);
-                return acquireTokenLocalCall(callbackHandle, activity, useDialog, request);
-            } else {
+                return result;
```
imakhalova: you don't try to aquireToken with UI? you implicitly return token failure as successful case?

src/src/com/microsoft/aad/adal/AuthenticationContext.java
imakhalova: As I understand at this point you can have null authResult and invoke success callback

src/src/com/microsoft/aad/adal/AuthenticationContext.java
```diff
@@ -1261,63 +1269,100 @@ private AuthenticationResult localFlow(CallbackHandler callbackHandle,
...
+            {
+                callbackHandle.onError(authenticationException);
+                return null;
+            }
+            
+            if (authResult != null && !StringExtensions.IsNullOrBlank(authResult.getAccessToken()))
+            {
+                callbackHandle.onSuccess(authResult);
+                return authResult;
+            }
```
imakhalova: what happens in else? Callback never invoked and no exception...we just hang?

src/src/com/microsoft/aad/adal/AuthenticationResult.java
```diff
@@ -117,6 +119,21 @@
         mErrorCodes = errorCodes;
         mStatus = AuthenticationStatus.Failed;
     }
+    
+    AuthenticationResult(final String accessToken, final String refreshToken, final Date expires, final boolean isBroad,
```
imakhalova: Is it used?

src/src/com/microsoft/aad/adal/AuthenticationContext.java
```diff
@@ -1625,10 +1753,13 @@ private AuthenticationResult refreshToken(final CallbackHandler callbackHandle,
                 // remove item from cache to avoid same usage of
                 // refresh token in next acquireToken call
                 removeItemFromCache(refreshItem);
```
imakhalova: Is it possible that one refreshToken for familyId is expired, but I've got other, can it remove all of them?
And I don't see the logic where you put familyId to cache

src/src/com/microsoft/aad/adal/AuthenticationContext.java
weijjia: if refresh token request success, and we get a token, then return the token back. The following piece of code will future check if refresh token doesn't exist or refresh token fails to give you the refresh token, if it's non-silent request, we will try with interactive flow.

src/src/com/microsoft/aad/adal/AuthenticationContext.java
imakhalova: Does it mean that even if we don't have userId we still keep item in cache? how do we ever use them?

src/src/com/microsoft/aad/adal/AuthenticationContext.java
weijjia: family id today is only a flag(it's just 1 today) returned from Evo sts, it's across tenant, across authority.  We also don't return it back to client. When we receive a token response from server with FoCi marked, we'll store it in the cache. For client using us, you acquire token following the way you were following today. If we don't find regular refresh token or mrrt, we will try to find a token cache with FoCI marked, and if found, we will try to use it.

src/src/com/microsoft/aad/adal/AuthenticationContext.java
imakhalova: Got it, thanks

src/src/com/microsoft/aad/adal/AuthenticationContext.java
weijjia: refresh token request doesn't involve with UI. I'm not returning token failure as successful case, I'm just returning the AuthenticationResult(could contain token or error), caller will check if it contains token, if it does, callback on success will be called. If it contains error, caller will check if it's non-silent request, and if it is, will start the interactive flow. The whole purpose of this method is purely just trying to get token with refresh token.

src/src/com/microsoft/aad/adal/AuthenticationContext.java
imakhalova: Oh, that clarifies the logic a little, thanks for explanation. I didn't know how API works, is there any spec around this? 
I thought the server will accept refreshtoken that I got from another app and return new refreshtoken for my clientId. And it knows somehow that we're within one family.
When do we receive token with FoCi? What if we just store them by different CacheKey - using just userId and familyId (currently flag) ? That way you won't need to loop thought all items in Cache

src/src/com/microsoft/aad/adal/AuthenticationContext.java
imakhalova: What if you just return null here, then you don't need to check authResult and accessToken, you'll check only result

src/src/com/microsoft/aad/adal/AuthenticationContext.java
weijjia: TokenCacheItem is created from AuthenticationResult. what do you mean by removing all of them?
Let's say if i found a regular refresh token or mrrt, and i try to get AT with it. But refresh token request fail(expired), then removeCacheItem will try to remove the following keys: 1) the one used to find RT or MRRT 2) key with user id 3) key with displayable

For family client id refresh token, if the one we use to get AT expired, we won't remove it. You can only update your own entry (with matched client id, authority, resource).

imakhalova: Ok, then the question is why we don't remove refreshToken for FamilyId? Should we remove it if it's expired?
What is MRRT?

src/src/com/microsoft/aad/adal/AuthenticationContext.java
weijjia: Yup, there is spec for this I think. 
Server side has the logic to check if FoCI flag should be returned along with the token. If we store familyid as part of the key, since client app itself doesn't know which family they belong to, and if it's in broker case, how do we look up the cache? Since this feature only applies to first party apps today, it makes things easier, and we want to have a simple design for now.

src/src/com/microsoft/aad/adal/AuthenticationContext.java
weijjia: authresult could have the error code and error description, if i simply return null, then i lose those info.

src/src/com/microsoft/aad/adal/AuthenticationContext.java
weijjia: MRRT is multi resource refresh token. 
My understanding is, you shouldn't remove other app's entry. It also depends on the detailed error code server return to us, if they explicitly tell us refresh token request fails because token is expired, then we probably should consider for this. Ryan and I talked, there should be further improvement on this, we can probably have a timestamp  on the token cache item, when we try to find family item, we usually find the latest one.

src/src/com/microsoft/aad/adal/AuthenticationContext.java
imakhalova: Then you don't need familyId as part of the key, just use username. It'll simplify the logic a lot.
I don't get why do you need to store this flag in Cache, because if you store it by different CachKey it will be obvious that you put it there because of this flag.
And then within getRefreshToken you'll need something like 
            if (item == null || StringExtensions.IsNullOrBlank(item.getRefreshToken())) {
                // if not present, check universal item for family
                Logger.v(TAG, "Looking for Refresh token with Family Client Id");
                keyUsed = userId;
                item = mTokenCacheStore.getItem(keyUsed);
                familyResource = true;
            }
                refreshItem = new RefreshItem(keyUsed, request, item, multiResource, familyResource);
It will simplify logic a lot

src/src/com/microsoft/aad/adal/AuthenticationContext.java
imakhalova: We get to the point of removal when we know that it's expired. Otherwise you Cache keeps growing and filled with expired tokens...

src/src/com/microsoft/aad/adal/AuthenticationContext.java
imakhalova: Oh, right, and we don't throw this error because we will try second attempt to get token...

src/src/com/microsoft/aad/adal/AuthenticationContext.java
brentschmaltz: +1 on not changing style within a file.

src/src/com/microsoft/aad/adal/AuthenticationContext.java
weijjia: I totally agree for regular refresh token cache entry, and that is what we're doing. For family item, the other reason we don't update cache entry for other client id is, refresh token can record more info, i.e mfa.

src/src/com/microsoft/aad/adal/AuthenticationContext.java
weijjia: it could possible be. And i think it's always to safe to check.

src/src/com/microsoft/aad/adal/AuthenticationResult.java
weijjia: i don't want them to be overridden.

imakhalova: And why is it so? It might be useful for UnitTesting to be able to override these methods

src/src/com/microsoft/aad/adal/AuthenticationContext.java
weijjia: Had a discussion on this, and i'll change it back to the current style.

src/src/com/microsoft/aad/adal/AuthenticationResult.java
weijjia: why you want to override a set or get method?

src/src/com/microsoft/aad/adal/TokenCacheItem.java
```diff
@@ -57,6 +57,8 @@
     private boolean mIsMultiResourceRefreshToken;
 
     private String mTenantId;
+    
+    private String mFamilyClientId;
```
imakhalova: Have you checked the upgrade scenario?
If user currently has token without mFamilyClientId will it work fine?

imakhalova: The reason I'm asking that this class implements Serializable, but I'm not sure if it actually ever used to serialize this object.

src/src/com/microsoft/aad/adal/AuthenticationContext.java
```diff
@@ -1461,6 +1525,65 @@ private RefreshItem getRefreshToken(final AuthenticationRequest request) {
...
+                final UserInfo tokenUserInfo = tokenCacheItem.getUserInfo();
+                if (tokenUserInfo!= null 
+                    && tokenUserInfo.getUserId() != null 
+                    && tokenUserInfo.getUserId().equalsIgnoreCase(userId))
+                {
+                    if (!StringExtensions.IsNullOrBlank(tokenCacheItem.getFamilyClientId())
+                            && !StringExtensions.IsNullOrBlank(tokenCacheItem.getRefreshToken()))
+                    {
+                        Logger.v(TAG + methodName, "Found family item matching the given user id, and the clientId selected for FoCI is: " + tokenCacheItem.getClientId());
+                        return tokenCacheItem;
```
imakhalova: How do you verify that it's fresh one? What if one of family token expired and you need to use another one which is valid yet...

src/src/com/microsoft/aad/adal/AuthenticationContext.java
```diff
@@ -1448,6 +1497,21 @@ private RefreshItem getRefreshToken(final AuthenticationRequest request) {
...
+            {
+                if (!UrlExtensions.isADFSAuthority(StringExtensions.getUrl(mAuthority)))
+                {
+                    Logger.v(TAG + methodName, "No refresh token found, trying to find family client id item.");
+                    item = findFamilyItemForUser(request.getUserIdentifierType(), userId);
+                }
+                else 
+                {
+                    Logger.v(TAG + methodName, "No refresh token found, skip the family client id item lookup for ADFS authority.");
+                }
```
imakhalova: you don't actually set keyUsed and your RefreshItem will be created with MultiResourceRefreshTokenKey

src/src/com/microsoft/aad/adal/AuthenticationContext.java
```diff
@@ -1261,63 +1269,101 @@ private AuthenticationResult localFlow(CallbackHandler callbackHandle,
...
+            }
+        } 
+        
+        // refresh token does not exist or refresh token request failed to give back the token. 
+        // If it's non-silent request, will try to acquire token interactively. 
+        // if it's silent request, will return the AUTH_REFRESH_FAILED_PROMPT_NOT_ALLOWED.
+        if (refreshItem == null || authResult == null 
+                || (authResult != null && StringExtensions.IsNullOrBlank(authResult.getAccessToken()))) 
+        {
+            Logger.v(TAG, "Refresh token is not available or refresh token request failed to return token.");
```
imakhalova: What if request with refreshtoken for my client app failed, shouldn't I try to hit server with refreshToken for my family?

imakhalova: Scenario: signed in to Word->didn't use app for a while and token got expired-> go to Excel and sign in -> go back to Word -> it will check token and will ask for credentials...but it shouldn't

src/src/com/microsoft/aad/adal/AuthenticationContext.java
```diff
@@ -1261,63 +1268,86 @@ private AuthenticationResult localFlow(CallbackHandler callbackHandle,
...
         if (!promptUser(request.getPrompt()) && refreshItem != null
                 && !StringExtensions.IsNullOrBlank(refreshItem.mRefreshToken)) {
             Logger.v(TAG, "Refresh token is available and it will attempt to refresh token");
-            return refreshToken(callbackHandle, activity, useDialog, request, refreshItem, true);
-        } else {
-            Logger.v(TAG, "Refresh token is not available");
+            try {
+                authResult = getTokenWithRefreshToken(activity, useDialog, request, refreshItem, true);
+            } catch (AuthenticationException authenticationException) {
+                callbackHandle.onError(authenticationException);
```
agamemnum: Above code checks `callbackHandle.callback != null` shouldn't you do it in here? Actually as this function is private I recommend to remove dual interface support on it (when callback can be null and you need to throw directly)

src/src/com/microsoft/aad/adal/AuthenticationContext.java
```diff
@@ -1261,63 +1268,86 @@ private AuthenticationResult localFlow(CallbackHandler callbackHandle,
...
+                return null;
+            }
+            
+            if (authResult != null && !StringExtensions.IsNullOrBlank(authResult.getAccessToken())) {
+                callbackHandle.onSuccess(authResult);
+                return authResult;
```
agamemnum: Avoid having deep early returns - code becomes hardly readable

src/src/com/microsoft/aad/adal/AuthenticationContext.java
```diff
@@ -1261,63 +1268,86 @@ private AuthenticationResult localFlow(CallbackHandler callbackHandle,
...
-                    // onActivityResult will receive the response
-                    if (!startAuthenticationActivity(activity, request)) {
-                        callbackHandle.onError(new AuthenticationException(
-                                ADALError.DEVELOPER_ACTIVITY_IS_NOT_RESOLVED));
-                    }
-                }
+                acquireTokenInteractively(callbackHandle, activity, request, useDialog);
             } else {
-
+                final String errorInfo = authResult == null? "" : authResult.getErrorLogInfo();
```
agamemnum: nit: formatting

src/src/com/microsoft/aad/adal/AuthenticationContext.java
```diff
@@ -1261,63 +1268,86 @@ private AuthenticationResult localFlow(CallbackHandler callbackHandle,
...
+                    ADALError.DEVICE_CONNECTION_IS_NOT_AVAILABLE);
+            callbackHandle.onError(exc);
+            return;
+        }
+        
+        // start activity if other options are not available
+        // delegate map is used to remember callback if another
+        // instance of authenticationContext is created for config
+        // change or similar at client app.
+        mAuthorizationCallback = callbackHandle.callback;
```
agamemnum: If we are putting callback in the map, why do we need callback as a member? And what happens if two requests have came?

src/src/com/microsoft/aad/adal/AuthenticationContext.java
```diff
@@ -1461,6 +1503,53 @@ private RefreshItem getRefreshToken(final AuthenticationRequest request) {
...
 
+    private TokenCacheItem findFamilyItemForUser(final UserIdentifierType userIdentifierType, final String userId) {
+        final String methodName = ":findFamilyItemForUser";
+            
+        if (StringExtensions.IsNullOrBlank(userId)) {
+            Logger.v(TAG + methodName, "User id is not provided, cannot continue with finding family item.");
+            return null;
+        }
+
+        Iterator<TokenCacheItem> allItems = mTokenCacheStore.getAll();
```
agamemnum: I think it would be much better if TokenCacheStore supports `Iterable<T>` - so you don't need to rebuild all iteration logic yourself.

src/src/com/microsoft/aad/adal/AuthenticationContext.java
```diff
@@ -1461,6 +1503,53 @@ private RefreshItem getRefreshToken(final AuthenticationRequest request) {
...
+        }
+        
+        if (userIdentifierType == UserIdentifierType.UniqueId) {
+            Logger.v(TAG + methodName, "UserIdentifier type is unique id, will look for the family item based on unique id.");
+            while (allItems.hasNext()) {
+                final TokenCacheItem tokenCacheItem = allItems.next();
+                final UserInfo tokenUserInfo = tokenCacheItem.getUserInfo();
+                if (tokenUserInfo!= null 
+                    && tokenUserInfo.getUserId() != null 
+                    && tokenUserInfo.getUserId().equalsIgnoreCase(userId)) {
```
agamemnum: nit:invert compare to simplify code, i.e. `userId.equalsIgnoreCase(tokenUserInfo.getUserId())`. This way you don't need to check for null

src/src/com/microsoft/aad/adal/AuthenticationContext.java
weijjia: Well, callbackHandle itself is already checking if callback is null, and it will throw out the exception if it is null.

src/src/com/microsoft/aad/adal/AuthenticationContext.java
weijjia: We'll need more refactor in the code to make this goal, that's beyond current scope.

src/src/com/microsoft/aad/adal/AuthenticationContext.java
weijjia: I don't think there is any need to store the callback as a member. But I don't want to change the behavior for the current code unless we get time looking into it.

src/src/com/microsoft/aad/adal/TokenCacheItem.java
weijjia: yup, we also have unit test for this. Also adding new fields into serializable class is a compatible change.

src/src/com/microsoft/aad/adal/AuthenticationContext.java
weijjia: @imakhalova I do miss the comment here. It's already merged, and i'm opening up an issue https://github.com/AzureAD/azure-activedirectory-library-for-android/issues/525 for us to tackle the issues.

src/src/com/microsoft/aad/adal/AuthenticationContext.java
weijjia: I think one method should only handle one thing, the whole purpose of getTokenithRefreshToken should just be returning back the result, and its caller should be responsible for using the callback to deliver the result back to calling app.

PR #506, opened 2016-02-19
Log exception stack trace in external logger as additional message

imakhalova:
Let client app see the cause of error


weijjia:
:shipit:

Issue #493, opened 2016-02-09
No way to turn-off Hardware accelration in AuthenticationActivity webView


talk2arpit:
We have seen couple of crashes on some Android M devices due to Web View crash because of bad GPU drivers (ref: [https://code.google.com/p/chromium/issues/detail?can=2&q=541145&colspec=ID%20Pri%20M%20Stars%20ReleaseBlock%20Cr%20Status%20Owner%20Summary%20OS%20Modified&id=541145] or [https://code.google.com/p/chromium/issues/detail?id=554944], which supposedly can be fixed for time being by explicitly disabling Hardware acceleration in WebView i.e. 'webView.setLayerType(View.LAYER_TYPE_SOFTWARE, null)' which is by default always enable so there be should some way caller can set WebView Layer type in AuthenticationActivity webView e.g. 
AuthenticationSettings.INSTANCE.setLayerType(...) or AuthenticationSettings.INSTANCE.setWebViewHardwareAccelration(Boolean enableOrDisable)

PR #509, opened 2016-02-23
Fix of #493: No way to turn-off Hardware acceleration in AuthenticationActivity WebView

talk2arpit:
Issue: By default WebView is always hardwareAccelerated due to which on
some devices it crashes intermittently because of bad GPU drivers. There
is a way to turn-off hardware acceleration by setting webview layer type
to 'LAYER_TYPE_SOFTWARE' but as webView used in AuthenticationActivity
and AuthenticationDialog is not directly accessible to caller so there
is no way caller could disable it.

Changes: Added a method in AuthenticationSettings i.e.
setHardwareAcceleration, which caller can use to disable WebView
hardware acceleration.

### DIFF
--- a/src/src/com/microsoft/aad/adal/AuthenticationSettings.java
+++ b/src/src/com/microsoft/aad/adal/AuthenticationSettings.java
@@ -187,8 +187,9 @@ public enum AuthenticationSettings {
 
     /**
      * Set flag to use or skip broker.
+     * By default, the flag value is false and ADAL will not talk to broker.
      * 
-     * @param useBroker False to not use broker
+     * @param useBroker True to use broker
      */
     public void setUseBroker(boolean useBroker) {
         this.mUseBroker = useBroker;
--- a/src/src/com/microsoft/aad/adal/AuthenticationSettings.java
+++ b/src/src/com/microsoft/aad/adal/AuthenticationSettings.java
@@ -38,6 +38,8 @@ public enum AuthenticationSettings {
     private Class<?> mClazzDeviceCertProxy;
 
     private String mActivityPackageName;
+    
+    private boolean mEnableHardwareAcceleration = true;
 
     /**
      * SharedPreference package name to load this file from different context.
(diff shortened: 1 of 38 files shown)


src/src/com/microsoft/aad/adal/AuthenticationSettings.java
```diff
@@ -253,4 +255,31 @@ public void setReadTimeOut(int timeOutMillis) {
...
+     * {@link AuthenticationActivity} and {@link AuthenticationDialog}. 
+     * By default hardware acceleration is enable in WebView.
+     *  
+     * @param enable
+     *             if true, WebView would be hardwareAccelerated else it 
+     *             would be disable.
+     *             
+     * @see #isWebViewHardwareAccelerated()
+     */
+    public void setWebViewHardwareAcceleration(boolean enable) {
```
weijjia: What you really want to achieve is to disable the webiew acceleration. I would like to call this disableWebviewHardwareAcceleration.

talk2arpit: By default, hardware acceleration is always enable in View, which can be disabled. [ref: http://developer.android.com/guide/topics/graphics/hardware-accel.html ]. Due to hardware accelerated AuthenticationActivity WebView, we have seen couple of crashes on some devices so we want to disable it in our app.
Here, this new method can be used to disable webview hardware acceleration.  Although by renaming this method to disableWebViewHardwareAccelertaion, would also solve the purpose but for completeness purpose i.e. enable/disable and query hardware acceleration in a same session, I use this name, something similar to setSkipBorker.

weijjia: Can we have the the name as setEnableHardwareAcceleration(boolean enabled)/getEnableHardwareAcceleration?

RandalliLama: This sounds too much like it is changing the default.  The current default is to use hardware acceleration and that should remain the default.  The caller should have a flag they can set to true to disable hardware acceleration.  That would imply the name should be something like:

setDisableWebViewHardwareAcceleration
getDisableWebViewHardwareAcceleration

talk2arpit: I have renamed the methods as suggested. Kindly review the latest changes.

addev-ashish:
#493

PR #510, opened 2016-02-24
Reveal uncaught exceptions

imakhalova:
All async tasks currently swallow runtime exceptions because if you use ExecutorSerice.submit() the only way to get exceptions is to call blocking get() method of the Future that was returned by submit.
When switch to execute it will throw exceptions to thread parent.
To support backward compat for existing public methods creating SettableFuture that will provide Future behavior (blocking get() until get results)

Also fixed some unit tests that apprerently were failing half way with NullPointerExceptions and no one was aware of it.

### DIFF
--- a/src/src/com/microsoft/aad/adal/AuthenticationContext.java
+++ b/src/src/com/microsoft/aad/adal/AuthenticationContext.java
@@ -30,6 +30,7 @@ import java.util.concurrent.ExecutionException;
 import java.util.concurrent.ExecutorService;
 import java.util.concurrent.Executors;
 import java.util.concurrent.Future;
+import java.util.concurrent.FutureTask;
 import java.util.concurrent.locks.Lock;
 import java.util.concurrent.locks.ReentrantReadWriteLock;
 
@@ -576,8 +577,7 @@ public class AuthenticationContext {
      */
     public AuthenticationResult acquireTokenSilentSync(String resource, String clientId,
             String userId) {
-        Future<AuthenticationResult> futureResult = acquireTokenSilent(resource, clientId, userId,
-                null);
+        Future<AuthenticationResult> futureResult = acquireTokenSilent(resource, clientId, userId, null);
         try {
             return futureResult.get();
         } catch (InterruptedException e) {
@@ -625,7 +625,7 @@ public class AuthenticationContext {
      *         {@link UserInfo}.
      */
     public Future<AuthenticationResult> acquireTokenSilent(String resource, String clientId,
-            String userId, AuthenticationCallback<AuthenticationResult> callback) {
+            String userId, final AuthenticationCallback<AuthenticationResult> callback) {
         if (StringExtensions.IsNullOrBlank(resource)) {
             throw new IllegalArgumentException("resource");
         }
@@ -638,7 +638,25 @@ public class AuthenticationContext {
         request.setSilent(true);
         request.setPrompt(PromptBehavior.Auto);
         request.setUserIdentifierType(UserIdentifierType.UniqueId);
-        return acquireTokenLocal(null, false, request, callback);
+        final SettableFuture<AuthenticationResult> futureTask = new SettableFuture();
+        acquireTokenLocal(null, false, request, new AuthenticationCallback<AuthenticationResult>() {
+            @Override
+            public void onSuccess(AuthenticationResult result) {
+                if (callback != null) {
+                    callback.onSuccess(result);
+                }
+                futureTask.set(result);
+            }
+
+            @Override
+            public void onError(Exception exc) {
+                if (callback != null) {
+                    callback.onError(exc);
+                }
+                futureTask.setException(exc);
+            }
+        });
+        return futureTask;
     }
 
     /**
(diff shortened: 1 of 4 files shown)


src/src/com/microsoft/aad/adal/AuthenticationContext.java
```diff
@@ -1907,7 +1923,7 @@ private void checkInternetPermission() {
         }
 
         public boolean isConnectionAvailable() {
-            ConnectivityManager connectivityManager = (ConnectivityManager)mConnectionContext
+            ConnectivityManager connectivityManager = (ConnectivityManager) mConnectionContext
```
RandalliLama: Please don't change whitespace unless it has a functional impact somehow.

src/src/com/microsoft/aad/adal/AuthenticationContext.java
```diff
@@ -1926,4 +1942,27 @@ public static String getVersionName() {
         // value
         return "1.1.12";
     }
+
+    /**
+     * A {@link Future}  whose result can be set by a {@link #set(Object)} or {@link #setException(Throwable)}
+     */
+    final static class SettableFuture<V> extends FutureTask<V> {
```
RandalliLama: Sorry if this would be obvious to a regular Java programmer, but why are you doing this?

RandalliLama: After re-reading you original PR comment I think it get it a bit better.  However, I'm not sure I understand why this change is necessary.  Why is it a problem that the caller needs to call the blocking get() function?  Can't they do that after they are certain the operation has completed and therefore ensure that no blocking will actually occur?  I'm not saying that it is the best interface, just that if it is possible to make it work then I don't see any reason to change it at this time.

imakhalova: The problem is that we currently don't notify user that some RuntimeException happened (callback won't be invoked if this happens). So caller doesn't know when to call get() function.
Moreover I see that all our async calls don't check Future at all, so RuntimeExceptions got swallowed. 
I'm actually open to any proposals how to fix it...

src/src/com/microsoft/aad/adal/AuthenticationContext.java
imakhalova: done

src/src/com/microsoft/aad/adal/AuthenticationContext.java
RandalliLama: Got it.  I'll give it a bit more thought, but maybe you have the best solution.

RandalliLama:
Here is a crazy idea that I think does a good job of centralizing the behavior change that we want and doesn't disturb very much code to do it.

Define a function like this:

```java
	// Runs a task on an ExecutorService without catching or transferring RuntimeExceptions
	// back to the current thread, while preserving all of the other semantics of
	// ExecutorService.submit().
	Future<V> safeExecutor(ExecutorService execService, Callable<V> callable)
	{
		Callable<V> task = new Callable<V>() {
			@Override
			public AuthenticationResult call() {
				V result;
				
				Semaphore semaphore = new Semaphore(1);
				semaphore.acquire();

				Runnable executed = new Runnable() {
					result = callable.call();
					semaphore.release();
				}
				execService.execute(runnableTask);
				semaphore.acquire();

				return result;
			}
		}
	 	return execService.submit(task);
	}
```

And then replace this:
```java
        return sThreadExecutor.submit(new Callable<AuthenticationResult>() {

            @Override
            public AuthenticationResult call() {
                Logger.v(TAG, "Running task in thread:" + android.os.Process.myTid());
                return acquireTokenLocalCall(callbackHandle, activity, useDialog, request);
            }
        });
```

With this:
```java
		return safeExecutor(sExectuorService, new Callable<AuthenticationResult>() {
			@Override
			public AuthenticationResult call() {
	        	Logger.v(TAG, "Running task in thread:" + android.os.Process.myTid());
				return acquireTokenLocalCall(callbackHandle, activity, useDialog, request);
			}
		});
```

imakhalova:
I need to try this out, my guess it that the outer submit should swallow the Runtime exception of inner execute. But I'll give it a try

imakhalova:
This actually causes deadlock, because during execution of 1 task we submit another one and wait until it's done, but thread pool allows only 1 thread (I guess initially that was meant to avoid concurrency issues)

RandalliLama:
Ok, I was hoping that it would be smart enough to re-use the thread.  But that was probably asking too much.  I give up.  

:shipit:

src/src/com/microsoft/aad/adal/AuthenticationContext.java
agamemnum: @RandalliLama There is another option - we inherit from the ThreadPoolExecutor and override `afterExecute` inside we check if "task" is a `Future` and check it's result. If result is `RuntimeException` we crash.

RandalliLama: If that works it sounds like a great idea.  It would be easy to read and only affects the issue at hand.

imakhalova: I have to warn about the problem that this solution can cause.
The main concern is method   
public Future<AuthenticationResult> acquireTokenSilent(String resource, String clientId,
            String userId, AuthenticationCallback<AuthenticationResult> callback).
First I don't really like it's public, why caller would ever need Future and deal with it.... I think it meant to be private. But anyway it's already public and as I understand we want to support it.

So the problem if we throw within afterExecutre AuthenticationException,
even if user calls Future.get() and than catches AuthenticationException we still crash...
I guess it might be ok, since if we throw AuthenticationException instead of returning them in callback we treat them as IlligalStateException....But I want you to know about this consequence....

imakhalova: This approach will cause break to Sync method, I'll keep the strategy with SettableFuture, but I create a method that can avoid all this hacky staff. And marked old method as deprecated, when clients will switch we can remove old method with SettableFuture

RandalliLama: Sounds reasonable.

weijjia:
@imakhalova Can you resolve the merging conflicts in this PR? I

imakhalova:
@weijjia, done. All conflicts were automatically resolved by git

PR #514, opened 2016-02-29
remove static sKey and sMacKey, also fix issue 512

weijjia:
1. fix issue 512
2. remove the static class variable for solving the decryption problem when two broker apps are involved: 
1) installing azure authenticator, and doing wpj from azure authenticator, perform CA flow
2) install company portal
3) uninstall AA app. 
Company portal requires user to do sign in, and when sign-in happened, the static variable will be set. When CA flow happened, data will always be encrypted with company portal's key, and will be decrypted with the key stored in account manager userdata.

### DIFF
--- a/src/src/com/microsoft/aad/adal/DefaultTokenCacheStore.java
+++ b/src/src/com/microsoft/aad/adal/DefaultTokenCacheStore.java
@@ -115,7 +115,7 @@ public class DefaultTokenCacheStore implements ITokenCacheStore, ITokenStoreQuer
 
     private String decrypt(final String key, final String value) {
         if (StringExtensions.IsNullOrBlank(key)) {
-            throw new IllegalArgumentException("key is null");
+            throw new IllegalArgumentException("key is null or blank");
         }
         
         try {
--- a/src/src/com/microsoft/aad/adal/DefaultTokenCacheStore.java
+++ b/src/src/com/microsoft/aad/adal/DefaultTokenCacheStore.java
@@ -113,7 +113,7 @@ public class DefaultTokenCacheStore implements ITokenCacheStore, ITokenStoreQuer
         return null;
     }
 
-    private String decrypt(final String key, String value) {
+    private String decrypt(final String key, final String value) {
         try {
             return sHelper.decrypt(value);
         } catch (Exception e) {
--- a/src/src/com/microsoft/aad/adal/DefaultTokenCacheStore.java
+++ b/src/src/com/microsoft/aad/adal/DefaultTokenCacheStore.java
@@ -118,7 +118,7 @@ public class DefaultTokenCacheStore implements ITokenCacheStore, ITokenStoreQuer
             return sHelper.decrypt(value);
         } catch (Exception e) {
             Logger.e(TAG, "Decryption failure", "", ADALError.ENCRYPTION_FAILED, e);
-            if (!StringExtensions.IsNullOrBlank(value)) {
+            if (!StringExtensions.IsNullOrBlank(key)) {
                 removeItem(key);
                 Logger.v(TAG, String.format("Decryption error, item removed for key: '%s'", key));
             }
--- a/src/src/com/microsoft/aad/adal/DefaultTokenCacheStore.java
+++ b/src/src/com/microsoft/aad/adal/DefaultTokenCacheStore.java
@@ -114,14 +114,16 @@ public class DefaultTokenCacheStore implements ITokenCacheStore, ITokenStoreQuer
     }
 
     private String decrypt(final String key, final String value) {
+        if (StringExtensions.IsNullOrBlank(key)) {
+            throw new IllegalArgumentException("key is null");
+        }
+        
         try {
             return sHelper.decrypt(value);
         } catch (Exception e) {
             Logger.e(TAG, "Decryption failure", "", ADALError.ENCRYPTION_FAILED, e);
-            if (!StringExtensions.IsNullOrBlank(key)) {
-                removeItem(key);
-                Logger.v(TAG, String.format("Decryption error, item removed for key: '%s'", key));
-            }
+            removeItem(key);
+            Logger.v(TAG, String.format("Decryption error, item removed for key: '%s'", key));
         }
 
         return null;
--- a/src/src/com/microsoft/aad/adal/DefaultTokenCacheStore.java
+++ b/src/src/com/microsoft/aad/adal/DefaultTokenCacheStore.java
@@ -25,6 +25,7 @@ import java.util.Date;
 import java.util.HashSet;
 import java.util.Iterator;
 import java.util.Map;
+import java.util.Map.Entry;
 
 import javax.crypto.NoSuchPaddingException;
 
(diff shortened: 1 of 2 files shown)


weijjia:
@sdabbiru Can you also take a look at this?

src/src/com/microsoft/aad/adal/DefaultTokenCacheStore.java
```diff
@@ -112,16 +113,16 @@ private String encrypt(String value) {
...
         } catch (Exception e) {
             Logger.e(TAG, "Decryption failure", "", ADALError.ENCRYPTION_FAILED, e);
             if (!StringExtensions.IsNullOrBlank(value)) {
                 Logger.v(TAG, String.format("Decryption error for key: '%s'. Item will be removed",
-                        value));
-                removeItem(value);
-                Logger.v(TAG, String.format("Item removed for key: '%s'", value));
+                        key));
+                removeItem(key);
+                Logger.v(TAG, String.format("Item removed for key: '%s'", key));
```
brentschmaltz: Why do we have two log statements here? Could we have just one?

src/src/com/microsoft/aad/adal/StorageHelper.java
```diff
@@ -148,6 +148,9 @@ final private void loadSecretKeyForAPI() throws NoSuchAlgorithmException,
             InvalidKeySpecException {
         // Loading key only once for performance. If API is upgraded, it will
         // restart the device anyway. It will load the correct key for new API.
+        // Remove the static class variable sKey and sMacKey. Static variables are
```
brentschmaltz: It is not clear what this comment is for.

weijjia: I just want to make a note why we remove the static here. I'll update.

src/src/com/microsoft/aad/adal/DefaultTokenCacheStore.java
```diff
@@ -112,16 +113,14 @@ private String encrypt(String value) {
         return null;
     }
 
-    private String decrypt(String value) {
+    private String decrypt(final String key, String value) {
         try {
             return sHelper.decrypt(value);
         } catch (Exception e) {
             Logger.e(TAG, "Decryption failure", "", ADALError.ENCRYPTION_FAILED, e);
             if (!StringExtensions.IsNullOrBlank(value)) {
```
imakhalova: do you need to check it for null? remove anyway, there shouldn't be empty records in cache...

src/src/com/microsoft/aad/adal/StorageHelper.java
```diff
@@ -127,7 +127,7 @@
 
     private static String sBlobVersion;
 
-    private static SecretKey sKey = null, sMacKey = null;
+    private SecretKey sKey = null, sMacKey = null;
```
imakhalova: nit: if it's not static anymore, use prefix m

src/src/com/microsoft/aad/adal/StorageHelper.java
```diff
@@ -127,7 +127,7 @@
 
     private static String sBlobVersion;
```
imakhalova: should this also be non static? what if it fails to get key and macKey for one StorageHelper instance, but won't fail for another?

imakhalova:
interesting bug, good catch

src/src/com/microsoft/aad/adal/DefaultTokenCacheStore.java
weijjia: thanks for catching..i should check for key

src/src/com/microsoft/aad/adal/StorageHelper.java
weijjia: I'm not worrying about this the version too much at this moment. The problem with the current static key and mackey is we have a early return when starting doing the encryption, and it's causing failure for broker apps. i don't get your concern for failing to get key for one instance but not the other instance.

src/src/com/microsoft/aad/adal/DefaultTokenCacheStore.java
imakhalova: don't check key for null, remove the record for any key

src/src/com/microsoft/aad/adal/StorageHelper.java
imakhalova: The same problem that you currently fixing:
One 1 set Key and MacKey and version to VERSION_ANDROID_KEY_STORE
Another app sets another Key and MacKey but for some reason sets version to VERSION_USER_DEFINED
Then if you're using static version you won't be able to decrypt anything with first Key, because it thinks it has wrong version.

weijjia: k, so the key and macKey is no longer static, they belong to the storagehelper instance. If one instance fail to get the key and macKey, then I don't even have anything to encrypt, right? Generally, I don't want us to have the static class variable in the class, it's dangerous here. 

I'll stop arguing here, i do agree this should also be non-static. :)

src/src/com/microsoft/aad/adal/StorageHelper.java
```diff
@@ -147,8 +147,8 @@ public StorageHelper(Context ctx) throws NoSuchAlgorithmException, NoSuchPadding
     final private void loadSecretKeyForAPI() throws NoSuchAlgorithmException,
             InvalidKeySpecException {
         // Loading key only once for performance. If API is upgraded, it will
-        // restart the device anyway. It will load the correct key for new API.
-        if (sKey != null && sMacKey != null)
+        // restart the device anyway. It will load the correct key for new API. 
+        if (mKey != null && mMacKey != null)
```
sdabbiru: In case of company portal, because authenticator process executes within CP process, the removal of static is not sufficient. You need to remove the entire if statement.

weijjia: but it's different instance.

sdabbiru: After we talked, I agree and am okay with this change

sdabbiru:
:shipit:

src/src/com/microsoft/aad/adal/DefaultTokenCacheStore.java
```diff
@@ -112,16 +113,14 @@ private String encrypt(String value) {
         return null;
     }
 
-    private String decrypt(String value) {
+    private String decrypt(final String key, String value) {
```
kpanwar: make value as final too. Consider updating eclipse's rules to handle all the formatting for you.

src/src/com/microsoft/aad/adal/DefaultTokenCacheStore.java
```diff
@@ -112,16 +113,14 @@ private String encrypt(String value) {
...
         try {
             return sHelper.decrypt(value);
         } catch (Exception e) {
             Logger.e(TAG, "Decryption failure", "", ADALError.ENCRYPTION_FAILED, e);
-            if (!StringExtensions.IsNullOrBlank(value)) {
-                Logger.v(TAG, String.format("Decryption error for key: '%s'. Item will be removed",
-                        value));
-                removeItem(value);
-                Logger.v(TAG, String.format("Item removed for key: '%s'", value));
+            if (!StringExtensions.IsNullOrBlank(key)) {
```
kpanwar: this check for null should happen before you decrypt

src/src/com/microsoft/aad/adal/DefaultTokenCacheStore.java
```diff
@@ -213,17 +212,22 @@ public void removeAll() {
 
         @SuppressWarnings("unchecked")
```
kpanwar: why are we suppressing warnings?

weijjia: sharePreference.getAll is returning Map<String, ?>

kpanwar:
:shipit:

PR #516, opened 2016-03-02
Fix test issues for eclipse test build

imakhalova:
ResourceAuthenticationChallengeException needs to be public since tests are in different package
assertThrowsException should allow Runnable and ThrowableRunnable as argument


weijjia:
We still missing some imports in tests/Functional/src/com/microsoft/aad/adal/test/AuthenticationParamsTests.java I believe.

PR #517, opened 2016-03-02
CallbackHandle checks if callback isn't null

imakhalova:
There is no need to double check if callback is null
And if callback is null, that means nothing should happen, user doesn't want to know about results of operation
Internally we always set callback to non-null value

### DIFF
--- a/src/src/com/microsoft/aad/adal/AuthenticationContext.java
+++ b/src/src/com/microsoft/aad/adal/AuthenticationContext.java
@@ -1023,7 +1023,7 @@ public class AuthenticationContext {
 
     private Handler mHandler;
 
-    class CallbackHandler {
+    private static class CallbackHandler {
         private Handler mRefHandler;
 
         private AuthenticationCallback<AuthenticationResult> callback;
@@ -1231,12 +1231,8 @@ public class AuthenticationContext {
                 } catch (AuthenticationException ex) {
                     // pass back to caller for known exceptions such as failure
                     // to encrypt
-                    if (callbackHandle.callback != null) {
-                        callbackHandle.onError(ex);
-                        return null;
-                    } else {
-                        throw ex;
-                    }
+                    callbackHandle.onError(ex);
+                    return null;
                 }
             } else {
                 Logger.v(TAG, "User is not specified for background token request");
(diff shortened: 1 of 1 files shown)


weijjia:
why do you want to make the change?

src/src/com/microsoft/aad/adal/AuthenticationContext.java
```diff
@@ -1023,7 +1023,7 @@ public boolean cancelAuthenticationActivity(int requestId) {
 
     private Handler mHandler;
 
-    class CallbackHandler {
+    private static class CallbackHandler {
```
imakhalova: Acutally, I think withn CallbackHandler we also shouldn't check if callback null anymore, because we don't allow null callbacks

imakhalova:
1) this is pre requisite for another PR (AuthenticationException derive from Exceptions)
2) we don't allow null callbacks anymore, if it's null we should fail with null pointer exception to notify developer (but I think we check it for all async tasks and throw IlligalArgumentException)
3) removing extra ifs that are not necessary (simplifies code reading)

weijjia:
We absolutely should check for if callback is null and throw illegalArgumentException in all the async tasks. We shouldn't just let it fail and throw an nullPointerException.

imakhalova:
That logic I haven't change, It's still in place. We throw IlligalArgumentException if callback is null.
But I just wonder why do you think it makes difference for developer - to get Nullpointer exception that callback is null or IlligalArgumentException?))

imakhalova:
@weijjia do you have any other concerns?

weijjia:
I might misunderstand here. But where is the change we don't allow null callback anymore? 
For returning nullpointer exception and illegalArugument, I think as a library, if we don't allow null param, we should throw an illegalArgument exception and let the developer know you couldn't pass us a null param instead of throwing the nullpointer exception during runtime.

imakhalova:
Null callback was allowed only for silent requests...and since it's not null for silent request (see Reveal uncaught exceptions PR), there is no methods that can have null callback to acquire token.
Or if I miss something can you point me to any method that allows null callback and doesn't have check/throw IlligalArgumentException?

weijjia:
Got it. This makes sense then. I thought with your last change, there shouldn't be null backs anymore. I got confuse with the description "if it's null we should fail with null pointer exception to notify developer"

PR #518, opened 2016-03-02
AuthenticationException derive from Exceptions

imakhalova:
It allows to detect where this exception need to be caught
Also derive internal errors (AuthenticationServerProtocolException and ResourceAuthenticationChallengeException) from AuthenticationException to keep exception system consistent
Some developer exceptions better to stay as Runtime, using IlligalStateException/IlligalArgumentException for that cases

### DIFF
--- a/src/src/com/microsoft/aad/adal/AuthenticationContext.java
+++ b/src/src/com/microsoft/aad/adal/AuthenticationContext.java
@@ -537,7 +537,7 @@ public class AuthenticationContext {
     private String checkInputParameters(String resource, String clientId, String redirectUri,
             PromptBehavior behavior, AuthenticationCallback<AuthenticationResult> callback) {
         if (mContext == null) {
-            throw new AuthenticationException(ADALError.DEVELOPER_CONTEXT_IS_NOT_PROVIDED);
+            throw new IllegalArgumentException("context", new AuthenticationException(ADALError.DEVELOPER_CONTEXT_IS_NOT_PROVIDED));
         }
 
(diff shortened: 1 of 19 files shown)


weijjia:
@imakhalova You should see travis is reporting errors.

imakhalova:
It requires another PR to be merged first, then I'll merge it with dev branch.
But I'm sharing this change to see if anyone has concerns

src/src/com/microsoft/aad/adal/AuthenticationContext.java
```diff
@@ -537,7 +537,7 @@ public void startActivityForResult(Intent intent, int requestCode) {
     private String checkInputParameters(String resource, String clientId, String redirectUri,
             PromptBehavior behavior, AuthenticationCallback<AuthenticationResult> callback) {
         if (mContext == null) {
-            throw new AuthenticationException(ADALError.DEVELOPER_CONTEXT_IS_NOT_PROVIDED);
+            throw new IllegalArgumentException("context", new AuthenticationException(ADALError.DEVELOPER_CONTEXT_IS_NOT_PROVIDED));
```
agamemnum: Hm. Why do you need an inner exception in here? It's anyways a breaking change.

imakhalova: I don't actually need inner exception, just left it here for long memory)
And it's actually not breaking change, because this method is being called from async methods, which throws it is as RuntimeException (not returned in callback)

RandalliLama:
:shipit:

PR #519, opened 2016-03-02
adding missing import

weijjia:
1. adding missing import introduced by exception PR
2. update target api version and java version for tests to make it consistent with library


kpanwar:
:shipit:

PR #520, opened 2016-03-04
The redirectUri should not be checked if it is a silent request

heidijinxujia:
The broker redirectUri will be checked only if the request from client App is not silent. Otherwise the acquiretokensilent() might be interrupted by DeveloperAuthenticationException (I know it is a confusing exception name, any suggestion?)

- add the condition to go to the verifyBrokerRedirectUri()
- modify one unit test

### DIFF
--- a/src/src/com/microsoft/aad/adal/AuthenticationContext.java
+++ b/src/src/com/microsoft/aad/adal/AuthenticationContext.java
@@ -1280,7 +1280,9 @@ public class AuthenticationContext {
             //check if the redirectUri is valid
             try
             {
-                verifyBrokerRedirectUri(request);
+                if(!request.isSilent()){
+                    verifyBrokerRedirectUri(request);
+                }
             }
             catch(AuthenticationException exception)
             {
--- a/src/src/com/microsoft/aad/adal/AuthenticationContext.java
+++ b/src/src/com/microsoft/aad/adal/AuthenticationContext.java
@@ -1211,47 +1211,36 @@ public class AuthenticationContext {
         final String actualUri = getRedirectUriForBroker();
         String errMsg = "";
         
-        if(StringExtensions.IsNullOrBlank(inputUri))
-        {
+        if(StringExtensions.IsNullOrBlank(inputUri)) {
             errMsg = "The redirectUri is null or blank. "
                     + "so the redirect uri is expected to be:" + actualUri;
             Logger.e(TAG + methodName, errMsg , "", ADALError.DEVELOPER_REDIRECTURI_INVALID); 
             throw new DeveloperAuthenticationException(ADALError.DEVELOPER_REDIRECTURI_INVALID, errMsg);
-        }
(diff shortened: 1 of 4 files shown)


heidijinxujia:
@weijjia

src/src/com/microsoft/aad/adal/AuthenticationContext.java
```diff
@@ -1280,7 +1280,9 @@ private AuthenticationResult acquireTokenAfterValidation(CallbackHandler callbac
             //check if the redirectUri is valid
             try
             {
-                verifyBrokerRedirectUri(request);
+                if(!request.isSilent()){
```
weijjia: Can we add comments explaining why we're doing this?

weijjia:
it looks good to me. we should have this shipped ASAP, otherwise it will block all the ppl using 1.1.13 if they want to acquiretokensilent with broker.

heidijinxujia:
@weijjia  Do you want it shipped now? Or let Kanishk or Brent have a look at it first?

heidijinxujia:
I could not ship it without your permission. :)

heidijinxujia:
@kpanwar  @brentschmaltz Thank you!

src/src/com/microsoft/aad/adal/AuthenticationContext.java
```diff
@@ -1202,58 +1202,47 @@ private boolean promptUser(PromptBehavior prompt) {
      * redirectUri format %PREFIX://%PACKAGE_NAME/%SIGNATURE
      * 
      * @param request
-     * @throws AuthenticationException
+     * @throws DeveloperAuthenticationException
      * @return true if the redirectUri is valid or fail and throw the AuthenticationException
      */
-    private boolean verifyBrokerRedirectUri(final AuthenticationRequest request) {   
+    private boolean verifyBrokerRedirectUri(final AuthenticationRequest request) throws DeveloperAuthenticationException {
```
kpanwar: Did we decide to use DeveloperAuthenticationException or still not confirmed?

heidijinxujia: not confirmed, I'm considering to change it back to AuthenticationException

kpanwar:
:shipit:

imakhalova:
Nay I ask why redirectUri don't need to be checked for silent request? What do you mean by interrupted? You catch this exception somewhere internally...

weijjia:
For silent request, it's just a post request to a sever, and today we don't send redirect uri(server side also don't ask for it). Since we don't send it, and we also don't ask for developer to provide the redirect uri for silent request, if it's we check it for silent request, then we're basically failing every silent request to broker. 
I'm also not quite sure what Interrupted means here, @heidijinxujia Can you please chime in? But the whole idea here, if developer set useBroker, we'll calculate the redirect uri(to talk to broker, the redirect uri need to be in a strict format msauth://packagename/Base64encodedSignature), if the redirect uri doesn't match what we expect, we'll return the error back to developer.

PR #522, opened 2016-03-04
 Remove the skipBroker flag and change it to a useBroker flag 

heidijinxujia:
Remove the skipBroker flag and change it to a useBroker flag to make broker more obvious

### DIFF
--- a/src/src/com/microsoft/aad/adal/AuthenticationSettings.java
+++ b/src/src/com/microsoft/aad/adal/AuthenticationSettings.java
@@ -44,6 +44,9 @@ public enum AuthenticationSettings {
      */
     private String mSharedPrefPackageName;
 
+    /**
+     * @deprecated As of release 1.1.14, replaced by {@link #mUseBroker}
+     */
     @Deprecated private boolean mSkipBroker = false;
     private boolean mUseBroker = true;
 
@@ -159,20 +162,16 @@ public enum AuthenticationSettings {
     }
 
     /**
-     * Skip broker usage.
-     * 
-     * @return true if broker is not used.
+     * @deprecated As of release 1.1.14, replaced by {@link #getUseBroker()}
      */
-    public boolean getSkipBroker() {
+    @Deprecated public boolean getSkipBroker() {
         return mSkipBroker;
     }
 
     /**
-     * Sets flag to skip broker.
-     * 
-     * @param skip True to not use broker
+     * @deprecated As of release 1.1.14, replaced by {@link #setUseBroker()}
      */
-    public void setSkipBroker(boolean skip) {
+    @Deprecated public void setSkipBroker(boolean skip) {
         mSkipBroker = skip;
     }
 
--- a/src/src/com/microsoft/aad/adal/AuthenticationSettings.java
+++ b/src/src/com/microsoft/aad/adal/AuthenticationSettings.java
@@ -44,7 +44,8 @@ public enum AuthenticationSettings {
      */
     private String mSharedPrefPackageName;
 
-    private boolean mSkipBroker = false;
+    @Deprecated private boolean mSkipBroker = false;
+    private boolean mUseBroker = true;
 
     /**
      * Expiration buffer in seconds.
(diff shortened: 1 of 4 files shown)


heidijinxujia:
@weijjia Please have a look.

src/src/com/microsoft/aad/adal/AuthenticationSettings.java
```diff
@@ -44,7 +44,11 @@
      */
     private String mSharedPrefPackageName;
 
-    private boolean mSkipBroker = false;
+    /**
+     * @deprecated As of release 1.1.14, replaced by {@link #mUseBroker}
+     */
+    @Deprecated private boolean mSkipBroker = false;
+    private boolean mUseBroker = true;
```
weijjia: nit: put it in a new line

src/src/com/microsoft/aad/adal/AuthenticationSettings.java
```diff
@@ -158,24 +162,38 @@ public void setActivityPackageName(String activityPackageName) {
...
     }
 
     /**
-     * Sets flag to skip broker.
-     * 
-     * @param skip True to not use broker
+     * @deprecated As of release 1.1.14, replaced by {@link #setUseBroker()}
      */
-    public void setSkipBroker(boolean skip) {
+    @Deprecated public void setSkipBroker(boolean skip) {
```
weijjia: Well, what happen then if user still use the deprecated method? we need to be backward compatible, right?

src/src/com/microsoft/aad/adal/AuthenticationSettings.java
weijjia: do we want to use broker by default?

src/src/com/microsoft/aad/adal/AuthenticationSettings.java
```diff
@@ -44,7 +44,8 @@
      */
     private String mSharedPrefPackageName;
 
-    private boolean mSkipBroker = false;
+    @Deprecated private boolean mSkipBroker = false;
```
imakhalova: Have you tried to use refactoring/rename variable?
deprecate private variable seems useless...just remove it

imakhalova:
Why do you need this change? What problem does it solve?

src/src/com/microsoft/aad/adal/AuthenticationSettings.java
imakhalova: +1

src/src/com/microsoft/aad/adal/AuthenticationSettings.java
heidijinxujia: I have one question, why we set default skipBroker as false? @weijjia

weijjia: yup, today we set skipBroker to be false to use broker by default. I'm not quite sure if this is also something we want to change.

heidijinxujia: OK. Then this is why I set the useBroker to true by default.

src/src/com/microsoft/aad/adal/AuthenticationSettings.java
heidijinxujia: Do you think it will be good that we just use useBroker in setSkipBroker and getSkipBroker instead of @Deprecated?

src/src/com/microsoft/aad/adal/AuthenticationSettings.java
heidijinxujia: @weijjia

src/src/com/microsoft/aad/adal/AuthenticationSettings.java
```diff
@@ -158,24 +158,38 @@ public void setActivityPackageName(String activityPackageName) {
...
+     * @return true if broker is used.
      */
-    public void setSkipBroker(boolean skip) {
-        mSkipBroker = skip;
+    public boolean getUseBroker() {
+        return mUseBroker;
     }
 
     /**
+     * Set flag to use or skip broker.
```
weijjia: Can we put what is the default value for use broker in the comment?

src/src/com/microsoft/aad/adal/AuthenticationSettings.java
```diff
@@ -44,7 +44,7 @@
      */
     private String mSharedPrefPackageName;
 
-    private boolean mSkipBroker = false;
+    private boolean mUseBroker = true;
```
weijjia: my first thought on this is, if i want to use broker i'll call setUseBroker, and it will be true. 
I think the default should be false.

weijjia:
My only other doubt is the default value for useBroker, I think it should be false and caller needs to set it if they want to use broker. 

Besides, i think i don't have any other comments.

heidijinxujia:
@kpanwar  @brentschmaltz  Could you have a look on this? I need to merge and release it by today. Thank you very much!

Issue #530, opened 2016-03-08
Address remaining comments in PR 496


weijjia:
1. ITokenCache support Iterable<T>
2. For userId comparison, invert compare to simplify code.

PR #531, opened 2016-03-08
address comments left in PR 496 and update tests

weijjia:
For issue #530, address comments left in PR 496
1) support iterable<T> in ITokenCacheStore
2) For userId comparison, invert compare to simplify code.
3) nit: format, adding space for final String errorInfo = authResult == null ? "" : authResult.getErrorLogInfo();

### DIFF
--- a/src/src/com/microsoft/aad/adal/AuthenticationContext.java
+++ b/src/src/com/microsoft/aad/adal/AuthenticationContext.java
@@ -1,20 +1,25 @@
-// Copyright © Microsoft Open Technologies, Inc.
+// Copyright (c) Microsoft Corporation.
+// All rights reserved.
 //
-// All Rights Reserved
+// This code is licensed under the MIT License.
 //
-// Licensed under the Apache License, Version 2.0 (the "License");
-// you may not use this file except in compliance with the License.
-// You may obtain a copy of the License at
+// Permission is hereby granted, free of charge, to any person obtaining a copy
+// of this software and associated documentation files(the "Software"), to deal
+// in the Software without restriction, including without limitation the rights
+// to use, copy, modify, merge, publish, distribute, sublicense, and / or sell
+// copies of the Software, and to permit persons to whom the Software is
+// furnished to do so, subject to the following conditions :
 //
-// http://www.apache.org/licenses/LICENSE-2.0
+// The above copyright notice and this permission notice shall be included in
+// all copies or substantial portions of the Software.
 //
-// THIS CODE IS PROVIDED *AS IS* BASIS, WITHOUT WARRANTIES OR CONDITIONS
-// OF ANY KIND, EITHER EXPRESS OR IMPLIED, INCLUDING WITHOUT LIMITATION
-// ANY IMPLIED WARRANTIES OR CONDITIONS OF TITLE, FITNESS FOR A
-// PARTICULAR PURPOSE, MERCHANTABILITY OR NON-INFRINGEMENT.
-//
-// See the Apache License, Version 2.0 for the specific language
-// governing permissions and limitations under the License.
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+// IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
 
 package com.microsoft.aad.adal;
 
@@ -134,8 +139,7 @@ public class AuthenticationContext {
      * @throws NoSuchAlgorithmException Encryption Algorithm does not exist in
      *             the device. Please see the log record for details.
      */
-    public AuthenticationContext(Context appContext, String authority, boolean validateAuthority)
-            throws NoSuchAlgorithmException, NoSuchPaddingException {
+    public AuthenticationContext(Context appContext, String authority, boolean validateAuthority) {
         // Fixes are required for SDK 16-18
         // The fixes need to be applied before any use of Java Cryptography
         // Architecture primitives. Default cache uses encryption
@@ -554,7 +558,7 @@ public class AuthenticationContext {
     private String checkInputParameters(String resource, String clientId, String redirectUri,
             PromptBehavior behavior, AuthenticationCallback<AuthenticationResult> callback) {
         if (mContext == null) {
-            throw new AuthenticationException(ADALError.DEVELOPER_CONTEXT_IS_NOT_PROVIDED);
+            throw new IllegalArgumentException("context", new AuthenticationException(ADALError.DEVELOPER_CONTEXT_IS_NOT_PROVIDED));
         }
 
         if (StringExtensions.IsNullOrBlank(resource)) {
@@ -879,7 +883,7 @@ public class AuthenticationContext {
                                     result = oauthRequest.getToken(endingUrl);
                                     Logger.v(TAG, "OnActivityResult processed the result. "
                                             + authenticationRequest.getLogInfo());
-                                } catch (IOException | AuthenticationServerProtocolException exc) {
+                                } catch (IOException | AuthenticationException exc) {
                                     String msg = "Error in processing code to get token. "
                                             + authenticationRequest.getLogInfo() + correlationInfo;
                                     Logger.e(TAG, msg,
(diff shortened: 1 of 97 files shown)


sdabbiru:
:shipit:

weijjia:
@imakhalova @agamemnum Can you guys take a look at the above change? This is basically to address comments @agamemnum left.

src/src/com/microsoft/aad/adal/DefaultTokenCacheStore.java
```diff
@@ -235,7 +235,7 @@ public void removeAll() {
             }
         }
 
-        return tokens.iterator();
+        return tokens;
```
brentschmaltz: From the code it appears that 'nulls' may be in the Iterable<TokenCacheItem> returned. Because we create the collection as ArrayList<TokenCacheItem> tokens = new ArrayList<TokenCacheItem>(results.values().size()); If decrypt fails, we will not add the tokenCacheItem.

src/src/com/microsoft/aad/adal/DefaultTokenCacheStore.java
```diff
@@ -245,13 +245,15 @@ public void removeAll() {
...
-        Iterator<TokenCacheItem> results = this.getAll();
+        Iterable<TokenCacheItem> results = this.getAll();
         HashSet<String> users = new HashSet<String>();
-
-        while (results.hasNext()) {
-            TokenCacheItem item = results.next();
-            if (item.getUserInfo() != null && !users.contains(item.getUserInfo().getUserId())) {
-                users.add(item.getUserInfo().getUserId());
+        
+        if (results != null) {
```
imakhalova: Why do you always check it for null, your implementation of getAll() will never return null, if null is unexpected don't check for null

tests/Functional/src/com/microsoft/aad/adal/test/AuthenticationContextTest.java
```diff
@@ -1159,7 +1158,7 @@ public void testScenario_LoginHint_IdToken_Different() throws InterruptedExcepti
         // Same call with correct upn will return from cache
         final CountDownLatch signalCallback2 = new CountDownLatch(1);
         callback.mSignal = signalCallback2;
-        context.acquireToken(null, "resource", "clientid", "redirectUri", idtoken.upn, callback);
+        context.acquireToken(testActivity, "resource", "clientid", "redirectUri", idtoken.upn, callback);
```
imakhalova: I think I introduces it, so let me actually fix it, in separate PR

src/src/com/microsoft/aad/adal/AuthenticationContext.java
```diff
@@ -1639,19 +1637,17 @@ private TokenCacheItem findFamilyItemForUser(final UserIdentifierType userIdenti
             return null;
         }
 
-        Iterator<TokenCacheItem> allItems = mTokenCacheStore.getAll();
-        if (allItems == null || !allItems.hasNext()) {
+        Iterable<TokenCacheItem> allItems = mTokenCacheStore.getAll();
+        if (allItems == null || !allItems.iterator().hasNext()) {
```
imakhalova: For your purposes I think that getAll should return actually Collection<> (Not just iterable)...
It has method isEmpty()

src/src/com/microsoft/aad/adal/AuthenticationContext.java
```diff
@@ -241,7 +239,7 @@ public boolean contains(String key) {
                 }
 
                 @Override
-                public Iterator<TokenCacheItem> getAll() {
+                public Iterable<TokenCacheItem> getAll() {
```
agamemnum: I meant you don't need this method at all.. `ITokenCacheStore` should be `Iterable< TokenCacheItem >`

src/src/com/microsoft/aad/adal/AuthenticationContext.java
agamemnum: IMHO -  `ITokenCacheStore` need to have `hasTokes` or something like that. `ITokenCacheStore` should become just a collection on it's own.

tests/Functional/src/com/microsoft/aad/adal/test/AuthenticationContextTest.java
weijjia: :) Sure. I saw the tests failed and then fixed them. Will let you know next time. Oh, by the way, i'm looking at the tests failure you mentioned in the email. That should be the service we are hitting for testing purpose, but they're no longer working I think. I'm going to ask around who set them up and who is maintaining them.

src/src/com/microsoft/aad/adal/AuthenticationContext.java
weijjia: The original reason I make it Iterator is mainly because DefaultTokenCache not only implement the ITokenCacheStore but also the ITokenStoreQuery, ITokenStoreQuery has the getAll, and it returns the Iterator. 

I agree with you guys, the code needs to be improved a lot. The interface for token cache should be able to do the query. There may be some other clients out there providing their customized cache, and they need to implement the interface, I just want to reduce the changes that could happen to them as little as possible.

src/src/com/microsoft/aad/adal/AuthenticationContext.java
weijjia: Got it now, Thanks. But I don't think I want to cover it at this moment. That could be the future improvement as what I mentioned above. What do you think?

src/src/com/microsoft/aad/adal/DefaultTokenCacheStore.java
weijjia: Well, I think it's always safe to check for null.

src/src/com/microsoft/aad/adal/DefaultTokenCacheStore.java
weijjia: I don't get it. How null could be possible? the tokens could be empty if nothing added, and then it would be an empty iterable.

src/src/com/microsoft/aad/adal/DefaultTokenCacheStore.java
imakhalova: Yes, it's safe when object can be null, and in this case it actually superficial code. (extra 2 lines + extra brackets + extra tabs for inner code)...

weijjia: hashset users will never be null. Well, I got your point, and will make a update.

src/src/com/microsoft/aad/adal/AuthenticationContext.java
agamemnum: OK.

src/src/com/microsoft/aad/adal/DefaultTokenCacheStore.java
```diff
@@ -287,14 +287,15 @@ public void removeAll() {
      */
     @Override
     public ArrayList<TokenCacheItem> getTokensForUser(String userid) {
```
agamemnum: overall: All interfaces should use generic interface like `List<>` or `Set<>` and not concrete implementations

agamemnum: nit: userId

src/src/com/microsoft/aad/adal/DefaultTokenCacheStore.java
brentschmaltz: The list is created of a fixed size
ArrayList<TokenCacheItem> tokens = new ArrayList<TokenCacheItem>(results.values().size());

Addition is conditional
if (decryptedValue != null)

Will null be returned for any item not set?

src/src/com/microsoft/aad/adal/DefaultTokenCacheStore.java
brentschmaltz: @agamemnum do you have a suggested replacement for 'userid'?

src/src/com/microsoft/aad/adal/DefaultTokenCacheStore.java
imakhalova: I think you got confused by argument to ArrayList constructor. It's not the size of ArrayList, it's capacity (how much memory to allocate in advance).
But if you try to check tokens.size(), you'll get 0. There are no elements in the list (no null elements there)

brentschmaltz: That clears it up then.
thanks.

src/src/com/microsoft/aad/adal/DefaultTokenCacheStore.java
agamemnum: @brentschmaltz No I just saying that `I` should be capitalized :-)

weijjia: Agree, interface should not use the concrete implementation as return type.

weijjia:
@agamemnum @imakhalova I now understand that @agamemnum  want the ItokenCacheStore to be a collection of its own. But I really don't want to change it that much at this moment. That will be a breaking change for client app implementing their own cache. If you guys are fine, i'm going to keep the current behavior, still returning Iterator, this at lease won't cause another change on the client app using customized cache. What do you guys think?

imakhalova:
Yes, I see that getCache().getAll is actually public methods...So yes, it potentially can break users

PR #535, opened 2016-03-12
fix test app because of exception change

weijjia:
update test app because of the exception change.


weijjia:
@imakhalova You can also take a look at this one. Because the thrown exception change, we need to make corresponding update also in the test.

imakhalova:
Change is fine, merge it.
But I think we need to spend some time to fix our infrastructure.
Because our Travis build doesn't check if test project builds. Can you please take a look at this, why it always passes even if tests won't compile. It will save us lots of dev time. 
Thank you,
Irina

weijjia:
It's interesting. If you see the console log for travis, it actually does compile the test project.

heidijinxujia:
It is good in my view.

heidijinxujia:
:shipit:

PR #536, opened 2016-03-14
use real Spec.Builder instead of reflective methods

imakhalova:
KeyPairGeneratorSpec was deprecated since API 23,
and we were getting javax.crypto.BadPaddingException: error:0407109F:rsa routines:RSA_padding_check_PKCS1_type_2:pkcs decoding error 
on android M devices because our encryption type requires this flag 
setEncryptionPaddings(KeyProperties.ENCRYPTION_PADDING_RSA_PKCS1)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,12 +5,12 @@ android:
     - build-tools-21.1.1
 env:
     matrix:
-      - ANDROID_SDKS=android-21 ANDROID_TARGET=android-21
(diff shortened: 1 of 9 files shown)


imakhalova:
Covered this in another PR (#539 StorageHelper is singleton) .
But if you're going to upgrade android sdk to latest one, you can apply this fix later

PR #538, opened 2016-03-14
update for policheck results

weijjia:
Update for policheck results.

### DIFF
--- a/tests/Functional/src/com/microsoft/aad/adal/test/AuthenticationParamsTests.java
+++ b/tests/Functional/src/com/microsoft/aad/adal/test/AuthenticationParamsTests.java
@@ -196,7 +196,7 @@ public class AuthenticationParamsTests extends AndroidTestHelper {
                 AuthenticationParameters.AUTH_HEADER_INVALID_FORMAT);
 
         callParseResponseForException(
-                new HttpWebResponse(401, null, getHeader("WWW-Authenticate", "Bearer foo ")),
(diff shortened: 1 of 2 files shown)


tests/Functional/src/com/microsoft/aad/adal/test/AuthenticationParamsTests.java
```diff
@@ -196,7 +196,7 @@ public void testParseResponseNegative() throws IllegalArgumentException,
                 AuthenticationParameters.AUTH_HEADER_INVALID_FORMAT);
 
         callParseResponseForException(
-                new HttpWebResponse(401, null, getHeader("WWW-Authenticate", "Bearer foo ")),
+                new HttpWebResponse(401, null, getHeader("WWW-Authenticate", "Bearer test ")),
```
heidijinxujia: For Line 118 Should it also be changed to "test" instead of  "foo"?

weijjia: The policheck doesn't complain about it so I didn't change.

heidijinxujia: Okay, I just think it might confused people. Everything else looks good in my view.

weijjia: why it confuse people? I'm just do the change based on the policheck result.

heidijinxujia:
:shipit:

PR #539, opened 2016-03-14
Cleanup around StorageHelper and its tests

imakhalova:
If SecretKey is null or failed to generate throw IlligalStateException
(we should not allow null secret keys)
Rewrote all tests around encryption to use real methods instead of reflective ones

### DIFF
--- a/src/src/com/microsoft/aad/adal/AuthenticationActivity.java
+++ b/src/src/com/microsoft/aad/adal/AuthenticationActivity.java
@@ -105,7 +105,7 @@ public class AuthenticationActivity extends Activity {
     private String mQueryParameters;
 
     private boolean mPkeyAuthRedirect = false;
-    private StorageHelper mHelper;
+    private StorageHelper mStorageHelper;
 
     // Broadcast receiver is needed to cancel outstanding AuthenticationActivity
     // for this AuthenticationContext since each instance of context can have
@@ -269,6 +269,7 @@ public class AuthenticationActivity extends Activity {
                 " device:" + android.os.Build.VERSION.RELEASE + " " + android.os.Build.MANUFACTURER
                         + android.os.Build.MODEL);
 
+        mStorageHelper = new StorageHelper(getApplicationContext());
         setupWebView(mRedirectUrl, mQueryParameters, mAuthRequest);
         
         if (savedInstanceState == null) {
@@ -283,8 +284,6 @@ public class AuthenticationActivity extends Activity {
         } else {
             Logger.v(TAG, "Reuse webview");
         }
-
-        mHelper = new StorageHelper(getApplicationContext());
     }
 
     private boolean isCallerBrokerInstaller() {
@@ -847,7 +846,7 @@ public class AuthenticationActivity extends Activity {
                 appIdList = "";
             } else {
                 try {
-                    appIdList = mHelper.decrypt(appIdList);
+                    appIdList = mStorageHelper.decrypt(appIdList);
                 } catch (GeneralSecurityException | IOException ex) {
                     Logger.e(TAG, "appUIDList failed to decrypt", "appIdList:" + appIdList,
                             ADALError.ENCRYPTION_FAILED, ex);
@@ -860,7 +859,7 @@ public class AuthenticationActivity extends Activity {
             if (!appIdList.contains(AuthenticationConstants.Broker.USERDATA_UID_KEY
                     + mAppCallingUID)) {
                 Logger.i(TAG, "Account has new calling UID:" + mAppCallingUID, "");
-                String encryptedValue = mHelper.encrypt(appIdList
+                String encryptedValue = mStorageHelper.encrypt(appIdList
                         + AuthenticationConstants.Broker.USERDATA_UID_KEY
                         + mAppCallingUID);
                 mAccountManager
@@ -937,7 +936,7 @@ public class AuthenticationActivity extends Activity {
 
             TokenCacheItem item = new TokenCacheItem(mRequest, result.taskResult, false);
             String json = gson.toJson(item);
-            String encrypted = mHelper.encrypt(json);
+            String encrypted = mStorageHelper.encrypt(json);
 
             // Single user and cache is stored per account
             String key = CacheKey.createCacheKey(mRequest, null);
@@ -951,7 +950,7 @@ public class AuthenticationActivity extends Activity {
                 // ADAL stores MRRT refresh token separately
                 TokenCacheItem itemMRRT = new TokenCacheItem(mRequest, result.taskResult, true);
                 json = gson.toJson(itemMRRT);
-                encrypted = mHelper.encrypt(json);
+                encrypted = mStorageHelper.encrypt(json);
                 key = CacheKey.createMultiResourceRefreshTokenKey(mRequest, null);
                 saveCacheKey(key, newaccount, mAppCallingUID);
                 mAccountManager.setUserData(
--- a/src/src/com/microsoft/aad/adal/AuthenticationActivity.java
+++ b/src/src/com/microsoft/aad/adal/AuthenticationActivity.java
@@ -105,6 +105,7 @@ public class AuthenticationActivity extends Activity {
     private String mQueryParameters;
 
     private boolean mPkeyAuthRedirect = false;
+    private StorageHelper mHelper;
 
     // Broadcast receiver is needed to cancel outstanding AuthenticationActivity
     // for this AuthenticationContext since each instance of context can have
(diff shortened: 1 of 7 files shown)


src/src/com/microsoft/aad/adal/DefaultTokenCacheStore.java
```diff
@@ -99,18 +95,23 @@ public DefaultTokenCacheStore(Context context) {
             throw new IllegalArgumentException("Context is null");
```
agamemnum: nit: move it to the first line if null throw right away. Just simpler to read code

src/src/com/microsoft/aad/adal/DefaultTokenCacheStore.java
```diff
@@ -99,18 +95,23 @@ public DefaultTokenCacheStore(Context context) {
...
         }
 
-        synchronized (sLock) {
-            if (sHelper == null) {
-                Logger.v(TAG, "Started to initialize storage helper");
-                sHelper = new StorageHelper(mContext);
-                Logger.v(TAG, "Finished to initialize storage helper");
-            }
+        try {
```
agamemnum: nit: formatting

src/src/com/microsoft/aad/adal/StorageHelper.java
```diff
@@ -136,36 +159,38 @@ public StorageHelper(Context ctx) {
...
-                    return;
-                } catch (IOException | GeneralSecurityException e) {
-                    Logger.e(TAG, "Failed to get private key from AndroidKeyStore", "",
-                            ADALError.ANDROIDKEYSTORE_FAILED, e);
-                }
-            }
+            return mKey;
 
+        final byte[] secretKeyData = AuthenticationSettings.INSTANCE.getSecretKeyData();
+        if (secretKeyData != null) {
```
agamemnum: nit: generally more readable code would be something like
```
if(secretKeyData == null && Build.VERSION.SDK_INT < 18) 
       throw new IllegalStateException("Secret key must be provided for API < 18. " +
+                    "Use AuthenticationSettings.INSTANCE.setSecretKey()");


And then just normal code below. if key present or not. 
```

src/src/com/microsoft/aad/adal/StorageHelper.java
```diff
@@ -136,36 +159,38 @@ public StorageHelper(Context ctx) {
      * @return
      * @throws NoSuchAlgorithmException
      */
-    final private void loadSecretKeyForAPI() throws NoSuchAlgorithmException {
+    public SecretKey loadSecretKeyForAPI(Context context) throws IOException, GeneralSecurityException {
         // Loading key only once for performance. If API is upgraded, it will
         // restart the device anyway. It will load the correct key for new API. 
         if (mKey != null && mMacKey != null)
```
agamemnum: Do we need to worry about thread safety? It looks like code wasn't really handling it before.

src/src/com/microsoft/aad/adal/StorageHelper.java
```diff
@@ -497,53 +515,18 @@ private synchronized KeyPair getKeyPairFromAndroidKeyStore() throws GeneralSecur
...
-            }
-        } catch (NoSuchMethodException e) {
-            Logger.e(TAG, "android.security.KeyPairGeneratorSpec.Builder is not found", "",
-                    ADALError.ANDROIDKEYSTORE_KEYPAIR_GENERATOR_FAILED, e);
-        }
 
-        return null;
+    @TargetApi(18)
+    private AlgorithmParameterSpec getKeyPairGeneratorSpec(Context context, Date start, Date end) {
+        String certInfo = String.format("CN=%s, OU=%s", KEY_STORE_CERT_ALIAS,
```
agamemnum: Locale!

src/src/com/microsoft/aad/adal/DefaultTokenCacheStore.java
```diff
@@ -99,18 +95,23 @@ public DefaultTokenCacheStore(Context context) {
...
     }
 
+    /**
+     * Method that allows to mock StorageHelper class and use custom encryption in UTs
+     * @return
+     */
+    protected StorageHelper getStorageHelper() {
+        return StorageHelper.getInstance();
+    }
     private String encrypt(String value) {
```
agamemnum: nit: space

src/src/com/microsoft/aad/adal/StorageHelper.java
```diff
@@ -400,20 +422,20 @@ final private SecretKey generateSecretKey() throws NoSuchAlgorithmException {
...
         // Loading file and unwrapping this key is causing performance issue.
-        if (sSecretKeyFromAndroidKeyStore != null) {
-            return sSecretKeyFromAndroidKeyStore;
+        if (mSecretKeyFromAndroidKeyStore != null) {
+            return mSecretKeyFromAndroidKeyStore;
         }
 
         // Store secret key in a file after wrapping
-        File keyFile = new File(mContext.getDir(mContext.getPackageName(), Context.MODE_PRIVATE),
+        File keyFile = new File(context.getDir(context.getPackageName(), Context.MODE_PRIVATE),
```
agamemnum: It would be great to have a function - "getKeyFile" and do not copy-paste same logic to format file name several times...

src/src/com/microsoft/aad/adal/StorageHelper.java
imakhalova: I'll make method synchronized

src/src/com/microsoft/aad/adal/StorageHelper.java
imakhalova: Right!

src/src/com/microsoft/aad/adal/AuthenticationActivity.java
```diff
@@ -282,6 +283,8 @@ public void run() {
         } else {
             Logger.v(TAG, "Reuse webview");
         }
+
+        mHelper = new StorageHelper(getApplicationContext());
```
weijjia: nit: can we rename it to mStorageHelpler?

weijjia: AuthenticationActivity could be called in two mode, with broker or in local flow. If it's in local flow, we then initialize the  variable, but no one is using it.

src/src/com/microsoft/aad/adal/DefaultTokenCacheStore.java
```diff
@@ -53,52 +53,57 @@
...
             }
+        }
+        mPrefs = mContext.getSharedPreferences(SHARED_PREFERENCE_NAME, Activity.MODE_PRIVATE);
+        if (mPrefs == null) {
+            throw new IllegalStateException(ADALError.DEVICE_SHARED_PREF_IS_NOT_AVAILABLE.getDescription());
+        }
 
-        } else {
-            throw new IllegalArgumentException("Context is null");
+        try {
```
weijjia: why you want to read the key here?

imakhalova: because if we can't load secretkey at this point we won't be able to encrypt/decrypt anything and this cache becomes useless...

src/src/com/microsoft/aad/adal/AuthenticationActivity.java
imakhalova: ok, I'll rename it.
No harm if we have initialized storage helper... It might reuse the same activity later for broker flow

src/src/com/microsoft/aad/adal/DefaultTokenCacheStore.java
```diff
@@ -53,52 +53,57 @@
...
+        } catch (IOException | GeneralSecurityException e) {
+            throw new IllegalStateException("Failed to get private key from AndroidKeyStore", e);
         }
+    }
 
+    /**
+     * Method that allows to mock StorageHelper class and use custom encryption in UTs
+     * @return
+     */
+    protected StorageHelper getStorageHelper() {
```
agamemnum: nit: just make it syncronized?

src/src/com/microsoft/aad/adal/DefaultTokenCacheStore.java
imakhalova: It's like what you're going to do if mPrefs are not currently available...?

src/src/com/microsoft/aad/adal/DefaultTokenCacheStore.java
imakhalova: I don't see a big difference, but ok

src/src/com/microsoft/aad/adal/DefaultTokenCacheStore.java
imakhalova: @weijjia, what is your concern?

weijjia:
:shipit:

PR #545, opened 2016-03-16
Merge current dev branch to gradle project

imakhalova:
Trying to keep android studio project up to date, so it will simplify switch. I hope that you will switch in near future, because Android Studio provides lot of useful features for android developers.


imakhalova:
Seems like the switch to gradle build is in progress

PR #546, opened 2016-03-16
Fixing some UTF encoding issues

imakhalova:
Noticed that there are some invisible characters in 2 files, most probably files weren't saved in UTF8


imakhalova:
Recreated pull request against dev branch

Issue #521, opened 2016-03-04
Update to support muti-user for company portal


weijjia:
ADAL is assuming only azure authenticator app is going to provide multi-user support for the new prt flow in broker. We need to change it since company portal is also going to have multi-user support.

addev-ashish:
Fix already merged, closed.

PR #550, opened 2016-03-18
allow company portal to provide multi-user support

weijjia:
issue #521 
With the new broker with PRT support, both broker apps (company portal and azure authenticator) will provide multi-user support.

### DIFF
--- a/src/src/com/microsoft/aad/adal/BrokerProxy.java
+++ b/src/src/com/microsoft/aad/adal/BrokerProxy.java
@@ -103,7 +103,8 @@ class BrokerProxy implements IBrokerProxy {
         // authenticator
         // 4- signature of the broker is valid
         // 5- account exists
-        return AuthenticationSettings.INSTANCE.getUseBroker() && verifyManifestPermissions()
+        return AuthenticationSettings.INSTANCE.getUseBroker() 
+                && verifyManifestPermissions()
                 && checkAccount(mAcctManager, "", "")
                 && !packageName.equalsIgnoreCase(AuthenticationSettings.INSTANCE.getBrokerPackageName())
                 && !packageName.equalsIgnoreCase(AuthenticationConstants.Broker.AZURE_AUTHENTICATOR_APP_PACKAGE_NAME)
@@ -299,7 +300,6 @@ class BrokerProxy implements IBrokerProxy {
                 Logger.v(TAG, "Broker doesn't return expire date, set it current date plus one hour");
                 final Calendar currentTime = new GregorianCalendar();
                 currentTime.add(Calendar.SECOND, AuthenticationConstants.DEFAULT_EXPIRATION_TIME_SEC);
-                ;
                 expires = currentTime.getTime();
             } else {
                 expires = new Date(bundleResult.getLong(AuthenticationConstants.Broker.ACCOUNT_EXPIREDATE));
--- a/src/src/com/microsoft/aad/adal/BrokerProxy.java
+++ b/src/src/com/microsoft/aad/adal/BrokerProxy.java
@@ -472,7 +472,7 @@ class BrokerProxy implements IBrokerProxy {
                 if (authenticator.packageName
                         .equalsIgnoreCase(AuthenticationConstants.Broker.AZURE_AUTHENTICATOR_APP_PACKAGE_NAME)
                         || authenticator.packageName
-                                .equalsIgnoreCase(AuthenticationConstants.Broker.COMPANY_PORTAL_PACKAGE_NAME)
+                                .equalsIgnoreCase(AuthenticationConstants.Broker.COMPANY_PORTAL_APP_PACKAGE_NAME)
                         || authenticator.packageName
                                 .equalsIgnoreCase(AuthenticationSettings.INSTANCE.getBrokerPackageName())) {
                     // Existing broker logic only connects to broker for token
--- a/src/src/com/microsoft/aad/adal/BrokerProxy.java
+++ b/src/src/com/microsoft/aad/adal/BrokerProxy.java
@@ -467,8 +467,8 @@ class BrokerProxy implements IBrokerProxy {
                 Account[] accountList = mAcctManager
                         .getAccountsByType(AuthenticationConstants.Broker.BROKER_ACCOUNT_TYPE);
 
-                // Authenticator installed from Company portal
-                // This supports only one account
+                // For new broker with PRT support, both company portal and
+                // azure authenticator will be able to support multi-user.
                 if (authenticator.packageName
                         .equalsIgnoreCase(AuthenticationConstants.Broker.AZURE_AUTHENTICATOR_APP_PACKAGE_NAME)
                         || authenticator.packageName
--- a/src/src/com/microsoft/aad/adal/BrokerProxy.java
+++ b/src/src/com/microsoft/aad/adal/BrokerProxy.java
@@ -103,16 +103,13 @@ class BrokerProxy implements IBrokerProxy {
         // authenticator
         // 4- signature of the broker is valid
         // 5- account exists
-        return AuthenticationSettings.INSTANCE.getUseBroker()
-                && verifyManifestPermissions()
+        return AuthenticationSettings.INSTANCE.getUseBroker() && verifyManifestPermissions()
                 && checkAccount(mAcctManager, "", "")
-                && !packageName.equalsIgnoreCase(AuthenticationSettings.INSTANCE
-                        .getBrokerPackageName())
-                && !packageName
-                        .equalsIgnoreCase(AuthenticationConstants.Broker.AZURE_AUTHENTICATOR_APP_PACKAGE_NAME)
+                && !packageName.equalsIgnoreCase(AuthenticationSettings.INSTANCE.getBrokerPackageName())
+                && !packageName.equalsIgnoreCase(AuthenticationConstants.Broker.AZURE_AUTHENTICATOR_APP_PACKAGE_NAME)
                 && verifyAuthenticator(mAcctManager);
     }
-    
+
     /**
      * Do this check after other checks.
      */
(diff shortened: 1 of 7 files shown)


src/src/com/microsoft/aad/adal/BrokerProxy.java
```diff
@@ -103,16 +103,13 @@ public boolean canSwitchToBroker() {
         // authenticator
         // 4- signature of the broker is valid
         // 5- account exists
-        return AuthenticationSettings.INSTANCE.getUseBroker()
-                && verifyManifestPermissions()
+        return AuthenticationSettings.INSTANCE.getUseBroker() && verifyManifestPermissions()
                 && checkAccount(mAcctManager, "", "")
```
sdabbiru: Why do we have checkAccount call?

src/src/com/microsoft/aad/adal/BrokerProxy.java
```diff
@@ -103,16 +103,13 @@ public boolean canSwitchToBroker() {
         // authenticator
         // 4- signature of the broker is valid
         // 5- account exists
-        return AuthenticationSettings.INSTANCE.getUseBroker()
-                && verifyManifestPermissions()
+        return AuthenticationSettings.INSTANCE.getUseBroker() && verifyManifestPermissions()
```
sdabbiru: Let's have only one API call per line.

src/src/com/microsoft/aad/adal/BrokerProxy.java
```diff
@@ -304,22 +291,22 @@ private AuthenticationResult getResultFromBrokerResponse(Bundle bundleResult) th
...
             final Date expires;
             if (bundleResult.getLong(AuthenticationConstants.Broker.ACCOUNT_EXPIREDATE) == 0) {
                 Logger.v(TAG, "Broker doesn't return expire date, set it current date plus one hour");
                 final Calendar currentTime = new GregorianCalendar();
-                currentTime.add(Calendar.SECOND, AuthenticationConstants.DEFAULT_EXPIRATION_TIME_SEC);;
-                expires = currentTime.getTime(); 
-            }
-            else {
+                currentTime.add(Calendar.SECOND, AuthenticationConstants.DEFAULT_EXPIRATION_TIME_SEC);
+                ;
```
sdabbiru: ?

src/src/com/microsoft/aad/adal/BrokerProxy.java
```diff
@@ -493,29 +467,18 @@ private boolean checkAccount(final AccountManager am, String username, String un
...
-                    // Check azure authenticator and allow calls for test
-                    // versions
-                } else if (authenticator.packageName
                         .equalsIgnoreCase(AuthenticationConstants.Broker.AZURE_AUTHENTICATOR_APP_PACKAGE_NAME)
                         || authenticator.packageName
-                                .equalsIgnoreCase(AuthenticationSettings.INSTANCE
-                                        .getBrokerPackageName())) {
-
+                                .equalsIgnoreCase(AuthenticationConstants.Broker.COMPANY_PORTAL_APP_PACKAGE_NAME)
+                        || authenticator.packageName
```
sdabbiru: Is the 3rd condition really needed?

src/src/com/microsoft/aad/adal/BrokerProxy.java
```diff
@@ -493,29 +467,18 @@ private boolean checkAccount(final AccountManager am, String username, String un
...
+                                .equalsIgnoreCase(AuthenticationSettings.INSTANCE.getBrokerPackageName())) {
```
sdabbiru: Why the 3rd condition?

src/src/com/microsoft/aad/adal/BrokerProxy.java
weijjia: ha, auto-reformat found out issues in the code

src/src/com/microsoft/aad/adal/BrokerProxy.java
weijjia: I didn't change the logic around here. We always has this check for testing purpose I think.

src/src/com/microsoft/aad/adal/BrokerProxy.java
weijjia: I'll have to disable the auto-format.

src/src/com/microsoft/aad/adal/BrokerProxy.java
weijjia: 1) if it's old broker, then there has to be one account existed 2) or it's the new broker which has the support to add user through broker.

src/src/com/microsoft/aad/adal/BrokerProxy.java
```diff
@@ -146,16 +144,14 @@ public boolean canUseLocalCache() {
      */
     private boolean verifyManifestPermissions() {
         PackageManager pm = mContext.getPackageManager();
-        boolean permission = PackageManager.PERMISSION_GRANTED == pm.checkPermission(
-                "android.permission.GET_ACCOUNTS", mContext.getPackageName())
-                && PackageManager.PERMISSION_GRANTED == pm.checkPermission(
-                        "android.permission.MANAGE_ACCOUNTS", mContext.getPackageName())
-                && PackageManager.PERMISSION_GRANTED == pm.checkPermission(
-                        "android.permission.USE_CREDENTIALS", mContext.getPackageName());
+        boolean permission = PackageManager.PERMISSION_GRANTED == pm.checkPermission("android.permission.GET_ACCOUNTS",
```
imakhalova: 1) There is method ContextCompat.checkSelfPermission 
2) It seems like this permissions will be deprecated in Android N
3) Why do we need MANAGE_ACCOUNTS permission from app that isn't broker? Can it manage accounts?
4) Not sure why you moving first argument to line above, IMHO previous version showed number of arguments better.

src/src/com/microsoft/aad/adal/BrokerProxy.java
```diff
@@ -497,29 +471,18 @@ private boolean checkAccount(final AccountManager am, String username, String un
...
+                        || authenticator.packageName
+                                .equalsIgnoreCase(AuthenticationSettings.INSTANCE.getBrokerPackageName())) {
                     // Existing broker logic only connects to broker for token
                     // requests if account exists. New version can allow to
                     // add accounts through Adal.
-                    if (hasSupportToAddUserThroughBroker()) {
+                    if (hasSupportToAddUserThroughBroker(authenticator.packageName)) {
                         Logger.v(TAG, "Broker supports to add user through app");
                         return true;
                     } else if (accountList != null && accountList.length > 0) {
```
imakhalova: you need to verity all accounts, not the first one

src/src/com/microsoft/aad/adal/BrokerProxy.java
```diff
@@ -559,12 +521,15 @@ private boolean verifyAccount(Account[] accountList, String username, String uni
...
-                        + ".ui.AccountChooserActivity");
+    /**
+     * True if broker has multi-user support. ADAL is checking the existence of
+     * .ui.AccountChooserActivity which only exists in the new broker with PRT
+     * support.
+     */
+    private boolean hasSupportToAddUserThroughBroker(final String brokerPackageName) {
+        final Intent intent = new Intent();
+        intent.setPackage(brokerPackageName);
+        intent.setClassName(brokerPackageName, brokerPackageName + ".ui.AccountChooserActivity");
```
imakhalova: It's usually bad practice to hardcode class names, there are helper methods such as queryIntentActivities to detect activity that can handle intent...

src/src/com/microsoft/aad/adal/BrokerProxy.java
weijjia: For the old version of brokers, there will only be one account existed.

src/src/com/microsoft/aad/adal/BrokerProxy.java
weijjia: Manage_Accounts and Use_Credential is already deprecated in android M, so when we move to M, this will be changed. 
Well, the reason we need MANAGE_ACCOUTNS is in the case of talking to broker, adal is calling addAccounts. 
For the format change, hmm, i enable the auto-format for edited lines. I can change it back.

src/src/com/microsoft/aad/adal/BrokerProxy.java
weijjia: I agree. This is an agreement we have with broker and adal. I talked to @sdabbiru for this, we can probably try to get protocol version from broker. But for now, we'll still keep it.

src/src/com/microsoft/aad/adal/BrokerProxy.java
```diff
@@ -420,18 +404,15 @@ public Intent getIntentForBrokerActivity(final AuthenticationRequest request) {
...
-            Logger.e(TAG, "Authenticator cancels the request", "", ADALError.AUTH_FAILED_CANCELLED,
-                    e);
+            Logger.e(TAG, "Authenticator cancels the request", "", ADALError.AUTH_FAILED_CANCELLED, e);
         } catch (AuthenticatorException e) {
             //
             // TODO add retry logic since authenticator is not responding to
             // the request
-            Logger.e(TAG, "Authenticator cancels the request", "",
-                    ADALError.BROKER_AUTHENTICATOR_NOT_RESPONDING);
+            Logger.e(TAG, "Authenticator cancels the request", "", ADALError.BROKER_AUTHENTICATOR_NOT_RESPONDING);
```
sdabbiru: can we log the exception?

src/src/com/microsoft/aad/adal/BrokerProxy.java
```diff
@@ -420,18 +404,15 @@ public Intent getIntentForBrokerActivity(final AuthenticationRequest request) {
...
         } catch (IOException e) {
             // Authenticator gets problem from webrequest or file read/write
-            Logger.e(TAG, "Authenticator cancels the request", "",
-                    ADALError.BROKER_AUTHENTICATOR_IO_EXCEPTION);
+            Logger.e(TAG, "Authenticator cancels the request", "", ADALError.BROKER_AUTHENTICATOR_IO_EXCEPTION);
```
sdabbiru: can we log the exception?

sdabbiru:
:shipit:

Issue #472, opened 2016-01-12
Library manifest should not set android:allowBackup to true 


twentynothing:
I don't think library's manifest should set this to be true. This causes a merging error when the application has it set to false.

Ricalo:
I agree. The application should be able to set that attribute to fit its needs.
I filed a similar issue on the Office-365-SDK-for-Android officedev/Office-365-SDK-for-Android#96

A possible workaround would be to use a `'tools:replace="android:allowBackup"'` in the `<application>` element in your Android manifest, but that's just a workaround.

weijjia:
You can always use tool:replace to override what's set in the library.

twentynothing:
Sure. "tool:replace " works. 
However library setting "android:allowBackup=true" is an implicit change to application behavior. 
If the library absolutely needs this to work, then it should be documented and not allow app overwriting.
If not, why setting it in the first place?

bluehallu:
+1

If tools:replace works then it means the library doesn't actually need this to work!

weijjia:
I agree with you, as a library, we shouldn't care about setting it. I'll send a PR related to this.

addev-ashish:
Looks like this was fixed.

PR #559, opened 2016-03-25
remove allowBackup from manifest

weijjia:
issue #472 
As a library, we shouldn't care setting this. It should be up to the application itself for allowing backup or not.


kpanwar:
:shipit:

PR #560, opened 2016-03-27
bugfix: Set correct key for multifamily refresh token

imakhalova:
1)  Set correct key for multifamily refresh token so it will save it by correct key in cache
2) For family tokens do not take key for the same client (otherwise we might try just another token for this clientId)
3) Never remove Item from cache, because sometimes we've got server errors which can remove good token

### DIFF
--- a/src/src/com/microsoft/aad/adal/AuthenticationContext.java
+++ b/src/src/com/microsoft/aad/adal/AuthenticationContext.java
@@ -1590,10 +1590,9 @@ public class AuthenticationContext {
         
         String mTenantId;
 
-        public RefreshItem(String keyInCache, AuthenticationRequest request, TokenCacheItem item,
-                boolean multiResource) {
+        public RefreshItem(String keyInCache, AuthenticationRequest request, TokenCacheItem item) {
             mKey = keyInCache;
-            mMultiResource = multiResource;
+            mMultiResource = item.getIsMultiResourceRefreshToken();
 
             if (item != null) {
                 mRefreshToken = item.getRefreshToken();
@@ -1619,35 +1618,34 @@ public class AuthenticationContext {
         final String methodName = ":getRefreshToken";
         RefreshItem refreshItem = null;
         if (mTokenCacheStore != null) {
-            boolean multiResource = false;
-            // target refreshToken for this resource first. CacheKey will
-            // include the resourceId in the cachekey
-            Logger.v(TAG, "Looking for regular refresh token");
             String userId = request.getUserId();
             if (StringExtensions.IsNullOrBlank(userId)) {
                 // acquireTokenSilent expects userid field from UserInfo
                 userId = request.getLoginHint();
             }
-            String keyUsed = CacheKey.createCacheKey(request, userId);
+
+            // Target multiResource item in cache first because it's the most recent one.
+            // Cache key will not include resourceId.
+            Logger.v(TAG, "Looking for Multi Resource Refresh token");
+            String keyUsed  = CacheKey.createMultiResourceRefreshTokenKey(request, userId);
             TokenCacheItem item = mTokenCacheStore.getItem(keyUsed);
+
             if (item == null || StringExtensions.IsNullOrBlank(item.getRefreshToken())) {
-                // if not present, check multiResource item in cache. Cache key
-                // will not include resourceId in the cache key string.
-                Logger.v(TAG, "Looking for Multi Resource Refresh token");
-                keyUsed = CacheKey.createMultiResourceRefreshTokenKey(request, userId);
+                // if not present, check refreshToken for specified resource.
+                // CacheKey will include the resourceId
+                Logger.v(TAG, "Looking for regular refresh token");
+                keyUsed = CacheKey.createCacheKey(request, userId);
                 item = mTokenCacheStore.getItem(keyUsed);
-                multiResource = true;
             }
             
-            // If still cannot find an item, try with family clientid
-            // Disable this feature for ADFS authority
             if (item == null || StringExtensions.IsNullOrBlank(item.getRefreshToken())) {
+                // If still cannot find an item, try with family clientid
+                // Disable this feature for ADFS authority
                 if (!UrlExtensions.isADFSAuthority(StringExtensions.getUrl(mAuthority))) {
                     Logger.v(TAG + methodName, "No refresh token found, trying to find family client id item.");
                     item = findFamilyItemForUser(request.getUserIdentifierType(), userId, request.getClientId());
                     if (item != null) {
                         keyUsed = CacheKey.createCacheKey(item);
-                        multiResource = item.getIsMultiResourceRefreshToken();
                     }
                 } else {
                     Logger.v(TAG + methodName, "No refresh token found, skip the family client id item lookup for ADFS authority.");
@@ -1656,10 +1654,9 @@ public class AuthenticationContext {
 
             if (item != null && !StringExtensions.IsNullOrBlank(item.getRefreshToken())) {
                 String refreshTokenHash = getTokenHash(item.getRefreshToken());
-
                 Logger.v(TAG, "Refresh token is available and id:" + refreshTokenHash
                         + " Key used:" + keyUsed);
-                refreshItem = new RefreshItem(keyUsed, request, item, multiResource);
+                refreshItem = new RefreshItem(keyUsed, request, item);
             }
         }
 
--- a/src/src/com/microsoft/aad/adal/AuthenticationContext.java
+++ b/src/src/com/microsoft/aad/adal/AuthenticationContext.java
@@ -1644,7 +1644,11 @@ public class AuthenticationContext {
             if (item == null || StringExtensions.IsNullOrBlank(item.getRefreshToken())) {
                 if (!UrlExtensions.isADFSAuthority(StringExtensions.getUrl(mAuthority))) {
                     Logger.v(TAG + methodName, "No refresh token found, trying to find family client id item.");
-                    item = findFamilyItemForUser(request.getUserIdentifierType(), userId);
+                    item = findFamilyItemForUser(request.getUserIdentifierType(), userId, request.getClientId());
+                    if (item != null) {
+                        keyUsed = CacheKey.createCacheKey(item);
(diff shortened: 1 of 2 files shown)


weijjia:
Good catch on skipping the token item for same client id. But for the other two change you made, it needs more discussion. This is the behavior across ALL adal platforms for now.

imakhalova:
Explanation about change for multiresource token.
Here is scenario: you sign in and put refresh token for 2 different resources into cache.
Then you hit everyday 1 resource and put its refresh token into cache (including multiresource token).
Then suddenly in a week you need to hit 2nd resource -> you go to cache -> take token for that resource -> hit server and it says that token expired -> you return error that you need to sign in again (launch UI).
But we actually should try multiresource token first, because that one is updated anytime you refresh token for any resource.

src/src/com/microsoft/aad/adal/AuthenticationContext.java
```diff
@@ -1707,6 +1701,11 @@ private TokenCacheItem findFamilyItemForUser(final UserIdentifierType userIdenti
         Logger.v(TAG + methodName, "No family items found in the cache.");
         return null;
     }
+
+    private boolean isRefreshTokenForFamily(TokenCacheItem tokenCacheItem) {
+        return (!StringExtensions.IsNullOrBlank(tokenCacheItem.getFamilyClientId())
```
ntdev-amishra: nit: consider adding a log message here on failure, else from logs you/user will not be able to tell if check in 1686 or 1687 failed

imakhalova: I don't think that we actually want to know for each token why we don't consider it for usage (whether it's not a family token or it's not for our user). It's usually enough to know that we couldn't find anything useful for us

imakhalova:
@RandalliLama , @sdabbiru  , @heidijinxujia ,can you also please take a look at these PR and share your concerns. @weijjia needs your opinion to be able to merge this PR.

src/src/com/microsoft/aad/adal/AuthenticationContext.java
```diff
@@ -1619,32 +1618,35 @@ private RefreshItem getRefreshToken(final AuthenticationRequest request) {
...
             if (StringExtensions.IsNullOrBlank(userId)) {
                 // acquireTokenSilent expects userid field from UserInfo
                 userId = request.getLoginHint();
             }
-            String keyUsed = CacheKey.createCacheKey(request, userId);
+
+            // Target multiResource item in cache first because it's the most recent one.
+            // Cache key will not include resourceId.
+            Logger.v(TAG, "Looking for Multi Resource Refresh token");
+            String keyUsed  = CacheKey.createMultiResourceRefreshTokenKey(request, userId);
```
RandalliLama: @imakhalova When refreshing an AT we should only be trying to use an MRRT if the AT was originally issued along with an MRRT.  You should check mMultiResource before bothering to look up the MRRT cache entry.

RandalliLama: And really, that's the only thing you need to look for.  Use that one and then update both the MRRT entry and the current AT entry to preserve back compat with earlier versions of ADAL.  Every RT retrieved via an MRRT is itself an MRRT.  We should have been updating them all everytime we use any of them.  Alternatively, we should not store MRRTs with ATs and instead just have them refer to a single MRRT entry that gets updated each time it is used.

src/src/com/microsoft/aad/adal/AuthenticationContext.java
```diff
@@ -1619,32 +1618,35 @@ private RefreshItem getRefreshToken(final AuthenticationRequest request) {
...
             
-            // If still cannot find an item, try with family clientid
-            // Disable this feature for ADFS authority
             if (item == null || StringExtensions.IsNullOrBlank(item.getRefreshToken())) {
+                // If still cannot find an item, try with family clientid
+                // Disable this feature for ADFS authority
                 if (!UrlExtensions.isADFSAuthority(StringExtensions.getUrl(mAuthority))) {
                     Logger.v(TAG + methodName, "No refresh token found, trying to find family client id item.");
-                    item = findFamilyItemForUser(request.getUserIdentifierType(), userId);
+                    item = findFamilyItemForUser(request.getUserIdentifierType(), userId, request.getClientId());
```
RandalliLama: @imakhalova  We are going to temporarily disable Family of Client ID support.  The server is going to change its implementation to give us a single FoCI RT that can be used across the entire set of client ids so that we don't have to use buggy heuristics to find the most proofed up FoCI RT.  @weijjia Will send out a PR for that relatively soon.  Unless you want to disable it in this PR.

RandalliLama:
@imakhalova I don't want to stop removing RTs altogether without more consideration.  We have talked about doing something like this in the past, but we need to agree to do this across ADALs, as we want consistent cache logic across ADALs.  

@RPangrle FYI ^

src/src/com/microsoft/aad/adal/AuthenticationContext.java
imakhalova: I'm ok with disabling it, I was just proposing the fix for what we currently have in prod already.

imakhalova:
@RandalliLama , can you explain why we can't turn if off here?
I see that on Android this removal was introduced because logic was:
get RT from cache -> try refreh token -> fails -> remove token -> get  RT from cache -> nothing -> prompt UI
And @weijjia has changed the logic:
get RT from cache -> try refresh token -> fails -> remove token -> prompt UI
So I see the previous logic did require token removal because otherwise it will loop on refresh tokens-failures. But current logic allow this removal very easy and it allows client side to retry multiple times to refresh token and decide when to launch UI and override this refresh tokens.

src/src/com/microsoft/aad/adal/AuthenticationContext.java
RandalliLama: See https://github.com/AzureAD/azure-activedirectory-library-for-android/blob/69198ef1317fd15b473c3300c74ac8453a5a14ee/src/src/com/microsoft/aad/adal/Oauth2.java#L241 for how to determine whether a returned RT is in fact an MRRT.  At this time all RTs returned from AAD are MRRTs.  ADAL doesn't have a way of determining whether it is actually talking to AAD unfortunately, and in the future AADs behavior may change.  Thus, there is a precise way of determining whether the returned RT is in fact an MRRT or not.  Even more unfortunately, the method is of determining whether an RT is in fact an MRRT is absolutely not intuitive.  If the response from the server includes a specific resource then it is in fact an MRRT.  If it does not, then it is not an MRRT.

It appears that the first time the Android library receives an MRRT it treats it as special and places it in its own entry, and then puts a copy into the entry with the AT.  This is broken.  It would have been far better if it had stored exactly one copy of the MRRT.  None of the AT entries should have an RT.  They should instead refer directly to the MRRT.  Each time a new AT is needed the MRRT should be used.  Then the current MRRT should be replaced with the one that was received along with the AT.  Unfortunately, that was made difficult by a poor cache extension model that we were working with in the beginning.  The alternative approach is to not have a separate entry for the MRRT at all.  Instead, each AT is stored with the most current MRRT.  Whenever an MRRT is used, and a new MRRT is returned, then all of the existing copies need to be updated with the new one.  This is how the node.js ADAL works.

NOTE: The current FoCI design is completely contradictory to this and that is probably causing confusion.  However, the FoCI design is being changed so that it will in fact be symmetrical to this design.

The current Android library did not use either of these methods, and that is a major bug that you have discovered.  Whatever we change, we need to retain backward compatibility for anyone that may have decided to extend the cache.  Thus I am suggesting that you switch the code to use the first algorithm I detailed above, namely having a single master copy of the MRRT, but save a copy with the current AT for older consumers that expect the RT to be there as well.

So to go in more depth on your questions:

1) We should know whether an AT was returned along with an MRRT because we should be checking that whenever we get an AT and RT.  If we are not checking that in every case it is a bug.  We should also be saving that fact in every AT entry.  Whenever we need a AT we should first check if the AT was returned along with an MRRT, and if it was then we should be using the master copy of the MRRT.
2) The regular RT you refer to may not actually be a regular RT.  In fact, the only time that it would be is if the RT was obtained directly from ADFS rather than AAD.  The only way to know that for sure is to check and record whether the AT was returned along with an MRRT every single time we get a server response.
3) I think this question is essentially answered by the above.
4) As above, if an AT was ever returned with an MRRT then all subsequent AT refreshes will be returned with an MRRT as well.
5) This is why I am suggesting that the logic should be to always use the single master MRRT but save refreshed MRRT along with the AT anyway.  The code should no longer ever use the RT/MRRT in the AT entry if we know that the AT was returned with an MRRT.

src/src/com/microsoft/aad/adal/AuthenticationContext.java
RandalliLama: Unless I have missed something, the worst case is that we go back to the old behavior.  When the new design is implemented we would then have all of the goodness of FoCI back again.  With the current design there are cases where sign-in behavior may regress, and we want to avoid that.

Out of curiosity, have you already taken a dependency on sharing caches across apps with different client ids?

imakhalova: Nope, we don't share cache, 
It's just sometimes we looping through all our tokens and take one of our refresh tokens.
But I thought server might change the format of familyid (for example, from Boolean to Guid) , because we're not aware of new design. And it might break current implementation and fail on parsing server response.
But it seems like you are disabling only search of this familyId tokens.

RandalliLama:
It is simply to much of a change to make this quickly.  We need more discussion before I would be comfortable doing this.  In iOS we have introduced cache tombstoning.  This is allows us to determine exactly why a refresh token was deleted in the past without keeping potentially invalid RT around.  We then clean those tombstones up when they are too old to be interesting.  I very much want to implement that in Android as well.  If you want, perhaps you could pick up that work.  In the meantime, we need the input of other members of the team before turning of RT deletion, in particular @RPangrle and @kpanwar .  I think that deserves it's own issue or PR.

imakhalova:
Ok, I split up this PR in multiple PRs and we discuss each problem separately...

PR #566, opened 2016-03-30
disable family client id feature

weijjia:
Disable family client id feature for now. The current design doesn't actually the SSO problem across family apps when MFA is involved. Server is going to change its implementation to give us a single FoCI RT that can be used across the entire set of client ids within the same family.

### DIFF
--- a/src/src/com/microsoft/aad/adal/AuthenticationContext.java
+++ b/src/src/com/microsoft/aad/adal/AuthenticationContext.java
@@ -1622,7 +1622,7 @@ public class AuthenticationContext {
             boolean multiResource = false;
             // target refreshToken for this resource first. CacheKey will
             // include the resourceId in the cachekey
-            Logger.v(TAG, "Looking for regular refresh token");
+            Logger.v(TAG + methodName, "Looking for regular refresh token");
             String userId = request.getUserId();
             if (StringExtensions.IsNullOrBlank(userId)) {
                 // acquireTokenSilent expects userid field from UserInfo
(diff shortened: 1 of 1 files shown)


src/src/com/microsoft/aad/adal/AuthenticationContext.java
```diff
@@ -1633,27 +1633,19 @@ private RefreshItem getRefreshToken(final AuthenticationRequest request) {
...
                 // if not present, check multiResource item in cache. Cache key
                 // will not include resourceId in the cache key string.
-                Logger.v(TAG, "Looking for Multi Resource Refresh token");
+                Logger.v(TAG + methodName, "Looking for Multi Resource Refresh token");
                 keyUsed = CacheKey.createMultiResourceRefreshTokenKey(request, userId);
                 item = mTokenCacheStore.getItem(keyUsed);
                 multiResource = true;
             }
             
-            // If still cannot find an item, try with family clientid
```
imakhalova: Usually the best way temporary disable code is to wrap around if (isFeatureEnabled) and set it to false...unless you're going with totally different design

weijjia: I just want to have it temporarily ripped out. When server returns a single FoCI RT, it will make the look up logic way clearer I think.

RPangrle: I'd wait until we have that cache logic settled first before completely ripping out code.

RPangrle:
:shipit:

PR #567, opened 2016-03-30
Provide method that gets refresh token for group/family from cache

imakhalova:
Public method that will allow to get refresh token to exchange with other MS client apps


RandalliLama:
We won't be doing this.  We intentionally restrict access to refresh tokens.  It causes us no end of trouble when people screw this up.  We need to have a detailed conversation about your design for this cache sharing before we proceed.  Feel free to open an issue to discuss.

PR #572, opened 2016-04-05
update readme with requirements for talking to broker; add sample for authentication via broker

weijjia:
1. Add sample for authentication via broker
2. update readme on how to talk to broker and encryption key generation


README.md
```diff
@@ -233,46 +233,47 @@ If you're implementing your authentication logic in a Fragment, you'll need to w
...
   
-  The Callback will have an object of AuthenticationResult which has accesstoken, date expired, and idtoken info. 
+	The Callback will have an object of AuthenticationResult which has accesstoken, date expired, and idtoken info. 
+	
+	Optional:  **acquireTokenSilent**
 
-Optional:  **acquireTokenSilent**
-
-You can call **acquireTokenSilent** to handle caching, and token refresh. It provides sync version as well. It accepts userid as paremeter.
+	You can call **acquireTokenSilent** to handle caching, and token refresh. It provides sync version as well. It accepts userid as paremeter.
```
sdabbiru: sync -> async?
Paremeter ->parameter

README.md
```diff
@@ -233,46 +233,47 @@ If you're implementing your authentication logic in a Fragment, you'll need to w
...
+	* Resource is required and is the resource you are trying to access.
+	* Clientid is required and comes from the AzureAD Portal.
+	* You can setup redirectUri as your packagename. It is not required to be provided for the acquireToken call.
+	* PromptBehavior helps to ask for credentials to skip cache and cookie. 
+	* Callback will be called after authorization code is exchanged for a token. 
   
-  The Callback will have an object of AuthenticationResult which has accesstoken, date expired, and idtoken info. 
+	The Callback will have an object of AuthenticationResult which has accesstoken, date expired, and idtoken info. 
+	
+	Optional:  **acquireTokenSilent**
```
sdabbiru: I do not know what is "optional" about acquireTokenSilent... Usually "Optional" phrase is used for describing optional parameters. In this case, we are talking about a different method. What we should be explaining is about why one should call or not call acquireTokenSilent.

README.md
```diff
@@ -233,46 +233,47 @@ If you're implementing your authentication logic in a Fragment, you'll need to w
     }
     ```
 
-Explanation of the parameters:
+	Explanation of the parameters:
```
kpanwar: example of the parameters maybe helpful.

README.md
kpanwar: I agree. acquireTokenSilent MUST be used by the developers. API usage pattern calls out for calling silent first and then falling back to UI.

README.md
```diff
@@ -233,46 +233,47 @@ If you're implementing your authentication logic in a Fragment, you'll need to w
...
 
-Using this walkthrough, you should have what you need to successfully integrate with Azure Active Directory. For more examples of this working, viist the AzureADSamples/ repository on GitHub.
+	Microsoft Intune's Company portal App and Azure Authenticator App will provide the broker component. 
+	In order to acquire token via broker, the following requirements have to meet(Please check samples\userappwithbroker for authentication via broker):
+	* Starting version 1.1.14, developer has to explicitly specify set to use broker via:
+		```
+		AuthenticationSettings.Instance.setUseBroker(true);
+		```
+	* Developer needs to register special redirectUri for broker usage. RedirectUri is in the format of msauth://packagename/Base64UrlencodedSignature. You can get your redirecturi for your app using the script "brokerRedirectPrint.ps1" or use API call mContext.getBrokerRedirectUri. Signature is related to your signing certificates.
+	* If target version is lower than 23, calling app has to have the following permissions declared in manifest(http://developer.android.com/reference/android/accounts/AccountManager.html):
```
kpanwar: We should really consider putting a bash script as well. I would assume most of the android developers are on some sort of unix platform.

README.md
```diff
@@ -233,46 +233,47 @@ If you're implementing your authentication logic in a Fragment, you'll need to w
...
+		* GET_ACCOUNTS
+		* USE_CREDENTIALS
+		* MANAGE_ACOUNTS
+	* If target version is 23, USE_CREDENTIALS and MANAGE_ACCOUNTS are already deprecated. But GET_ACCOUNTS is under protection level "dangerous", calling app is responsible for requesting the run-time permisson.
```
kpanwar: maybe provide a link to android doc that outlines the call and callback handling for requesting permission?

README.md
```diff
@@ -233,46 +233,47 @@ If you're implementing your authentication logic in a Fragment, you'll need to w
...
+		```
+		AuthenticationSettings.Instance.setUseBroker(true);
+		```
+	* Developer needs to register special redirectUri for broker usage. RedirectUri is in the format of msauth://packagename/Base64UrlencodedSignature. You can get your redirecturi for your app using the script "brokerRedirectPrint.ps1" or use API call mContext.getBrokerRedirectUri. Signature is related to your signing certificates.
+	* If target version is lower than 23, calling app has to have the following permissions declared in manifest(http://developer.android.com/reference/android/accounts/AccountManager.html):
+		* GET_ACCOUNTS
+		* USE_CREDENTIALS
+		* MANAGE_ACOUNTS
+	* If target version is 23, USE_CREDENTIALS and MANAGE_ACCOUNTS are already deprecated. But GET_ACCOUNTS is under protection level "dangerous", calling app is responsible for requesting the run-time permisson. 
+	* If spoke to the broker version on V1 protocol, broker will only support single user. To talk to this version of broker, there has to be an account existed.
```
kpanwar: What is v1 protocol? How will a 3rd party developer know the difference? We should explain the experience or possible error codes the developer can see when using v1 vs v2 broker.

README.md
```diff
@@ -393,7 +394,17 @@ acquireToken method without activity supports dialog prompt.
 
 ### Encryption
 
-ADAL encrypts the tokens and store in SharedPreferences by default. You can look at the StorageHelper class to see the details. Android introduced AndroidKeyStore for 4.3(API18) secure storage of private keys. ADAL uses that for API18 and above. If you want to use ADAL for lower SDK versions, you need to provide secret key at AuthenticationSettings.INSTANCE.setSecretKey
+ADAL encrypts the tokens and store in SharedPreferences by default. You can look at the StorageHelper class to see the details. Android introduced AndroidKeyStore for 4.3(API18) secure storage of private keys. ADAL uses that for API18 and above. If you want to use ADAL for lower SDK versions, you need to **provide secret key at AuthenticationSettings.INSTANCE.setSecretKey**
```
kpanwar: Consider rephrasing "Android introduced AndroidKeyStore for 4.3(API18) secure storage of private keys. ADAL uses that for API18 and above." --> "ADAL uses AndroidKeyStore in 4.3 (API18) and above for secure storage of private keys"

README.md
```diff
@@ -393,7 +394,17 @@ acquireToken method without activity supports dialog prompt.
...
+
+```Java
+	SecretKeyFactory keyFactory = SecretKeyFactory
+		.getInstance("PBEWithSHA256And256BitAES-CBC-BC");
```
kpanwar: does ADAL require the algorithm to be "PBEWithSHA256And256BitAES-CBC-BC"? If not, then lets not put something that may be perceived as a requirement. Someone may want to use SHA512 or SHA384.

README.md
```diff
@@ -393,7 +394,17 @@ acquireToken method without activity supports dialog prompt.
...
+    SecretKey generatedSecretKe = keyFactory.generateSecret(new PBEKeySpec("your_password".toCharArray(),
```
kpanwar: We should provide guidelines for key generation, providing code like this would result in people copy pasting the code as it and not changing "your_password" and other params. Also, "your_password" seems confusing.

README.md
weijjia: I'll add a link to the samples for examples for those parameter.

README.md
weijjia: I'll file a separate issue on that.

README.md
weijjia: Then we need to know the specific broker app version. With new broker, device registration could be triggered during CA flow. Ok, i'll add the diff back when the new broker is actually shipped.

kpanwar:
:shipit:

Issue #449, opened 2015-11-03
Improve build dependency setup.


RandalliLama:
There are several getLibs scripts lying around that a developer needs to find and run in order to get a working build.  These should consolidated into a single script that sets everything up, or they should be removed altogether and replace with something else.

weijjia:
Moved to gradle, and getLibs scripts are already removed.

Issue #448, opened 2015-11-03
Need to add comprehensive build instructions


RandalliLama:
There are apparently no instructions on how to install dependencies.  We should document that and any other non-obvious steps. If running tests is not obvious for instance, then that should be doc'ed.

weijjia:
Closing this since we switched to android studio.

PR #573, opened 2016-04-06
Switch the ADAL android to Android Studio

heidijinxujia:
The goal for this PR is to configure the gradle build for the android studio. Any code change should be in their own sub-branch from branch android-studio.

• this branch is a sub-branch from dev
• configure Findbugs, PMD, Android Lint and Checkstyle into the gradle build
• create the jar and aar files
• unify the names of the output
• remove Travis


adal/proguard-rules.pro
```diff
@@ -1,11 +1,8 @@
-# To enable ProGuard in your project, edit project.properties
-# to define the proguard.config property as described in that file.
-#
 # Add project specific ProGuard rules here.
```
weijjia: Do we have any proguard rules here? I wonder if we don't have rules, do we actually need it?

weijjia:
If we're going to merge the android studio config into dev now, just want to do sanity check and please make sure we have the correct setting for publish to maven central, then the next release for 1.1.17 will be pushed out with gradle build.

weijjia:
Also, IMO if pushing the change to dev, we should have travis configured. This just to make sure we won't have any PRs break our build.

adal/build.gradle
```diff
@@ -0,0 +1,151 @@
...
+
+group = 'com.microsoft.aad'
+version = '1.1.16'
+
+buildscript {
+    repositories {
+        jcenter()
+    }
+    dependencies {
+            classpath 'com.github.dcendents:android-maven-gradle-plugin:1.3'
```
weijjia: nit: use space instead of tab

adal/build.gradle
```diff
@@ -0,0 +1,151 @@
...
+
+    defaultConfig {
+        minSdkVersion 18
+        targetSdkVersion 21
+        versionCode 1
+        versionName "1.1.16"
+        project.archivesBaseName = "adal"
+        project.version = android.defaultConfig.versionName
+    }
+    buildTypes {
```
weijjia: nit: new line

adal/build.gradle
```diff
@@ -0,0 +1,151 @@
...
+        release {
+            minifyEnabled false
+            proguardFiles getDefaultProguardFile('proguard-android.txt'), 'proguard-rules.pro'
+        }
+    }
+    sourceSets {
```
weijjia: nit: new line

adal/build.gradle
```diff
@@ -0,0 +1,151 @@
...
+    libraryVariants.all { variant ->
+        variant.outputs.each { output ->
+            def outputFile = output.outputFile
+            if (outputFile != null && outputFile.name.endsWith('.aar')) {
+                def fileName = "${archivesBaseName}-${version}.aar"
+                output.outputFile = new File(outputFile.parent, fileName)
+            }
+        }
+    }
+
```
weijjia: nit: extra line

adal/build.gradle
```diff
@@ -0,0 +1,151 @@
...
+}
+
+dependencies {
+    compile fileTree(dir: 'libs', include: ['*.jar'])
```
weijjia: why do we need to compile the jars in lib folder?

adal/build.gradle
```diff
@@ -0,0 +1,151 @@
...
+    androidTestCompile 'org.mockito:mockito-core:1.10.19'
+    androidTestCompile 'com.google.dexmaker:dexmaker:1.2'
+    androidTestCompile 'com.google.dexmaker:dexmaker-mockito:1.2'
+    compile 'com.android.support:appcompat-v7:21.0.3'
+    compile 'com.google.code.gson:gson:2.2.4'
+    testCompile 'junit:junit:4.12'
```
weijjia: nit: but I prefer we separate the dependency like this:
// dependencies required by tests
list of the test dependencies

// dependencies required for compile
list of the dependencies required for compile

adal/build.gradle
```diff
@@ -0,0 +1,151 @@
...
+    destinationDir = reporting.file("$project.buildDir/outputs/jar/")
+}
+
+task javadoc(type: Javadoc) {
+    source = android.sourceSets.main.java.srcDirs
+    classpath = configurations.compile
+    destinationDir = reporting.file("$project.buildDir/outputs/jar/")
+}
+
+task javadocJar(type: Jar, dependsOn: javadoc) {
```
weijjia: what will be the output name for sourcejar and javadoc jar?

adal/proguard-rules.pro
heidijinxujia: "The ProGuard tool shrinks, optimizes, and obfuscates your code by removing unused code and renaming classes, fields, and methods with semantically obscure names." I don't think we have the  need to use it now. I remove it.

heidijinxujia:
OK. So there are two remaining things need to be done here. 
1. Deploying to maven central with Gradle
2. Travis configuration. 
@weijjia @RandalliLama

adal/proguard-rules.pro
weijjia: I know what proguards does, It's fine and nice to use it, most apps turn on proguard for release build. But we don't have any rules in the proguard file.

sample/src/test/java/com/microsoft/aad/adal/sample/ExampleUnitTest.java
```diff
@@ -0,0 +1,15 @@
+package com.microsoft.aad.adal.sample;
+
+import org.junit.Test;
+
+import static org.junit.Assert.*;
+
+/**
+ * To work on unit tests, switch the Test Artifact in the Build Variants view.
+ */
+public class ExampleUnitTest {
```
RandalliLama: Does this file need to be in the repo or did some tool generate it.  It doesn't seem like we need a sample unit test.

adal/build.gradle
RandalliLama: @weijjia @heidijinxujia  That's not a nit.  That's a rule.

adal/build.gradle
RandalliLama: @weijjia I'm not arguing with you, you are probably right, but could you suggest what you think would be a better place?

adal/build.gradle
RandalliLama: agree

RandalliLama:
@heidijinxujia Don't change the flag yet.  We can change that back as soon as we fix all the issues.  Which, of course, we will do before the next release.  I don't think we need to do it all in one PR though.  So don't bother with your #3 yet.

adal/build.gradle
weijjia: Well, I simply don't know why we need to lib folder. Do we have any particular jar to compile that cannot be grabbed from those central repos? The reason we have today AFAIK is for eclipse setup(In eclipse we set up a android project but not maven project, you have to put those jars somewhere, so eclipse will know their existence). But now we move to android studio, gradle build is integrated. Correct me if i'm wrong.

adal/build.gradle
weijjia: agree. :)

sample/build.gradle
```diff
@@ -0,0 +1,33 @@
+apply plugin: 'com.android.application'
+
+android {
+    compileSdkVersion 21
+    buildToolsVersion "21.1.1"
```
weijjia: why only sample app use build tool version 21.1.1 while others use 21.1.2?

sample/build.gradle
```diff
@@ -0,0 +1,33 @@
...
+    }
+
+    lintOptions {
+        abortOnError false
+    }
+}
+
+dependencies {
+    compile fileTree(dir: 'libs', include: ['*.jar'])
+    testCompile 'junit:junit:4.12'
```
weijjia: Do we have tests for sample app? If so, why do we need tests for sample app?

adal/src/androidTest/java/com/microsoft/aad/adal/StorageHelperTests.java
```diff
@@ -276,7 +279,7 @@ public void testKeyPair() throws
 
         assertTrue("Keystore has the alias", keyStore.containsAlias("AdalKey"));
     }
-
+*/
```
addev-ashish: is there a tab missing here? 
Also Why don't we just delete this function?

adal/build.gradle
```diff
@@ -0,0 +1,167 @@
...
+        }
+    }
+    sourceSets {
+        main {
+            manifest.srcFile 'src/main/AndroidManifest.xml'
+        }
+    }
+
+    lintOptions {
+        abortOnError false
```
agamemnum: Is code fails Lint? Why do you disable this?

adal/build.gradle
```diff
@@ -0,0 +1,167 @@
...
+        }
+    }
+
+}
+
+dependencies {
+    androidTestCompile 'org.mockito:mockito-core:1.10.19'
+    androidTestCompile 'com.google.dexmaker:dexmaker:1.2'
+    androidTestCompile 'com.google.dexmaker:dexmaker-mockito:1.2'
+    compile 'com.android.support:appcompat-v7:21.0.3'
```
agamemnum: This appcompat is really old. Consider using at least 22

adal/build.gradle
```diff
@@ -0,0 +1,167 @@
...
+            classpath 'com.github.dcendents:android-maven-gradle-plugin:1.3'
+    }
+}
+
+android {
+    compileSdkVersion 21
+    buildToolsVersion "21.1.2"
+
+    defaultConfig {
+        minSdkVersion 18
```
agamemnum: it should be 14

adal/build.gradle
heidijinxujia: Hi @agamemnum , thank you for your review. The deployment of the gradle build is still under construction. The flag is set to be false temporarily, and I am fixing the lint errors these days. All the flags of code analysis tools will be set back to be true after finishing fixing all errors reported (ASAP). @RandalliLama

adal/build.gradle
heidijinxujia: OK. Thanks.

addev-ashish:
#449 #448

addev-ashish:
>                 return true;

This code should not change?

---
Refers to: adal/src/main/java/com/microsoft/aad/adal/AuthenticationActivity.java:625 in 0fb7c6f. [](commit_id = 0fb7c6faf6bcd68355db915a6748ef86387027e8, deletion_comment = False)

adal/src/main/java/com/microsoft/aad/adal/AuthenticationActivity.java
```diff
@@ -959,13 +948,13 @@ private void setAccount(final TokenTaskResult result)
...
 
-            TokenCacheItem item = TokenCacheItem.createRegularTokenCacheItem(mRequest.getAuthority(), mRequest.getResource(), 
+            TokenCacheItem item = TokenCacheItem.createRegularTokenCacheItem(mRequest.getAuthority(), mRequest.getResource(),
                     mRequest.getClientId(), result.taskResult);
             String json = gson.toJson(item);
             String encrypted = mStorageHelper.encrypt(json);
 
             // Single user and cache is stored per account
-            String key = CacheKey.createCacheKeyForRTEntry(mAuthRequest.getAuthority(), mAuthRequest.getResource(), 
+            String key = CacheKey.createCacheKeyForRTEntry(mAuthRequest.getAuthority(), mAuthRequest.getResource(),
```
addev-ashish: looks like the FOCI change

PR #581, opened 2016-04-09
Family client id feature is disabled for now, tests for token acquisition with family item should be disabled

weijjia:
Now the family client id feature is disabled, the logic for looking into cache with "FoCI" flag won't be hit, corresponding tests should also be suppressed. 

Still keep the test for 'FoCI' flag is correctly stored into cache. Server is still returning the "FoCI" flag for first party apps. Those tests don't needs to be suppressed for the sake for making sure the "FoCI" flag is correctly stored for token cache item.


imakhalova:
merge it

kpanwar:
Would still prefer the non-executing code to be removed. :shipit:

PR #583, opened 2016-04-11
If AT was obtained by MRRT then don't check regular RT to refresh token

imakhalova:
Here is scenario: you sign in and put refresh token for 2 different resources into cache.
 Then you hit everyday 1 resource and put its refresh token into cache (including multiresource token).
 Then suddenly in a week you need to hit 2nd resource -> you go to cache -> take token for that resource -> hit server and it says that token expired -> you return error that you need to sign in again (launch UI).
 But we actually should try multiresource token if access token was obtained by MRRT or we don't have any access token for particular resource.


src/src/com/microsoft/aad/adal/AuthenticationContext.java
```diff
@@ -1430,7 +1430,8 @@ private AuthenticationResult localFlow(CallbackHandler callbackHandle,
 
         // Trying to find refresh token first, if existed, will try to use the refresh token. 
         Logger.v(TAG, "Checking refresh tokens");
-        RefreshItem refreshItem = getRefreshToken(request);
+        boolean useMultiResourceRefreshToken = cachedItem == null ? true : cachedItem.getIsMultiResourceRefreshToken();
```
weijjia: If you're not going to change the value, make it final whenever possible

imakhalova: sure, let's make it final

agamemnum: nit: boolean useMultiResourceRefreshToken = cachedItem == null || cachedItem.getIsMultiResourceRefreshToken();

weijjia:
left minor comment. I'm fine with the cache lookup, :shipit: if @RandalliLama is fine with it. :)

RandalliLama:
I'm looking at this.  Will need to sign-off tomorrow morning.

RandalliLama:
It seems factored weird in that I would expect getRefreshToken handle this since it is that methods job to find a refresh token regardless of what kind it is.  However, I won't block on that for now since it doesn't make the code significantly worse than it already is.

:shipit:

imakhalova:
It's actually good feedback, and let's make it better. 
What if getRefreshToken will check RT first and if it multiresource take MRRT from cache?

RandalliLama:
@imakhalova Exactly right.  I think that can wait for a better refactor however.  I we need a cache layer that handles all of this transparently to the rest of the code.  The cache logic should be in it's own class rather than spread across the various classes.

PR #584, opened 2016-04-12
do not remove token in case of Server Error without body in response

imakhalova:
I've got sign out with some log that clearly says that I've got my token wiped because of some server error.
There is some unit test (testRefreshTokenWebRequestHasError) that shows that it's not expected behavior (we should not wipe token when we've got server error without any body in response). But this UT had an error, it checked whether it has token in cache before the aquireToken call was finished and didn't catch that token was actually wiped (And this UT actually should fail)
Here is my proposal how to fix the bug and the test.

And here is the log:
SERVER_ERROR:2016-03-26 07:27:35-74175305-f579-4212-8f49-d552fd095f8d-IOException:https://login.windows.net/common/oauth2/token ver:1.1.16 
03-26 00:27:35.609 9276-9484/? V/HttpWebRequest: 2016-03-26 07:27:35-74175305-f579-4212-8f49-d552fd095f8d-Status code:504 ver:1.1.16
03-26 00:27:35.610 9276-9484/? V/HttpWebRequest: 2016-03-26 07:27:35-74175305-f579-4212-8f49-d552fd095f8d-Response is received ver:1.1.16
03-26 00:27:35.611 9276-9484/? V/Oauth: 2016-03-26 07:27:35-74175305-f579-4212-8f49-d552fd095f8d-Token request does not have exception ver:1.1.16
03-26 00:27:35.612 9276-9484/? V/Oauth: 2016-03-26 07:27:35-74175305-f579-4212-8f49-d552fd095f8d-Server error message: ver:1.1.16
03-26 00:27:35.613 9276-9484/? V/AuthenticationContext: 2016-03-26 07:27:35-74175305-f579-4212-8f49-d552fd095f8d-Refresh token is not returned or empty ver:1.1.16
03-26 00:27:35.613 9276-9484/? W/AuthenticationContext: AUTH_FAILED_NO_TOKEN:2016-03-26 07:27:35-74175305-f579-4212-8f49-d552fd095f8d-Refresh token did not return accesstoken. ver:1.1.16 Request authority:https://login.windows.net/common resource:https://officeapps.live.com clientid:b26aadf8-566f-4478-926f-589f601d9c74 ErrorCode:504 ErrorDescription:
03-26 00:27:35.614 9276-9484/? V/AuthenticationContext: 2016-03-26 07:27:35-74175305-f579-4212-8f49-d552fd095f8d-Remove refresh item from cache:https://login.windows.net/common$https://officeapps.live.com$b26aadf8-566f-4478-926f-589f601d9c74$n$26d2c960-2104-4fc4-84c8-b64b442841c6 ver:1.1.16
03-26 00:27:35.626 9276-9484/? V/AuthenticationContext: 2016-03-26 07:27:35-74175305-f579-4212-8f49-d552fd095f8d-Refresh token is not available or refresh

### DIFF
--- a/src/src/com/microsoft/aad/adal/AuthenticationContext.java
+++ b/src/src/com/microsoft/aad/adal/AuthenticationContext.java
@@ -1867,7 +1867,7 @@ public class AuthenticationContext {
         if (useCache) {
             if (result == null || StringExtensions.IsNullOrBlank(result.getAccessToken())) {
                 String errLogInfo = result == null ? "" : result.getErrorLogInfo();
-                Logger.w(TAG, "Refresh token did not return accesstoken.", request.getLogInfo()
+                Logger.e(TAG, "Refresh token did not return accesstoken.", request.getLogInfo()
                         + errLogInfo, ADALError.AUTH_FAILED_NO_TOKEN);
 
                 // remove item from cache to avoid same usage of
--- a/src/src/com/microsoft/aad/adal/Oauth2.java
+++ b/src/src/com/microsoft/aad/adal/Oauth2.java
@@ -511,7 +511,9 @@ class Oauth2 {
                 }
             }
 
-            if (response.getBody() != null) {
+            byte[] bodyMessage = response.getBody();
+            boolean isBodyEmpty = bodyMessage == null || bodyMessage.length == 0;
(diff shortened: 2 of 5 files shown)


weijjia:
:shipit:

Issue #585, opened 2016-04-13
NON_HTTPS_REDIRECT for about:blank


heidijinxujia:
We probably have the same issue with iOS's [https://github.com/AzureAD/azure-activedirectory-library-for-objc/issues/585](url) The "about:blank" should also be considered in the redirect URL loading checking.

PR #587, opened 2016-04-15
fix for whitelisting "about:blank"

heidijinxujia:
The "about:blank" should also be considered in the redirect URL loading checking.
see details #585

### DIFF
--- a/src/src/com/microsoft/aad/adal/AuthenticationActivity.java
+++ b/src/src/com/microsoft/aad/adal/AuthenticationActivity.java
@@ -615,10 +615,10 @@ public class AuthenticationActivity extends Activity {
         }
         
         public boolean processInvalidUrl(final WebView view, String url) {
-        	final String methodName = ":processInvalidUrl";
+            final String methodName = ":processInvalidUrl";
             if (isBrokerRequest(getIntent())
                     && url.startsWith(AuthenticationConstants.Broker.REDIRECT_PREFIX)) {
-            	Logger.e(TAG + methodName, String.format(
+                Logger.e(TAG + methodName, String.format(
                         "The RedirectUri is not as expected. Received %s and expected %s", url,
                         mRedirectUrl), "", ADALError.DEVELOPER_REDIRECTURI_INVALID);
                 returnError(ADALError.DEVELOPER_REDIRECTURI_INVALID, String.format(
@@ -629,13 +629,13 @@ public class AuthenticationActivity extends Activity {
             } else if (url.toLowerCase(Locale.US).equals("about:blank")) {
                 Logger.v(TAG, "It is an blank page request");
                 return true;
-            } else if(!url.toLowerCase(Locale.US).startsWith(AuthenticationConstants.Broker.REDIRECT_SSL_PREFIX)){
-            	Logger.e(TAG + methodName, "The webview was redirected to an unsafe URL.", "", ADALError.WEBVIEW_REDIRECTURL_NOT_SSL_PROTECTED);
-            	returnError(ADALError.WEBVIEW_REDIRECTURL_NOT_SSL_PROTECTED, "The webview was redirected to an unsafe URL.");
-            	view.stopLoading();
-            	return true;
+            } else if (!url.toLowerCase(Locale.US).startsWith(AuthenticationConstants.Broker.REDIRECT_SSL_PREFIX)) {
+                Logger.e(TAG + methodName, "The webview was redirected to an unsafe URL.", "", ADALError.WEBVIEW_REDIRECTURL_NOT_SSL_PROTECTED);
+                returnError(ADALError.WEBVIEW_REDIRECTURL_NOT_SSL_PROTECTED, "The webview was redirected to an unsafe URL.");
+                view.stopLoading();
+                return true;
             } else {
-            	return false;
+                return false;
             }
         }
         
@@ -1032,8 +1032,8 @@ public class AuthenticationActivity extends Activity {
                     }
                     
                     if (result.taskResult.getTenantId() != null) {
-                    	intent.putExtra(AuthenticationConstants.Broker.ACCOUNT_USERINFO_TENANTID, 
-                    			result.taskResult.getTenantId());
+                        intent.putExtra(AuthenticationConstants.Broker.ACCOUNT_USERINFO_TENANTID, 
+                                result.taskResult.getTenantId());
                     }
 
                     UserInfo userinfo = result.taskResult.getUserInfo();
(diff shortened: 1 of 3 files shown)


src/src/com/microsoft/aad/adal/BasicWebViewClient.java
```diff
@@ -269,6 +269,9 @@ public void run() {
             openLinkInBrowser(parameters.get(ApplicationReceiver.INSTALL_URL_KEY));
             view.stopLoading();
             return true;
+        } else if (url.startsWith("about:blank")) {
```
weijjia: IMO, this should be moved to the place where you check for SSL

src/src/com/microsoft/aad/adal/AuthenticationActivity.java
```diff
@@ -627,18 +626,17 @@ public boolean processInvalidUrl(final WebView view, String url) {
...
                 return true;
-            }
-            
-            //check if the redirect URL is under SSL protected
-            if(!url.toLowerCase(Locale.US).startsWith(AuthenticationConstants.Broker.REDIRECT_SSL_PREFIX))
-            {
+            } else if (url.startsWith("about:blank")) {
+                Logger.v(TAG, "It is an blank page request");
+                return true;
+            } else if(!url.toLowerCase(Locale.US).startsWith(AuthenticationConstants.Broker.REDIRECT_SSL_PREFIX)){
```
weijjia: I would rather have a new method i.e.
boolean isValidRedicectUrl(final String url) {
    return url.toLowerCase(Locale.US).startsWith(AuthenticationConstants.Broker.REDIRECT_SSL_PREFIX) || url.equalsIgnoreCase("about:blank");
}

src/src/com/microsoft/aad/adal/AuthenticationActivity.java
```diff
@@ -627,18 +626,17 @@ public boolean processInvalidUrl(final WebView view, String url) {
                         mRedirectUrl));
                 view.stopLoading();
                 return true;
-            }
-            
-            //check if the redirect URL is under SSL protected
-            if(!url.toLowerCase(Locale.US).startsWith(AuthenticationConstants.Broker.REDIRECT_SSL_PREFIX))
-            {
+            } else if (url.startsWith("about:blank")) {
```
weijjia: shouldn't we compare if it's equals to?

src/src/com/microsoft/aad/adal/AuthenticationActivity.java
heidijinxujia: @weijjia  In this function, you treated SSL-protected page and "about:blank" as the same, where as a result, not only the ssl-protected page will be opened in the webview, but also the "about:blank" empty page. I am wondering in which situation we need to open an blank webpage. Should we override the "about:blank" url loading, or not? Thanks.

src/src/com/microsoft/aad/adal/AuthenticationActivity.java
heidijinxujia: Fixed in the latest commit

src/src/com/microsoft/aad/adal/AuthenticationActivity.java
weijjia: IMO, you're checking if the redirect is invalid or not, and with your fix, you treat about:blank as a valid one, then both ssl protected page and about:blank are valid.

heidijinxujia: The "about:blank" is treated as invalid redirect URL here. The function processInvalidUrl returns true if the URL is invalid. 
....
else if (url.startsWith("about:blank")) {  
                Logger.v(TAG, "It is an blank page request");  
               return true;  
} ....

src/src/com/microsoft/aad/adal/AuthenticationActivity.java
```diff
@@ -627,18 +626,17 @@ public boolean processInvalidUrl(final WebView view, String url) {
...
+            } else if (url.toLowerCase(Locale.US).equals("about:blank")) {
+                Logger.v(TAG, "It is an blank page request");
+                return true;
+            } else if(!url.toLowerCase(Locale.US).startsWith(AuthenticationConstants.Broker.REDIRECT_SSL_PREFIX)){
             	Logger.e(TAG + methodName, "The webview was redirected to an unsafe URL.", "", ADALError.WEBVIEW_REDIRECTURL_NOT_SSL_PROTECTED);
             	returnError(ADALError.WEBVIEW_REDIRECTURL_NOT_SSL_PROTECTED, "The webview was redirected to an unsafe URL.");
             	view.stopLoading();
             	return true;
+            } else {
+            	return false;
```
RandalliLama: This needs to be indented.

src/src/com/microsoft/aad/adal/AuthenticationActivity.java
```diff
@@ -627,18 +626,17 @@ public boolean processInvalidUrl(final WebView view, String url) {
...
                 return true;
-            }
-            
-            //check if the redirect URL is under SSL protected
-            if(!url.toLowerCase(Locale.US).startsWith(AuthenticationConstants.Broker.REDIRECT_SSL_PREFIX))
-            {
+            } else if (url.toLowerCase(Locale.US).equals("about:blank")) {
+                Logger.v(TAG, "It is an blank page request");
+                return true;
+            } else if(!url.toLowerCase(Locale.US).startsWith(AuthenticationConstants.Broker.REDIRECT_SSL_PREFIX)){
```
addev-ashish: nit: 2 spaces missing one after if and once before {
} else if (textTextText) {

src/src/com/microsoft/aad/adal/AuthenticationActivity.java
heidijinxujia: @weijjia

weijjia: I'm fine with your current change. Make sure you apply all the comments.

tests/Functional/src/com/microsoft/aad/adal/test/AuthenticationActivityUnitTest.java
```diff
@@ -279,6 +279,25 @@ public void testWebview_sslprotectedredirectURL() throws IllegalArgumentExceptio
...
+        startActivity(intentToStartActivity, null, null);
+        activity = getActivity();
+        /*
+         * case 1: url = "about:blank"
+         */
+        String url = "about:blank";
+        WebViewClient client = getCustomWebViewClient();
+        WebView mockview = new WebView(getActivity().getApplicationContext());
+        ReflectionUtils.setFieldValue(activity, "mSpinner", null);
+        //shouldOverrideUrlLoading should prevent non https redirects in the web view
```
weijjia: remove code that is commented out.

weijjia:
sign off after you apply all the comments.

heidijinxujia:
Applied all the comments and plan to sign off.

weijjia:
:shipit:

PR #588, opened 2016-04-15
Refresh token removal

weijjia:
1. Only remove token when receiving invalid grant
2. update parsing token response: if receiving 200, 400, 401, also try to read json response, if failed, creating AuthenticationResult with statusCode as errorCode, and raw response as error description

### DIFF
--- a/src/src/com/microsoft/aad/adal/ADALError.java
+++ b/src/src/com/microsoft/aad/adal/ADALError.java
@@ -515,7 +515,12 @@ public enum ADALError {
     /**
      * Resource authentication challenge failure
      */
-    RESOURCE_AUTHENTICATION_CHALLENGE_FAILURE("Resource authentication challenge failure");
+    RESOURCE_AUTHENTICATION_CHALLENGE_FAILURE("Resource authentication challenge failure"),
+    
+    /**
+     * Represents a general OAUTH Error
+     */
+    OAUTH_ERROR("Oauth Error");
 
     private String mDescription;
 
--- a/src/src/com/microsoft/aad/adal/ADALError.java
+++ b/src/src/com/microsoft/aad/adal/ADALError.java
@@ -520,7 +520,7 @@ public enum ADALError {
     /**
      * Represents a general OAUTH Error
      */
-    OAUTH_ERROR("Oauth Error");
+    OAUTH2_ERROR("Oauth2 Error");
 
     private String mDescription;
 
--- a/src/src/com/microsoft/aad/adal/AuthenticationConstants.java
+++ b/src/src/com/microsoft/aad/adal/AuthenticationConstants.java
@@ -149,6 +149,13 @@ public class AuthenticationConstants {
         
         static final String ADAL_CLIENT_FAMILY_ID = "foci";
     }
+    
+    public static final class OAuth2ErrorCode {
+        /**
+         * Oauth2 error code invalid_grant. 
+         */
+        static final String INVALID_GRANT = "invalid_grant";
+    }
 
     public static final class AAD {
 
--- a/src/src/com/microsoft/aad/adal/AuthenticationContext.java
+++ b/src/src/com/microsoft/aad/adal/AuthenticationContext.java
@@ -1875,7 +1875,6 @@ public class AuthenticationContext {
                         request.getLogInfo() + errorLogInfo, ADALError.AUTH_FAILED_NO_TOKEN);
                 
                 // check error code, only remove token from cache if receive invalid_grant from server
-                Logger.v(TAG + "tessting", "error_code:" + result.getErrorCode());
                 if (result.getErrorCode().equalsIgnoreCase(AuthenticationConstants.OAuth2ErrorCode.INVALID_GRANT)) {
                     Logger.v(TAG, "Removing token cache for invalid_grant error returned from server.");
                     removeItemFromCache(refreshItem);
--- a/src/src/com/microsoft/aad/adal/AuthenticationContext.java
+++ b/src/src/com/microsoft/aad/adal/AuthenticationContext.java
@@ -1472,7 +1472,7 @@ public class AuthenticationContext {
                         ADALError.AUTH_REFRESH_FAILED_PROMPT_NOT_ALLOWED);
                 if (authResult != null && authResult.getErrorCode() != null) {
                     callbackHandle.onError(new AuthenticationException(ADALError.AUTH_REFRESH_FAILED_PROMPT_NOT_ALLOWED, 
-                            request.getLogInfo() + " " + errorInfo, new AuthenticationServerProtocolException(ADALError.OAUTH_ERROR,
+                            request.getLogInfo() + " " + errorInfo, new AuthenticationServerProtocolException(ADALError.OAUTH2_ERROR,
                                     authResult.getErrorCode(), authResult.getErrorDescription())));
                 } else {
                     callbackHandle.onError(new AuthenticationException(
--- a/src/src/com/microsoft/aad/adal/AuthenticationContext.java
+++ b/src/src/com/microsoft/aad/adal/AuthenticationContext.java
@@ -1865,14 +1865,22 @@ public class AuthenticationContext {
         // useCache will be false when calling from acquireTokenUsingRefreshToken, and if false, need to return
         // the error through callback. If true, just return the result back to localflow. 
         if (useCache) {
-            if (result == null || StringExtensions.IsNullOrBlank(result.getAccessToken())) {
-                String errLogInfo = result == null ? "" : result.getErrorLogInfo();
-                Logger.w(TAG, "Refresh token did not return accesstoken.", request.getLogInfo()
-                        + errLogInfo, ADALError.AUTH_FAILED_NO_TOKEN);
-
-                // remove item from cache to avoid same usage of
-                // refresh token in next acquireToken call
-                removeItemFromCache(refreshItem);
+            if (result == null) {
+                Logger.w(TAG, "Receive empty response from refresh token request.", request.getLogInfo(), 
+                        ADALError.AUTH_FAILED_NO_TOKEN);
+                return result;
+            } else if (StringExtensions.IsNullOrBlank(result.getAccessToken())) {
+                final String errorLogInfo = result.getErrorLogInfo();
+                Logger.w(TAG, "Refresh token request does not successfully return accesstoken.", 
+                        request.getLogInfo() + errorLogInfo, ADALError.AUTH_FAILED_NO_TOKEN);
+                
+                // check error code, only remove token from cache if receive invalid_grant from server
+                Logger.v(TAG + "tessting", "error_code:" + result.getErrorCode());
+                if (result.getErrorCode().equalsIgnoreCase(AuthenticationConstants.OAuth2ErrorCode.INVALID_GRANT)) {
+                    Logger.v(TAG, "Removing token cache for invalid_grant error returned from server.");
+                    removeItemFromCache(refreshItem);
+                }
+                
                 return result;
             } else {
                 Logger.v(TAG, "It finished refresh token request:" + request.getLogInfo());
--- a/src/src/com/microsoft/aad/adal/AuthenticationContext.java
+++ b/src/src/com/microsoft/aad/adal/AuthenticationContext.java
@@ -1441,10 +1441,14 @@ public class AuthenticationContext {
                 callbackHandle.onError(authenticationException);
                 return null;
             } catch (HttpResponseException httpResponseException) {
-                final AuthenticationException authenticationException = new AuthenticationException(ADALError.AUTH_REFRESH_FAILED_PROMPT_NOT_ALLOWED, 
-                        httpResponseException.getMessage() + request.getLogInfo());
-                callbackHandle.onError(authenticationException);
-                return null;
+                // Return back the http error for silent request
+                // Try interactive flow for non-silent request.
+                if (request.isSilent()) {
+                    final AuthenticationException authenticationException = new AuthenticationException(ADALError.AUTH_REFRESH_FAILED_PROMPT_NOT_ALLOWED, 
+                            httpResponseException.getMessage() + request.getLogInfo());
+                    callbackHandle.onError(authenticationException);
+                    return null;
+                }
             }
             
             if (authResult != null && !StringExtensions.IsNullOrBlank(authResult.getAccessToken())) {
@@ -2106,6 +2110,7 @@ public class AuthenticationContext {
                 } catch (final HttpResponseException httpResponseException) {
                     final AuthenticationException authenticationException = new AuthenticationException(
                             ADALError.AUTH_REFRESH_FAILED_PROMPT_NOT_ALLOWED, httpResponseException.getMessage());
+                    callbackHandle.onError(authenticationException);
                     return;
                 }
                 
--- a/src/src/com/microsoft/aad/adal/AuthenticationContext.java
+++ b/src/src/com/microsoft/aad/adal/AuthenticationContext.java
@@ -1871,11 +1871,11 @@ public class AuthenticationContext {
                 return result;
             } else if (StringExtensions.IsNullOrBlank(result.getAccessToken())) {
                 final String errorLogInfo = result.getErrorLogInfo();
-                Logger.w(TAG, "Refresh token request failed to return accesstoken.", 
+                Logger.e(TAG, "Refresh token request failed to return accesstoken.", 
                         request.getLogInfo() + errorLogInfo, ADALError.AUTH_FAILED_NO_TOKEN);
                 
                 // check error code, only remove token from cache if receive invalid_grant from server
-                if (result.getErrorCode().equalsIgnoreCase(AuthenticationConstants.OAuth2ErrorCode.INVALID_GRANT)) {
+                if (AuthenticationConstants.OAuth2ErrorCode.INVALID_GRANT.equalsIgnoreCase(result.getErrorCode())) {
                     Logger.v(TAG, "Removing token cache for invalid_grant error returned from server.");
                     removeItemFromCache(refreshItem);
                 }
--- a/src/src/com/microsoft/aad/adal/AuthenticationContext.java
+++ b/src/src/com/microsoft/aad/adal/AuthenticationContext.java
@@ -915,7 +915,7 @@ public class AuthenticationContext {
                                     result = oauthRequest.getToken(endingUrl);
                                     Logger.v(TAG, "OnActivityResult processed the result. "
                                             + authenticationRequest.getLogInfo());
-                                } catch (IOException | AuthenticationException exc) {
+                                } catch (IOException | AuthenticationException | HttpResponseException exc) {
                                     String msg = "Error in processing code to get token. "
                                             + authenticationRequest.getLogInfo() + correlationInfo;
                                     Logger.e(TAG, msg,
@@ -1440,6 +1440,11 @@ public class AuthenticationContext {
             } catch (AuthenticationException authenticationException) {
                 callbackHandle.onError(authenticationException);
                 return null;
+            } catch (HttpResponseException httpResponseException) {
+                final AuthenticationException authenticationException = new AuthenticationException(ADALError.AUTH_REFRESH_FAILED_PROMPT_NOT_ALLOWED, 
+                        httpResponseException.getMessage() + request.getLogInfo());
+                callbackHandle.onError(authenticationException);
+                return null;
             }
             
             if (authResult != null && !StringExtensions.IsNullOrBlank(authResult.getAccessToken())) {
@@ -1461,8 +1466,14 @@ public class AuthenticationContext {
                 // User does not want to launch activity
                 Logger.e(TAG, "Prompt is not allowed and failed to get token:", request.getLogInfo() + " " + errorInfo,
                         ADALError.AUTH_REFRESH_FAILED_PROMPT_NOT_ALLOWED);
-                callbackHandle.onError(new AuthenticationException(
-                        ADALError.AUTH_REFRESH_FAILED_PROMPT_NOT_ALLOWED, request.getLogInfo() + " " + errorInfo));
+                if (authResult != null && authResult.getErrorCode() != null) {
+                    callbackHandle.onError(new AuthenticationException(ADALError.AUTH_REFRESH_FAILED_PROMPT_NOT_ALLOWED, 
+                            request.getLogInfo() + " " + errorInfo, new AuthenticationServerProtocolException(ADALError.OAUTH_ERROR,
+                                    authResult.getErrorCode(), authResult.getErrorDescription())));
+                } else {
+                    callbackHandle.onError(new AuthenticationException(
+                            ADALError.AUTH_REFRESH_FAILED_PROMPT_NOT_ALLOWED, request.getLogInfo() + " " + errorInfo));
+                }
             }
         }
 
@@ -1821,10 +1832,11 @@ public class AuthenticationContext {
      * @param useCache refresh request can be explicit without cache usage.
      *            Error message should return without trying prompt.
      * @return
+     * @throws HttpResponseException 
      */
     private AuthenticationResult getTokenWithRefreshToken(final IWindowComponent activity, final boolean useDialog,
             final AuthenticationRequest request, final RefreshItem refreshItem, final boolean useCache) 
-            throws AuthenticationException {
+            throws AuthenticationException, HttpResponseException {
         Logger.v(TAG, "Process refreshToken for " + request.getLogInfo() + " refreshTokenId:"
                 + getTokenHash(refreshItem.mRefreshToken));
 
@@ -2086,13 +2098,14 @@ public class AuthenticationContext {
                 // Follow refresh logic now. Authority is valid or
                 // skipped validation
                 final AuthenticationResult authResult;
-                try
-                {
+                try {
                     authResult = getTokenWithRefreshToken(null, false, request, refreshItem, false);
-                }
-                catch (final AuthenticationException authException)
-                {
+                } catch (final AuthenticationException authException) {
                     callbackHandle.onError(authException);
+                    return;
+                } catch (final HttpResponseException httpResponseException) {
+                    final AuthenticationException authenticationException = new AuthenticationException(
+                            ADALError.AUTH_REFRESH_FAILED_PROMPT_NOT_ALLOWED, httpResponseException.getMessage());
                     return;
                 }
                 
--- a/src/src/com/microsoft/aad/adal/AuthenticationContext.java
+++ b/src/src/com/microsoft/aad/adal/AuthenticationContext.java
@@ -1866,12 +1866,12 @@ public class AuthenticationContext {
         // the error through callback. If true, just return the result back to localflow. 
         if (useCache) {
             if (result == null) {
-                Logger.w(TAG, "Receive empty response from refresh token request.", request.getLogInfo(), 
+                Logger.w(TAG, "Received empty response from refresh token request.", request.getLogInfo(), 
                         ADALError.AUTH_FAILED_NO_TOKEN);
                 return result;
             } else if (StringExtensions.IsNullOrBlank(result.getAccessToken())) {
                 final String errorLogInfo = result.getErrorLogInfo();
-                Logger.w(TAG, "Refresh token request does not successfully return accesstoken.", 
+                Logger.w(TAG, "Refresh token request failed to return accesstoken.", 
                         request.getLogInfo() + errorLogInfo, ADALError.AUTH_FAILED_NO_TOKEN);
                 
                 // check error code, only remove token from cache if receive invalid_grant from server
--- a/src/src/com/microsoft/aad/adal/AuthenticationServerProtocolException.java
+++ b/src/src/com/microsoft/aad/adal/AuthenticationServerProtocolException.java
@@ -32,7 +32,7 @@ public class AuthenticationServerProtocolException extends AuthenticationExcepti
     
     private String mProtocolCode;
 
-    protected AuthenticationServerProtocolException(final ADALError adalError, String detailMessage) {
+    protected AuthenticationServerProtocolException(final ADALError adalError, final String detailMessage) {
         super(adalError, detailMessage);
     }
     
--- a/src/src/com/microsoft/aad/adal/HttpResponseException.java
+++ b/src/src/com/microsoft/aad/adal/HttpResponseException.java
@@ -22,27 +22,30 @@
 // THE SOFTWARE.
 
 package com.microsoft.aad.adal;
-//
+
 /**
- * Represents server returned exception. 
+ * Represents HTTP error response. 
  */
-public class AuthenticationServerProtocolException extends AuthenticationException {
+class HttpResponseException extends Exception {
 
-    static final long serialVersionUID = 1;
+    /**
+     * Default serialized version
+     */
+    private static final long serialVersionUID = 1L;
     
-    private String mProtocolCode;
-
-    protected AuthenticationServerProtocolException(final ADALError adalError, String detailMessage) {
-        super(adalError, detailMessage);
+    private String mErrorCode;
+    
+    protected HttpResponseException(final String errorCode) {
+        super();
+        this.mErrorCode = errorCode;
     }
     
-    protected AuthenticationServerProtocolException(final ADALError adalError, final String protocolErrorCode, final String detailMessage) {
-        super(adalError, detailMessage);
-        this.mProtocolCode = protocolErrorCode;
+    protected HttpResponseException(final String errorCode, final String errorMessage) {
+        super(errorMessage);
+        this.mErrorCode = errorCode;
     }
     
-    public String getProtocolErrorCode(){
-        return mProtocolCode;
+    public String getErrorCode() {
+        return mErrorCode;
     }
 }
-
--- a/src/src/com/microsoft/aad/adal/Oauth2.java
+++ b/src/src/com/microsoft/aad/adal/Oauth2.java
@@ -635,7 +635,7 @@ class Oauth2 {
         return result;
     }
     
-    private AuthenticationResult parseJsonResponse(byte[] responseBody) throws JSONException {
+    private AuthenticationResult parseJsonResponse(final byte[] responseBody) throws JSONException {
         HashMap<String, String> responseItems = new HashMap<String, String>();
         
         final String jsonStringResult = new String(responseBody);
--- a/src/src/com/microsoft/aad/adal/Oauth2.java
+++ b/src/src/com/microsoft/aad/adal/Oauth2.java
@@ -522,7 +522,7 @@ class Oauth2 {
 
             if (result == null) {
                 // non-protocol related error
-                final String errMessage = response.getBody() == null ? 
+                final String errMessage = (response.getBody() == null || response.getBody().length == 0) ? 
                         "Status code:" + String.valueOf(response.getStatusCode()) : new String(response.getBody());
 
                 Logger.e(TAG, "Server error message", errMessage, ADALError.SERVER_ERROR);
@@ -608,11 +608,11 @@ class Oauth2 {
                     final String message = "Status code: " + statusCode + ". Cannot parse response as JSON " 
                             + ex.getMessage();
                     Logger.e(TAG, message, "", ADALError.SERVER_INVALID_JSON_RESPONSE, ex);
-                    result = new AuthenticationResult(String.valueOf(statusCode), webResponse.getBody().toString(), null);
+                    result = new AuthenticationResult(String.valueOf(statusCode), new String(webResponse.getBody()), null);
                 }
                 break;
             default:
-                result = new AuthenticationResult(String.valueOf(statusCode), webResponse.getBody().toString(), null);
+                result = new AuthenticationResult(String.valueOf(statusCode), new String(webResponse.getBody()), null);
             }
         }
 
--- a/src/src/com/microsoft/aad/adal/Oauth2.java
+++ b/src/src/com/microsoft/aad/adal/Oauth2.java
@@ -349,7 +349,7 @@ class Oauth2 {
         }
     }
 
-    public AuthenticationResult refreshToken(String refreshToken) throws IOException, AuthenticationException, HttpResponseException {
+    public AuthenticationResult refreshToken(final String refreshToken) throws IOException, AuthenticationException, HttpResponseException {
         String requestMessage = null;
         if (mWebRequestHandler == null) {
             Logger.v(TAG, "Web request is not set correctly");
--- a/src/src/com/microsoft/aad/adal/Oauth2.java
+++ b/src/src/com/microsoft/aad/adal/Oauth2.java
@@ -515,8 +515,15 @@ class Oauth2 {
             // Protocol related errors will read the error stream and report
             // the error and error description
             result = processTokenResponse(response);
-            ClientMetrics.INSTANCE.setLastError(result.getErrorCode());
-            ClientMetrics.INSTANCE.setLastErrorCodes(result.getErrorCodes());
+            
+            if (result == null) {
+                ClientMetrics.INSTANCE.setLastError(null);
+                if (response.getResponseException() != null) {
+                    throw response.getResponseException();
+                }
+            } else {
+                ClientMetrics.INSTANCE.setLastErrorCodes(result.getErrorCodes());
+            }
         } catch (UnsupportedEncodingException e) {
             ClientMetrics.INSTANCE.setLastError(null);
             Logger.e(TAG, e.getMessage(), "", ADALError.ENCODING_IS_NOT_SUPPORTED, e);
@@ -580,7 +587,7 @@ class Oauth2 {
             Logger.v(TAG, "Token request returned an empty response body.");
             throw new HttpResponseException(String.valueOf(statusCode));
         } else {
-            Logger.v(TAG, "Token request does not have exception.");
+            Logger.v(TAG, "Start parsing response body from token request.");
             switch(statusCode) {
             case HttpURLConnection.HTTP_OK:
             case HttpURLConnection.HTTP_BAD_REQUEST:
--- a/src/src/com/microsoft/aad/adal/Oauth2.java
+++ b/src/src/com/microsoft/aad/adal/Oauth2.java
@@ -590,11 +590,11 @@ class Oauth2 {
             }
         }
         
-        final boolean isResponseBodyEmpty = webResponse.getBody() == null || webResponse.getBody().length == 0;
+        final boolean isResponseBodyEmpty = (webResponse.getBody() == null) || (webResponse.getBody().length == 0);
         final int statusCode = webResponse.getStatusCode();
         if (isResponseBodyEmpty) {
             // Empty response body, return the status code as error code back
-            Logger.v(TAG, "Token request return an empty response body.");
+            Logger.v(TAG, "Token request returned an empty response body.");
             result = new AuthenticationResult(String.valueOf(statusCode), "", null);
         } else {
             switch(statusCode) {
@@ -605,7 +605,9 @@ class Oauth2 {
                     result = parseJsonResponse(webResponse.getBody());
                 } catch (final JSONException ex) {
                     // Cannot parse response as JSON, then return the raw response as error description
-                    Logger.e(TAG, "Cannot parse response as JSON" + ex.getMessage(), "", ADALError.SERVER_INVALID_JSON_RESPONSE, ex);
+                    final String message = "Status code: " + statusCode + ". Cannot parse response as JSON " 
+                            + ex.getMessage();
+                    Logger.e(TAG, message, "", ADALError.SERVER_INVALID_JSON_RESPONSE, ex);
                     result = new AuthenticationResult(String.valueOf(statusCode), webResponse.getBody().toString(), null);
                 }
                 break;
--- a/src/src/com/microsoft/aad/adal/Oauth2.java
+++ b/src/src/com/microsoft/aad/adal/Oauth2.java
@@ -349,7 +349,7 @@ class Oauth2 {
         }
     }
 
-    public AuthenticationResult refreshToken(String refreshToken) throws IOException, AuthenticationException {
+    public AuthenticationResult refreshToken(String refreshToken) throws IOException, AuthenticationException, HttpResponseException {
         String requestMessage = null;
         if (mWebRequestHandler == null) {
             Logger.v(TAG, "Web request is not set correctly");
@@ -381,10 +381,11 @@ class Oauth2 {
      *            or token for next step
      * @return Token in the AuthenticationResult. Null result if response does
      *         not have protocol error.
+     * @throws HttpResponseException 
      * @throws Exception
      */
     public AuthenticationResult getToken(String authorizationUrl)
-            throws IOException, AuthenticationServerProtocolException, AuthenticationException {
+            throws IOException, AuthenticationException, HttpResponseException {
 
         if (StringExtensions.IsNullOrBlank(authorizationUrl)) {
             throw new IllegalArgumentException("authorizationUrl");
@@ -431,9 +432,10 @@ class Oauth2 {
      * 
      * @param code
      * @return Token in the AuthenticationResult
+     * @throws HttpResponseException 
      * @throws Exception
      */
-    public AuthenticationResult getTokenForCode(String code) throws IOException, AuthenticationException {
+    public AuthenticationResult getTokenForCode(String code) throws IOException, AuthenticationException, HttpResponseException {
 
         String requestMessage = null;
         if (mWebRequestHandler == null) {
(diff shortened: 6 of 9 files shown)


src/src/com/microsoft/aad/adal/Oauth2.java
```diff
@@ -595,33 +589,29 @@ private AuthenticationResult processTokenResponse(HttpWebResponse webResponse) {
...
-                errMessage = "Status code:" + String.valueOf(webResponse.getStatusCode());
+            switch(statusCode) {
+            case HttpURLConnection.HTTP_OK:
+            case HttpURLConnection.HTTP_BAD_REQUEST:
+            case HttpURLConnection.HTTP_UNAUTHORIZED: 
+                try {
+                    result = parseJsonResponse(webResponse.getBody());
+                } catch (final JSONException ex) {
+                    // Cannot parse response as JSON, then return the raw response as error description
+                    Logger.e(TAG, "Cannot parse response as JSON" + ex.getMessage(), "", ADALError.SERVER_INVALID_JSON_RESPONSE, ex);
```
RandalliLama: @weijjia Should we log the status code here as well?

src/src/com/microsoft/aad/adal/AuthenticationContext.java
```diff
@@ -1865,14 +1865,21 @@ private AuthenticationResult getTokenWithRefreshToken(final IWindowComponent act
...
-            if (result == null || StringExtensions.IsNullOrBlank(result.getAccessToken())) {
-                String errLogInfo = result == null ? "" : result.getErrorLogInfo();
-                Logger.w(TAG, "Refresh token did not return accesstoken.", request.getLogInfo()
-                        + errLogInfo, ADALError.AUTH_FAILED_NO_TOKEN);
-
-                // remove item from cache to avoid same usage of
-                // refresh token in next acquireToken call
-                removeItemFromCache(refreshItem);
+            if (result == null) {
+                Logger.w(TAG, "Receive empty response from refresh token request.", request.getLogInfo(),
```
kpanwar: received*. prefer Past tense.

src/src/com/microsoft/aad/adal/AuthenticationContext.java
```diff
@@ -1865,14 +1865,21 @@ private AuthenticationResult getTokenWithRefreshToken(final IWindowComponent act
...
-                // remove item from cache to avoid same usage of
-                // refresh token in next acquireToken call
-                removeItemFromCache(refreshItem);
+            if (result == null) {
+                Logger.w(TAG, "Receive empty response from refresh token request.", request.getLogInfo(), 
+                        ADALError.AUTH_FAILED_NO_TOKEN);
+                return result;
+            } else if (StringExtensions.IsNullOrBlank(result.getAccessToken())) {
+                final String errorLogInfo = result.getErrorLogInfo();
+                Logger.w(TAG, "Refresh token request does not successfully return accesstoken.",
```
kpanwar: instead of "did not successfully", consider putting "failed to"

src/src/com/microsoft/aad/adal/Oauth2.java
weijjia: Yup, we should. I'll update it.

src/src/com/microsoft/aad/adal/AuthenticationContext.java
weijjia: sure, will have it updated.

src/src/com/microsoft/aad/adal/Oauth2.java
```diff
@@ -595,33 +589,29 @@ private AuthenticationResult processTokenResponse(HttpWebResponse webResponse) {
...
-                result = processUIResponseParams(responseItems);
-            } catch (final JSONException ex) {
-                // There is no recovery possible here, so
-                // catch the
-                // generic Exception
-                Logger.e(TAG, ex.getMessage(), "", ADALError.SERVER_INVALID_JSON_RESPONSE, ex);
-                result = new AuthenticationResult(JSON_PARSING_ERROR, ex.getMessage(), null);
-            }
+        
+        final boolean isResponseBodyEmpty = webResponse.getBody() == null || webResponse.getBody().length == 0;
```
kpanwar: consider putting brackets around each comparison. this way it will be harder to introduce a bug.

src/src/com/microsoft/aad/adal/Oauth2.java
```diff
@@ -595,33 +589,29 @@ private AuthenticationResult processTokenResponse(HttpWebResponse webResponse) {
...
+        final int statusCode = webResponse.getStatusCode();
+        if (isResponseBodyEmpty) {
+            // Empty response body, return the status code as error code back
+            Logger.v(TAG, "Token request return an empty response body.");
```
kpanwar: returned*

RandalliLama:
:shipit: with @kpanwar 's approval.

src/src/com/microsoft/aad/adal/Oauth2.java
```diff
@@ -595,33 +589,29 @@ private AuthenticationResult processTokenResponse(HttpWebResponse webResponse) {
...
+            switch(statusCode) {
+            case HttpURLConnection.HTTP_OK:
+            case HttpURLConnection.HTTP_BAD_REQUEST:
+            case HttpURLConnection.HTTP_UNAUTHORIZED: 
+                try {
+                    result = parseJsonResponse(webResponse.getBody());
+                } catch (final JSONException ex) {
+                    // Cannot parse response as JSON, then return the raw response as error description
+                    Logger.e(TAG, "Cannot parse response as JSON" + ex.getMessage(), "", ADALError.SERVER_INVALID_JSON_RESPONSE, ex);
+                    result = new AuthenticationResult(String.valueOf(statusCode), webResponse.getBody().toString(), null);
```
kpanwar: webresponse.getbody() could be null?

src/src/com/microsoft/aad/adal/Oauth2.java
```diff
@@ -595,33 +589,29 @@ private AuthenticationResult processTokenResponse(HttpWebResponse webResponse) {
...
+                }
+                break;
+            default:
+                result = new AuthenticationResult(String.valueOf(statusCode), webResponse.getBody().toString(), null);
```
kpanwar: webresponse.getbody() could be null.

src/src/com/microsoft/aad/adal/Oauth2.java
RandalliLama: @kpanwar You mean parenthesis, right?

src/src/com/microsoft/aad/adal/Oauth2.java
weijjia: nope, if check already cover the case webresponse.getBody==null

src/src/com/microsoft/aad/adal/Oauth2.java
weijjia: same as above

src/src/com/microsoft/aad/adal/AuthenticationContext.java
```diff
@@ -1865,14 +1865,21 @@ private AuthenticationResult getTokenWithRefreshToken(final IWindowComponent act
...
-                // remove item from cache to avoid same usage of
-                // refresh token in next acquireToken call
-                removeItemFromCache(refreshItem);
+            if (result == null) {
+                Logger.w(TAG, "Received empty response from refresh token request.", request.getLogInfo(), 
+                        ADALError.AUTH_FAILED_NO_TOKEN);
+                return result;
+            } else if (StringExtensions.IsNullOrBlank(result.getAccessToken())) {
+                final String errorLogInfo = result.getErrorLogInfo();
+                Logger.w(TAG, "Refresh token request failed to return accesstoken.",
```
agamemnum: nit: please make it "error" level so our code can send it in instrumentation.

kpanwar:
:shipit:

src/src/com/microsoft/aad/adal/AuthenticationContext.java
```diff
@@ -1865,14 +1865,21 @@ private AuthenticationResult getTokenWithRefreshToken(final IWindowComponent act
...
+                Logger.w(TAG, "Received empty response from refresh token request.", request.getLogInfo(), 
+                        ADALError.AUTH_FAILED_NO_TOKEN);
+                return result;
+            } else if (StringExtensions.IsNullOrBlank(result.getAccessToken())) {
+                final String errorLogInfo = result.getErrorLogInfo();
+                Logger.w(TAG, "Refresh token request failed to return accesstoken.", 
+                        request.getLogInfo() + errorLogInfo, ADALError.AUTH_FAILED_NO_TOKEN);
+                
+                // check error code, only remove token from cache if receive invalid_grant from server
+                if (result.getErrorCode().equalsIgnoreCase(AuthenticationConstants.OAuth2ErrorCode.INVALID_GRANT)) {
```
agamemnum: Consider changing an order of comparison to avoid NPE - i.e. write `AuthenticationConstants.OAuth2ErrorCode.INVALID_GRANT. equalsIgnoreCase(result.getErrorCode())`

src/src/com/microsoft/aad/adal/Oauth2.java
```diff
@@ -522,15 +522,10 @@ private AuthenticationResult postMessage(String requestMessage, HashMap<String,
...
             if (result == null) {
                 // non-protocol related error
-                String errMessage = null;
-                byte[] message = response.getBody();
-                if (message != null) {
-                    errMessage = new String(message);
-                } else {
-                    errMessage = "Status code:" + String.valueOf(response.getStatusCode());
-                }
+                final String errMessage = response.getBody() == null ?
```
agamemnum: It would be useful to check of empty size in here. I.e. `response.getBody() == null || response.getBody().size() == 0`

src/src/com/microsoft/aad/adal/AuthenticationContext.java
weijjia: done

src/src/com/microsoft/aad/adal/AuthenticationContext.java
weijjia: sure, will do it.

src/src/com/microsoft/aad/adal/Oauth2.java
```diff
@@ -595,33 +589,31 @@ private AuthenticationResult processTokenResponse(HttpWebResponse webResponse) {
...
+            case HttpURLConnection.HTTP_BAD_REQUEST:
+            case HttpURLConnection.HTTP_UNAUTHORIZED: 
+                try {
+                    result = parseJsonResponse(webResponse.getBody());
+                } catch (final JSONException ex) {
+                    // Cannot parse response as JSON, then return the raw response as error description
+                    final String message = "Status code: " + statusCode + ". Cannot parse response as JSON " 
+                            + ex.getMessage();
+                    Logger.e(TAG, message, "", ADALError.SERVER_INVALID_JSON_RESPONSE, ex);
+                    result = new AuthenticationResult(String.valueOf(statusCode), webResponse.getBody().toString(), null);
```
agamemnum: Did you mean `new String(webResponse.getBody())` instead of `webResponse.getBody().toString()`?

src/src/com/microsoft/aad/adal/Oauth2.java
weijjia: yup, sure. will update it.

src/src/com/microsoft/aad/adal/Oauth2.java
weijjia: ah, right, updated.

src/src/com/microsoft/aad/adal/Oauth2.java
```diff
@@ -642,4 +634,12 @@ private AuthenticationResult processTokenResponse(HttpWebResponse webResponse) {
 
         return result;
     }
+    
+    private AuthenticationResult parseJsonResponse(byte[] responseBody) throws JSONException {
+        HashMap<String, String> responseItems = new HashMap<String, String>();
```
agamemnum: nit: for consistency make this `final` too?

tests/Functional/src/com/microsoft/aad/adal/test/AuthenticationContextTest.java
```diff
@@ -1364,6 +1364,64 @@ public void testRefreshTokenRequestWithFamilyIdFailed() throws InterruptedExcept
...
+        final CountDownLatch signal = new CountDownLatch(1);
+        testActivity.mSignal = signal;
+        MockWebRequestHandler webrequest = new MockWebRequestHandler();
+        String responseBody = "{\"error\":\"invalid_grant\",\"error_description\":\"AADSTS70000: Authentication failed. Refresh Token is not valid.\r\nTrace ID: bb27293d-74e4-4390-882b-037a63429026\r\nCorrelation ID: b73106d5-419b-4163-8bc6-d2c18f1b1a13\r\nTimestamp: 2014-11-06 18:39:47Z\",\"error_codes\":[70000],\"timestamp\":\"2014-11-06 18:39:47Z\",\"trace_id\":\"bb27293d-74e4-4390-882b-037a63429026\",\"correlation_id\":\"b73106d5-419b-4163-8bc6-d2c18f1b1a13\",\"submit_url\":null,\"context\":null}";
+        webrequest.setReturnResponse(new HttpWebResponse(400, responseBody.getBytes(), null));
+        ReflectionUtils.setFieldValue(context, "mWebRequest", webrequest);
+        
+        try {
+            context.acquireTokenSilentSync("resource", "clientid", TEST_IDTOKEN_USERID);
+        } catch (AuthenticationException e) {
```
agamemnum: nit: assert here always - as you expect an exception

tests/Functional/src/com/microsoft/aad/adal/test/AuthenticationContextTest.java
```diff
@@ -1364,6 +1364,64 @@ public void testRefreshTokenRequestWithFamilyIdFailed() throws InterruptedExcept
...
+        final MockActivity testActivity = new MockActivity();
+        final CountDownLatch signal = new CountDownLatch(1);
+        testActivity.mSignal = signal;
+        MockWebRequestHandler webrequest = new MockWebRequestHandler();
+        String responseBody = "{\"error\":\"interaction_required\",\"error_description\":\"AADSTS70000: Authentication failed. Refresh Token is not valid.\r\nTrace ID: bb27293d-74e4-4390-882b-037a63429026\r\nCorrelation ID: b73106d5-419b-4163-8bc6-d2c18f1b1a13\r\nTimestamp: 2014-11-06 18:39:47Z\",\"error_codes\":[70000],\"timestamp\":\"2014-11-06 18:39:47Z\",\"trace_id\":\"bb27293d-74e4-4390-882b-037a63429026\",\"correlation_id\":\"b73106d5-419b-4163-8bc6-d2c18f1b1a13\",\"submit_url\":null,\"context\":null}";
+        webrequest.setReturnResponse(new HttpWebResponse(400, responseBody.getBytes(), null));
+        ReflectionUtils.setFieldValue(context, "mWebRequest", webrequest);
+        
+        try {
+            context.acquireTokenSilentSync("resource", "clientid", TEST_IDTOKEN_USERID);
```
agamemnum: Same here - we need to sure that exception is actually thrown.

tests/Functional/src/com/microsoft/aad/adal/test/AuthenticationContextTest.java
```diff
@@ -1364,6 +1364,64 @@ public void testRefreshTokenRequestWithFamilyIdFailed() throws InterruptedExcept
...
+        } catch (AuthenticationException e) {
```
agamemnum: I'm not an expert here, but shouldn't we wait for `signal` in here?

tests/Functional/src/com/microsoft/aad/adal/test/AuthenticationContextTest.java
weijjia: I got it. Good catch, thanks.

tests/Functional/src/com/microsoft/aad/adal/test/AuthenticationContextTest.java
```diff
@@ -1364,6 +1364,64 @@ public void testRefreshTokenRequestWithFamilyIdFailed() throws InterruptedExcept
         clearCache(context);
     }
     
+    @SmallTest
+    public void testRefreshTokenWithInvalidGrantReturned_CacheCleared() throws NoSuchFieldException, IllegalAccessException, InterruptedException {
```
imakhalova: where is the test for error response?

weijjia: what do you mean by error response?

src/src/com/microsoft/aad/adal/AuthenticationContext.java
```diff
@@ -1865,14 +1865,21 @@ private AuthenticationResult getTokenWithRefreshToken(final IWindowComponent act
...
-            if (result == null || StringExtensions.IsNullOrBlank(result.getAccessToken())) {
-                String errLogInfo = result == null ? "" : result.getErrorLogInfo();
-                Logger.w(TAG, "Refresh token did not return accesstoken.", request.getLogInfo()
-                        + errLogInfo, ADALError.AUTH_FAILED_NO_TOKEN);
-
-                // remove item from cache to avoid same usage of
-                // refresh token in next acquireToken call
-                removeItemFromCache(refreshItem);
+            if (result == null) {
+                Logger.w(TAG, "Received empty response from refresh token request.", request.getLogInfo(),
```
imakhalova: What do you mean by empty response here? in your code you'll always have non null result object

tests/Functional/src/com/microsoft/aad/adal/test/AuthenticationContextTest.java
```diff
@@ -984,14 +984,14 @@ public void testRefreshTokenWebRequestHasError() throws InterruptedException,
         webrequest.setReturnResponse(new HttpWebResponse(500, null, null));
         ReflectionUtils.setFieldValue(context, "mWebRequest", webrequest);
         final TestLogResponse response = new TestLogResponse();
-        response.listenLogForMessageSegments(signal, "Refresh token did not return accesstoken");
+        response.listenLogForMessageSegments(signal, "Server error message");
```
imakhalova: this function unlocks signal before acquireTokenSilentAsync exits...
you should wait until acquireTokenSilentAsync  invokes callback,
and check that you get exception in result

tests/Functional/src/com/microsoft/aad/adal/test/AuthenticationContextTest.java
imakhalova: Some 404 or any other error with empty body in response

weijjia: An existing test with 500 returned already covers it.

src/src/com/microsoft/aad/adal/Oauth2.java
```diff
@@ -595,33 +589,31 @@ private AuthenticationResult processTokenResponse(HttpWebResponse webResponse) {
...
-                result = processUIResponseParams(responseItems);
-            } catch (final JSONException ex) {
-                // There is no recovery possible here, so
-                // catch the
-                // generic Exception
-                Logger.e(TAG, ex.getMessage(), "", ADALError.SERVER_INVALID_JSON_RESPONSE, ex);
-                result = new AuthenticationResult(JSON_PARSING_ERROR, ex.getMessage(), null);
-            }
+        
+        final boolean isResponseBodyEmpty = (webResponse.getBody() == null) || (webResponse.getBody().length == 0);
```
imakhalova: It looks like processTokenResponse function is called in condition getBody()!=null
Do we want to remove this condition? since we're checking it inside this function?

src/src/com/microsoft/aad/adal/Oauth2.java
```diff
@@ -522,15 +522,10 @@ private AuthenticationResult postMessage(String requestMessage, HashMap<String,
...
-                    errMessage = new String(message);
-                } else {
-                    errMessage = "Status code:" + String.valueOf(response.getStatusCode());
-                }
+                final String errMessage = (response.getBody() == null || response.getBody().length == 0) ? 
+                        "Status code:" + String.valueOf(response.getStatusCode()) : new String(response.getBody());
 
-                Logger.v(TAG, "Server error message:" + errMessage);
+                Logger.e(TAG, "Server error message", errMessage, ADALError.SERVER_ERROR);
                 if (response.getResponseException() != null) {
```
imakhalova: why don't we throw anything in case if we don't have Exception in response, but result is null?

src/src/com/microsoft/aad/adal/AuthenticationContext.java
```diff
@@ -1865,14 +1865,21 @@ private AuthenticationResult getTokenWithRefreshToken(final IWindowComponent act
...
+                Logger.e(TAG, "Refresh token request failed to return accesstoken.", 
+                        request.getLogInfo() + errorLogInfo, ADALError.AUTH_FAILED_NO_TOKEN);
+                
+                // check error code, only remove token from cache if receive invalid_grant from server
+                if (AuthenticationConstants.OAuth2ErrorCode.INVALID_GRANT.equalsIgnoreCase(result.getErrorCode())) {
+                    Logger.v(TAG, "Removing token cache for invalid_grant error returned from server.");
+                    removeItemFromCache(refreshItem);
+                }
+                
                 return result;
```
imakhalova: I don't think that it's actually good to return result instead of error. It means that user will get again AUTH_REFRESH_FAILED_PROMPT_NOT_ALLOWED which won't say a lot to him

imakhalova: It's much better if user will get some other error in that case, than he can handle it differently

src/src/com/microsoft/aad/adal/Oauth2.java
weijjia: oh, sure, I should get rid of this redundant check

src/src/com/microsoft/aad/adal/AuthenticationContext.java
imakhalova: I want to get some other error instead of AUTH_REFRESH_FAILED_PROMPT_NOT_ALLOWED
Otherwise how we are going to distinguish why we weren't able to get token
And this change won't make any sense to us because even if you don't remove token we will sign out user, because this error is no difference with expired token error

RandalliLama: @imakhalova Can this be handled in a seperate issue, or do you believe this should block this PR?

agamemnum: @RandalliLama if we can't handle an error properly this change won't help us much.

tests/Functional/src/com/microsoft/aad/adal/test/AuthenticationContextTest.java
imakhalova: It has null body... with your code null and empty is 2 different scenarios

src/src/com/microsoft/aad/adal/Oauth2.java
imakhalova: what I'm actually asking should we remove external check (not internal)?

src/src/com/microsoft/aad/adal/AuthenticationContext.java
RandalliLama: @agamemnum Understood.  We are thinking through solutions that do not break current consumers but are appropriately extensible in the future.  Stay tuned...

agamemnum: @RandalliLama what about exception type inherited from current one? If someone needs to handle specifics they can do more specific catch. For all old clients exception would look identical.

RandalliLama: @agamemnum We are going to expose the specific oauth error via the cause property on the existing exception.  The cause will contain a AuthenticationServerProtocolException whose ADALError property will be 'OAuth'.  There will be a protocolSpecificError string typed property that will contain the actual OAuth error code.  You can look for that and if you find the value 'invalid_grant' you will know that the account is broken and the user will need to sign-in again rather than simply retrying.

src/src/com/microsoft/aad/adal/AuthenticationContext.java
```diff
@@ -1440,6 +1440,11 @@ private AuthenticationResult localFlow(CallbackHandler callbackHandle,
             } catch (AuthenticationException authenticationException) {
                 callbackHandle.onError(authenticationException);
                 return null;
+            } catch (HttpResponseException httpResponseException) {
+                final AuthenticationException authenticationException = new AuthenticationException(ADALError.AUTH_REFRESH_FAILED_PROMPT_NOT_ALLOWED,
```
kpanwar: this should fail only on silent calls and not PromptBehavior.Auto

src/src/com/microsoft/aad/adal/Oauth2.java
```diff
@@ -612,11 +596,25 @@ private AuthenticationResult processTokenResponse(HttpWebResponse webResponse) {
                 }
                 break;
             default:
-                result = new AuthenticationResult(String.valueOf(statusCode), new String(webResponse.getBody()), null);
+                throw new HttpResponseException(String.valueOf(statusCode), new String(webResponse.getBody()));
```
RandalliLama: Line 595 above should also be treated as an exception.

https://github.com/AzureAD/azure-activedirectory-library-for-android/pull/588/files/eddc1d8b2f879e4efa48a2cba59c91ad347f789c..96417f1d1e98c4b8aa279f704aadb89086edfe71#diff-e52b28383a3d6563a748d17a7629dd8fR595

src/src/com/microsoft/aad/adal/AuthenticationContext.java
```diff
@@ -2086,14 +2098,15 @@ public void run() {
...
-                }
-                catch (final AuthenticationException authException)
-                {
+                } catch (final AuthenticationException authException) {
                     callbackHandle.onError(authException);
                     return;
+                } catch (final HttpResponseException httpResponseException) {
+                    final AuthenticationException authenticationException = new AuthenticationException(
+                            ADALError.AUTH_REFRESH_FAILED_PROMPT_NOT_ALLOWED, httpResponseException.getMessage());
+                    return;
```
kpanwar: empty return? should this be a throw?

src/src/com/microsoft/aad/adal/Oauth2.java
```diff
@@ -595,8 +578,9 @@ private AuthenticationResult processTokenResponse(HttpWebResponse webResponse) {
         if (isResponseBodyEmpty) {
             // Empty response body, return the status code as error code back
             Logger.v(TAG, "Token request returned an empty response body.");
-            result = new AuthenticationResult(String.valueOf(statusCode), "", null);
+            throw new HttpResponseException(String.valueOf(statusCode));
         } else {
+            Logger.v(TAG, "Token request does not have exception.");
```
RandalliLama: This log entry doesn't make sense to me.  Was is it trying to communicate?  Should it just be removed?

src/src/com/microsoft/aad/adal/Oauth2.java
```diff
@@ -504,42 +506,21 @@ private AuthenticationResult postMessage(String requestMessage, HashMap<String,
...
+            // Protocol related errors will read the error stream and report
+            // the error and error description
+            result = processTokenResponse(response);
+            ClientMetrics.INSTANCE.setLastError(result.getErrorCode());
+            ClientMetrics.INSTANCE.setLastErrorCodes(result.getErrorCodes());
         } catch (UnsupportedEncodingException e) {
             ClientMetrics.INSTANCE.setLastError(null);
             Logger.e(TAG, e.getMessage(), "", ADALError.ENCODING_IS_NOT_SUPPORTED, e);
             throw e;
-        } catch (IOException e) {
```
RandalliLama: Why did you remove this?

weijjia: no one is throwing it.

tests/Functional/src/com/microsoft/aad/adal/test/AuthenticationContextTest.java
weijjia: updated.

tests/Functional/src/com/microsoft/aad/adal/test/AuthenticationContextTest.java
```diff
@@ -1467,7 +1467,70 @@ public void testRefreshTokenWithInteractionRequired_CacheNotCleared() throws NoS
...
+                AuthenticationConstants.UIRequest.BROWSER_FLOW,
+                testActivity.mStartActivityRequestCode);
+
+        context.onActivityResult(testActivity.mStartActivityRequestCode,
+                AuthenticationConstants.UIResponse.BROWSER_CODE_COMPLETE, getResponseIntent(callback, "resource", "clientid", "redirect", TEST_IDTOKEN_UPN));
+        signalCallback.await(CONTEXT_REQUEST_TIME_OUT, TimeUnit.MILLISECONDS);
+    }
     
+    @SmallTest
+    public void testAutoFlow_RefreshTokenRequestFailed_WithHTTError() throws NoSuchFieldException, IllegalAccessException,
```
kpanwar: HTTPError*

src/src/com/microsoft/aad/adal/ADALError.java
```diff
@@ -515,7 +515,12 @@
     /**
      * Resource authentication challenge failure
      */
-    RESOURCE_AUTHENTICATION_CHALLENGE_FAILURE("Resource authentication challenge failure");
+    RESOURCE_AUTHENTICATION_CHALLENGE_FAILURE("Resource authentication challenge failure"),
+    
+    /**
+     * Represents a general OAUTH Error
+     */
+    OAUTH_ERROR("Oauth Error");
```
kpanwar: OAUTH2_ERROR

src/src/com/microsoft/aad/adal/AuthenticationServerProtocolException.java
```diff
@@ -22,16 +22,27 @@
...
-class AuthenticationServerProtocolException extends AuthenticationException {
+public class AuthenticationServerProtocolException extends AuthenticationException {
 
     static final long serialVersionUID = 1;
+    
+    private String mProtocolCode;
 
-    public AuthenticationServerProtocolException(String detailMessage) {
-        super(ADALError.DEVICE_CHALLENGE_FAILURE, detailMessage);
+    protected AuthenticationServerProtocolException(final ADALError adalError, String detailMessage) {
```
kpanwar: final detailMessage?

src/src/com/microsoft/aad/adal/Oauth2.java
```diff
@@ -349,7 +349,7 @@ private static void extractJsonObjects(HashMap<String, String> responseItems, St
         }
     }
 
-    public AuthenticationResult refreshToken(String refreshToken) throws IOException, AuthenticationException {
+    public AuthenticationResult refreshToken(String refreshToken) throws IOException, AuthenticationException, HttpResponseException {
```
kpanwar: final string

tests/Functional/src/com/microsoft/aad/adal/test/AuthenticationContextTest.java
```diff
@@ -1364,6 +1367,171 @@ public void testRefreshTokenRequestWithFamilyIdFailed() throws InterruptedExcept
...
+        } catch (AuthenticationException e) {
+            assertEquals("Token is not exchanged",
+                    ADALError.AUTH_REFRESH_FAILED_PROMPT_NOT_ALLOWED, e.getCode());
+            assertNotNull(e.getCause() instanceof AuthenticationServerProtocolException);
+            AuthenticationServerProtocolException throwable = (AuthenticationServerProtocolException)e.getCause();
+            assertTrue(throwable.getProtocolErrorCode().equals("invalid_grant"));
+        }
+        
+        // verify that the cache is cleared
+        assertNull(mockCache.getItem(CacheKey.createCacheKey(VALID_AUTHORITY, "resource", "clientId", false, TEST_IDTOKEN_USERID)));
```
kpanwar: make it simple by checking count?

tests/Functional/src/com/microsoft/aad/adal/test/AuthenticationContextTest.java
```diff
@@ -1364,6 +1367,171 @@ public void testRefreshTokenRequestWithFamilyIdFailed() throws InterruptedExcept
...
+                AuthenticationConstants.UIRequest.BROWSER_FLOW,
+                testActivity.mStartActivityRequestCode);
+
+        context.onActivityResult(testActivity.mStartActivityRequestCode,
+                AuthenticationConstants.UIResponse.BROWSER_CODE_COMPLETE, getResponseIntent(callback, "resource", "clientid", "redirect", TEST_IDTOKEN_UPN));
+        signalCallback.await(CONTEXT_REQUEST_TIME_OUT, TimeUnit.MILLISECONDS);
+    }
+    
+    @SmallTest
+    public void testAutoFlow_RefreshTokenRequestFailed_WithHTTError() throws NoSuchFieldException, IllegalAccessException,
```
kpanwar: HTTPError

kpanwar:
fix spellings and then :shipit:

tests/Functional/src/com/microsoft/aad/adal/test/AuthenticationContextTest.java
weijjia: will keep it for now, and refactor later.

src/src/com/microsoft/aad/adal/Oauth2.java
```diff
@@ -504,33 +506,18 @@ private AuthenticationResult postMessage(String requestMessage, HashMap<String,
...
-                Logger.v(TAG, "Token request does not have exception");
-                result = processTokenResponse(response);
-                ClientMetrics.INSTANCE.setLastError(null);
-            }
-
+            // Protocol related errors will read the error stream and report
+            // the error and error description
+            result = processTokenResponse(response);
+            
             if (result == null) {
```
imakhalova: Is it possible here to be result null?
What is the expected behavior if result is null and response.getResponseException() == null?

src/src/com/microsoft/aad/adal/Oauth2.java
```diff
@@ -504,33 +506,18 @@ private AuthenticationResult postMessage(String requestMessage, HashMap<String,
...
-                byte[] message = response.getBody();
-                if (message != null) {
-                    errMessage = new String(message);
-                } else {
-                    errMessage = "Status code:" + String.valueOf(response.getStatusCode());
-                }
-
-                Logger.v(TAG, "Server error message:" + errMessage);
+                ClientMetrics.INSTANCE.setLastError(null);
                 if (response.getResponseException() != null) {
```
imakhalova: not sure why do we set and get this exception instead of just throwing it?

src/src/com/microsoft/aad/adal/AuthenticationContext.java
```diff
@@ -1440,6 +1440,15 @@ private AuthenticationResult localFlow(CallbackHandler callbackHandle,
             } catch (AuthenticationException authenticationException) {
                 callbackHandle.onError(authenticationException);
                 return null;
+            } catch (HttpResponseException httpResponseException) {
+                // Return back the http error for silent request
+                // Try interactive flow for non-silent request.
+                if (request.isSilent()) {
+                    final AuthenticationException authenticationException = new AuthenticationException(ADALError.AUTH_REFRESH_FAILED_PROMPT_NOT_ALLOWED,
```
imakhalova: where is the sign that it was server failure?

src/src/com/microsoft/aad/adal/AuthenticationContext.java
```diff
@@ -1440,6 +1440,15 @@ private AuthenticationResult localFlow(CallbackHandler callbackHandle,
...
                 callbackHandle.onError(authenticationException);
                 return null;
+            } catch (HttpResponseException httpResponseException) {
+                // Return back the http error for silent request
+                // Try interactive flow for non-silent request.
+                if (request.isSilent()) {
+                    final AuthenticationException authenticationException = new AuthenticationException(ADALError.AUTH_REFRESH_FAILED_PROMPT_NOT_ALLOWED, 
+                            httpResponseException.getMessage() + request.getLogInfo());
+                    callbackHandle.onError(authenticationException);
+                    return null;
```
imakhalova: Because of early returns and multiple conditions it's hard to follow what happens, would be hard to maintain

src/src/com/microsoft/aad/adal/AuthenticationContext.java
```diff
@@ -2079,14 +2102,16 @@ public void run() {
...
-                }
-                catch (final AuthenticationException authException)
-                {
+                } catch (final AuthenticationException authException) {
                     callbackHandle.onError(authException);
                     return;
+                } catch (final HttpResponseException httpResponseException) {
+                    final AuthenticationException authenticationException = new AuthenticationException(
+                            ADALError.AUTH_REFRESH_FAILED_PROMPT_NOT_ALLOWED, httpResponseException.getMessage());
+                    callbackHandle.onError(authenticationException);
```
imakhalova: Include at least httpResponseException as cause

src/src/com/microsoft/aad/adal/AuthenticationContext.java
imakhalova: @RandalliLama can you please help me to understand how can you break users when you return another ADAL error type for server errors? I thought ADAL error codes were created specifically for this purpose to distinguish why token acquire failed

agamemnum: @RandalliLama I prefer for us to not take any dependence on `invalid_grant` or any other internal service implementation details. I really like to get an exception that indicates to client that Network Error have occurred (ADAL treated an error as Network Error). With current approach ADAL would have hard time updating service and every client.

tests/Functional/src/com/microsoft/aad/adal/test/AuthenticationContextTest.java
```diff
@@ -1364,6 +1367,171 @@ public void testRefreshTokenRequestWithFamilyIdFailed() throws InterruptedExcept
...
+        webrequest.setReturnResponse(new HttpWebResponse(400, responseBody.getBytes(), null));
+        ReflectionUtils.setFieldValue(context, "mWebRequest", webrequest);
+        
+        try {
+            context.acquireTokenSilentSync("resource", "clientid", TEST_IDTOKEN_USERID);
+            // Exception should be thrown for acquireTokenSilentSync, should never reach the fail check.
+            fail();
+        } catch (AuthenticationException e) {
+            assertEquals("Token is not exchanged",
+                    ADALError.AUTH_REFRESH_FAILED_PROMPT_NOT_ALLOWED, e.getCode());
```
imakhalova: verify that this is because of server response, not server error

tests/Functional/src/com/microsoft/aad/adal/test/AuthenticationContextTest.java
```diff
@@ -983,15 +984,17 @@ public void testRefreshTokenWebRequestHasError() throws InterruptedException,
...
-        response.listenLogForMessageSegments(signal, "Refresh token did not return accesstoken");
 
         context.acquireTokenSilentAsync("resource", "clientid", TEST_IDTOKEN_USERID, callback);
 
         signal.await(CONTEXT_REQUEST_TIME_OUT, TimeUnit.MILLISECONDS);
 
+        assertNotNull(callback.mException);
+        assertTrue(callback.mException instanceof AuthenticationException);
+        AuthenticationException authenticationException = (AuthenticationException)callback.mException;
+        assertEquals(authenticationException.getCode(), ADALError.AUTH_REFRESH_FAILED_PROMPT_NOT_ALLOWED);
```
imakhalova: verify that this is because of server error, not because invalid grant

src/src/com/microsoft/aad/adal/HttpResponseException.java
```diff
@@ -0,0 +1,51 @@
...
+    /**
+     * Default serialized version
+     */
+    private static final long serialVersionUID = 1L;
+    
+    private String mErrorCode;
+    
+    protected HttpResponseException(final String errorCode) {
+        super();
+        this.mErrorCode = errorCode;
```
imakhalova: httpStatusCode

src/src/com/microsoft/aad/adal/AuthenticationContext.java
```diff
@@ -1461,8 +1470,14 @@ private AuthenticationResult localFlow(CallbackHandler callbackHandle,
                 // User does not want to launch activity
                 Logger.e(TAG, "Prompt is not allowed and failed to get token:", request.getLogInfo() + " " + errorInfo,
                         ADALError.AUTH_REFRESH_FAILED_PROMPT_NOT_ALLOWED);
-                callbackHandle.onError(new AuthenticationException(
-                        ADALError.AUTH_REFRESH_FAILED_PROMPT_NOT_ALLOWED, request.getLogInfo() + " " + errorInfo));
+                if (authResult != null && authResult.getErrorCode() != null) {
+                    callbackHandle.onError(new AuthenticationException(ADALError.AUTH_REFRESH_FAILED_PROMPT_NOT_ALLOWED,
```
imakhalova: So as a result we actually should sign out users only if we've got internal cause and message of this internal exception contains invalid_grant...
ok, I'm glad that we know it, but I'm sorry for other users of this lib...how the hell they should know such thing!

src/src/com/microsoft/aad/adal/AuthenticationContext.java
```diff
@@ -1865,14 +1881,21 @@ private AuthenticationResult getTokenWithRefreshToken(final IWindowComponent act
...
+                Logger.w(TAG, "Received empty response from refresh token request.", request.getLogInfo(), 
+                        ADALError.AUTH_FAILED_NO_TOKEN);
+                return result;
+            } else if (StringExtensions.IsNullOrBlank(result.getAccessToken())) {
+                final String errorLogInfo = result.getErrorLogInfo();
+                Logger.e(TAG, "Refresh token request failed to return accesstoken.", 
+                        request.getLogInfo() + errorLogInfo, ADALError.AUTH_FAILED_NO_TOKEN);
+                
+                // check error code, only remove token from cache if receive invalid_grant from server
+                if (AuthenticationConstants.OAuth2ErrorCode.INVALID_GRANT.equalsIgnoreCase(result.getErrorCode())) {
```
imakhalova: Does it mean that even if we have not INVALID_GRANT it will return please sign out user response?

src/src/com/microsoft/aad/adal/AuthenticationContext.java
imakhalova: And overall I'm just confused that we want to return AUTH_REFRESH_FAILED_PROMPT_NOT_ALLOWED always? what is the point to have other error codes?

src/src/com/microsoft/aad/adal/AuthenticationConstants.java
```diff
@@ -149,6 +149,13 @@
         
         static final String ADAL_CLIENT_FAMILY_ID = "foci";
     }
+    
+    public static final class OAuth2ErrorCode {
+        /**
+         * Oauth2 error code invalid_grant. 
+         */
+        static final String INVALID_GRANT = "invalid_grant";
```
agamemnum: If the plan to make us depend on this. This needs to be public

src/src/com/microsoft/aad/adal/AuthenticationContext.java
imakhalova: And I think AUTH_FAILED_NO_TOKEN more appropriate code, since it returned for any other exception that happened during token refresh

src/src/com/microsoft/aad/adal/AuthenticationContext.java
```diff
@@ -1440,6 +1440,15 @@ private AuthenticationResult localFlow(CallbackHandler callbackHandle,
...
+                // Try interactive flow for non-silent request.
+                if (request.isSilent()) {
+                    final AuthenticationException authenticationException = new AuthenticationException(ADALError.AUTH_REFRESH_FAILED_PROMPT_NOT_ALLOWED, 
+                            httpResponseException.getMessage() + request.getLogInfo());
+                    callbackHandle.onError(authenticationException);
+                    return null;
+                }
             }
             
             if (authResult != null && !StringExtensions.IsNullOrBlank(authResult.getAccessToken())) {
```
imakhalova: Can authResult be null here? what will happen if it's null?

sdabbiru: see below.

src/src/com/microsoft/aad/adal/AuthenticationContext.java
weijjia: We return the throwable containing the detailed oauth error code back.

src/src/com/microsoft/aad/adal/AuthenticationContext.java
weijjia: I understand what you're saying, but then it's a different behavior compared to what we have today.

src/src/com/microsoft/aad/adal/Oauth2.java
imakhalova: With current implementation we're swallowing this exception, because we processTokenResponse even if we've got IOException

src/src/com/microsoft/aad/adal/Oauth2.java
weijjia: It's possible, if server returns the JSON response i.e. with only error description not error code, the current code actually won't parse the response if that's the case, the empty result will be returned.

src/src/com/microsoft/aad/adal/AuthenticationContext.java
weijjia: AUTH_REFRESH_FAILE_PROMPT_NOT_ALLOWED is a general error code we currently return back to user indicating "silent request failed, you need to start interactive flow". 

The thing that's hard for us is, the above error code is what we return today, and an agreement we have with our consumers that it basically means you need interactive flow. Changing the error code will break the existing agreement we have.

src/src/com/microsoft/aad/adal/AuthenticationContext.java
weijjia: I do agree with you. I plan to do some refactor around this logic, and will send separate PR around it. Will let you know when I do it, and feel free to share thoughts.

tests/Functional/src/com/microsoft/aad/adal/test/AuthenticationContextTest.java
weijjia: good catch. I'll cover it in different PR.

src/src/com/microsoft/aad/adal/AuthenticationContext.java
weijjia: The design here is, we parse the response from server, and if server returns us detailed error code, we propagated back to user.

src/src/com/microsoft/aad/adal/AuthenticationContext.java
imakhalova: Then this change is also changing behavior, before this change for IOexceptions (occurring in postMessage) in response we were returning AUTH_FAILED_NO_TOKEN , now we return AUTH_REFRESH_FAILED_PROMPT_NOT_ALLOWED, because we will throw HttpResponseException.
So why this is allowed and another way around is not?

src/src/com/microsoft/aad/adal/AuthenticationContext.java
```diff
@@ -1440,6 +1440,15 @@ private AuthenticationResult localFlow(CallbackHandler callbackHandle,
             } catch (AuthenticationException authenticationException) {
                 callbackHandle.onError(authenticationException);
                 return null;
+            } catch (HttpResponseException httpResponseException) {
```
imakhalova: I don't see the point of creating this exception...you was able to get all needed info from result object

src/src/com/microsoft/aad/adal/Oauth2.java
imakhalova: Then throw a HttpResponseException here and never worry about checking if it's null

src/src/com/microsoft/aad/adal/AuthenticationContext.java
```diff
@@ -1440,6 +1440,15 @@ private AuthenticationResult localFlow(CallbackHandler callbackHandle,
...
+                // Return back the http error for silent request
+                // Try interactive flow for non-silent request.
+                if (request.isSilent()) {
+                    final AuthenticationException authenticationException = new AuthenticationException(ADALError.AUTH_REFRESH_FAILED_PROMPT_NOT_ALLOWED, 
+                            httpResponseException.getMessage() + request.getLogInfo());
+                    callbackHandle.onError(authenticationException);
```
imakhalova: derive HttpResponseException from AuthenticationException  and return it here in callback, you shouldn't generate new exception and loose stack trace

PR #591, opened 2016-04-19
set error code on AuthenticationResult for 400 and 401, remove refres…

weijjia:
1. set server returned error code on AuthenticationResult also for 400 and 401
2. remove refresh token only when receiving invalid_grant
3. add tests

### DIFF
--- a/src/src/com/microsoft/aad/adal/AuthenticationConstants.java
+++ b/src/src/com/microsoft/aad/adal/AuthenticationConstants.java
@@ -349,4 +349,12 @@ public class AuthenticationConstants {
     public static final int DEFAULT_EXPIRATION_TIME_SEC = 3600;
 
     public static final String AUTHENTICATION_FILE_DIRECTORY = "com.microsoft.aad.adal.authentication";
+    
+    protected static final class OAuth2ErrorCode {
+        /**       
+         * Oauth2 error code invalid_grant.
+         */       
+        static final String INVALID_GRANT = "invalid_grant";
+    }     
+
 }
--- a/src/src/com/microsoft/aad/adal/Oauth2.java
+++ b/src/src/com/microsoft/aad/adal/Oauth2.java
@@ -578,7 +578,7 @@ class Oauth2 {
      * @return
      */
     private AuthenticationResult processTokenResponse(HttpWebResponse webResponse) {
-        AuthenticationResult result = new AuthenticationResult();
+        AuthenticationResult result;
         String correlationIdInHeader = null;
         if (webResponse.getResponseHeaders() != null
                 && webResponse.getResponseHeaders().containsKey(
--- a/src/src/com/microsoft/aad/adal/Oauth2.java
+++ b/src/src/com/microsoft/aad/adal/Oauth2.java
@@ -579,7 +579,6 @@ class Oauth2 {
      */
     private AuthenticationResult processTokenResponse(HttpWebResponse webResponse) {
         AuthenticationResult result = new AuthenticationResult();
-        HashMap<String, String> responseItems = new HashMap<String, String>();
         String correlationIdInHeader = null;
         if (webResponse.getResponseHeaders() != null
                 && webResponse.getResponseHeaders().containsKey(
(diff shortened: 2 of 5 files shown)


kpanwar:
:shipit:

tests/Functional/src/com/microsoft/aad/adal/test/AuthenticationContextTest.java
```diff
@@ -2174,6 +2171,135 @@ public void testVerifyBrokerRedirectUri_invalidSignature() throws NoSuchAlgorith
...
+    @SmallTest     
+    public void testRefreshTokenWithInvalidGrantReturned_CacheCleared() throws NoSuchFieldException, IllegalAccessException, InterruptedException {
+        FileMockContext mockContext = new FileMockContext(getContext());
+        ITokenCacheStore mockCache = getCacheForRefreshToken(TEST_IDTOKEN_USERID, TEST_IDTOKEN_UPN);
+        final AuthenticationContext context = getAuthenticationContext(mockContext,
+                VALID_AUTHORITY, false, mockCache);
+        setConnectionAvailable(context, true); 
+        final MockActivity testActivity = new MockActivity();
+        final CountDownLatch signal = new CountDownLatch(1);
+        testActivity.mSignal = signal;
```
imakhalova: I don't think that you need activity and signal for silentsync test

weijjia: you're right, i'll have it removed.

tests/Functional/src/com/microsoft/aad/adal/test/AuthenticationContextTest.java
```diff
@@ -2174,6 +2171,135 @@ public void testVerifyBrokerRedirectUri_invalidSignature() throws NoSuchAlgorith
...
+            // Exception should be thrown for acquireTokenSilentSync, should never reach the fail check.
+            fail();
+        } catch (AuthenticationException e) {
+            assertEquals("Token is not exchanged",
+                    ADALError.AUTH_FAILED_NO_TOKEN, e.getCode());
+            // If AUTH_FAILE_NO_TOKEN is thrown, cause will be return in AuthenticationException
+            assertNotNull(e.getCause());
+        }     
+
+        // verify that the cache is cleared
```
imakhalova: not cleared?

tests/Functional/src/com/microsoft/aad/adal/test/AuthenticationContextTest.java
```diff
@@ -2174,6 +2171,135 @@ public void testVerifyBrokerRedirectUri_invalidSignature() throws NoSuchAlgorith
...
+     public void testRefreshTokenWithInteractionRequired_CacheNotCleared() throws NoSuchFieldException, IllegalAccessException, InterruptedException {
+         FileMockContext mockContext = new FileMockContext(getContext());
+         ITokenCacheStore mockCache = getCacheForRefreshToken(TEST_IDTOKEN_USERID, TEST_IDTOKEN_UPN);
+         final AuthenticationContext context = getAuthenticationContext(mockContext,
+                 VALID_AUTHORITY, false, mockCache);
+         setConnectionAvailable(context, true);
+         final MockActivity testActivity = new MockActivity();
+         final CountDownLatch signal = new CountDownLatch(1);
+         testActivity.mSignal = signal;
+         MockWebRequestHandler webrequest = new MockWebRequestHandler();
```
imakhalova: the same code is in multiple tests... may be better have a helper method that provides only response body?

tests/Functional/src/com/microsoft/aad/adal/test/OauthTests.java
```diff
@@ -62,6 +62,7 @@
 import com.microsoft.aad.adal.PromptBehavior;
 
 import android.annotation.SuppressLint;
+import android.test.AndroidTestCase;
```
imakhalova: revert?

tests/Functional/src/com/microsoft/aad/adal/test/AuthenticationContextTest.java
weijjia: ah, right. Thanks.

src/src/com/microsoft/aad/adal/AuthenticationConstants.java
```diff
@@ -349,4 +349,12 @@
     public static final int DEFAULT_EXPIRATION_TIME_SEC = 3600;
 
     public static final String AUTHENTICATION_FILE_DIRECTORY = "com.microsoft.aad.adal.authentication";
+    
+    protected static final class OAuth2ErrorCode {
```
imakhalova: I'm not sure why do you need this internal class and why is it protected?

imakhalova:
just some minor comments, looks good to me, signed off

src/src/com/microsoft/aad/adal/AuthenticationConstants.java
weijjia: my thought is, in case we are adding more error code in the future, group oauth error code constants to the internal class. I don't want the class to be visible outside which is only for internal use.

tests/Functional/src/com/microsoft/aad/adal/test/OauthTests.java
weijjia: nope, i cannot. I'm seeing whole bunch of error shown in eclipse when reverting it.

src/src/com/microsoft/aad/adal/Oauth2.java
```diff
@@ -579,7 +579,6 @@ public String encodeProtocolState() {
      */
     private AuthenticationResult processTokenResponse(HttpWebResponse webResponse) {
         AuthenticationResult result = new AuthenticationResult();
```
agamemnum: nit: consider having it only declared, but not initialized. This way you'll ensure that it's set correctly below and you aren't missing a case

src/src/com/microsoft/aad/adal/AuthenticationConstants.java
imakhalova: If you want to make this class visible within package only use package visibility, not protected

src/src/com/microsoft/aad/adal/Oauth2.java
```diff
@@ -578,8 +578,7 @@ public String encodeProtocolState() {
      * @return
      */
     private AuthenticationResult processTokenResponse(HttpWebResponse webResponse) {
-        AuthenticationResult result = new AuthenticationResult();
-        HashMap<String, String> responseItems = new HashMap<String, String>();
+        AuthenticationResult result;
```
imakhalova: nit: final?

weijjia: it cannot be, it will be overridden.

src/src/com/microsoft/aad/adal/AuthenticationConstants.java
weijjia: the only difference between protected and package visible is subclass visibility. But I agree though, this class is all about constants, cannot be subclassed anyways.

PR #592, opened 2016-04-19
Minor refactoring of RefreshItem, HttpWebResponse and HttpWebRequest  classes

imakhalova:
There is no functionality changes, but I think it adds readability to code:
make private variable final when it initialized in constructor and won't be changed
remove extra setters and getters
separates function that refreshes token and updates cache
extract logic that sets up http connection
throw IO exception in send method instead of storing it in response and throwing later (as practice shows this type logic can cause bugs by forgetting to throw this exception)
fix leaks in unclosed streams

### DIFF
--- a/src/src/com/microsoft/aad/adal/AuthenticationContext.java
+++ b/src/src/com/microsoft/aad/adal/AuthenticationContext.java
@@ -1594,11 +1594,10 @@ public class AuthenticationContext {
             mUserInfo = item.getUserInfo();
             mRawIdToken = item.getRawIdToken();
             mTenantId = item.getTenantId();
-            if (item.getUserInfo() != null) {
-                mKeyWithUserId = CacheKey.createCacheKey(request, item.getUserInfo()
-                        .getUserId());
-                mKeyWithDisplayableId = CacheKey.createCacheKey(request, item.getUserInfo()
-                        .getDisplayableId());
+            final UserInfo userInfo = item.getUserInfo();
+            if (userInfo != null) {
+                mKeyWithUserId = CacheKey.createCacheKey(request, userInfo.getUserId());
+                mKeyWithDisplayableId = CacheKey.createCacheKey(request, userInfo.getDisplayableId());
             }
         }
     }
@@ -1809,7 +1808,7 @@ public class AuthenticationContext {
     private AuthenticationResult getTokenWithRefreshTokenAndUpdateCache(
             final AuthenticationRequest request, final RefreshItem refreshItem)
             throws AuthenticationException {
-        AuthenticationResult result = getTokenWithRefreshToken(request, refreshItem.mRefreshToken);
+        final AuthenticationResult result = getTokenWithRefreshToken(request, refreshItem.mRefreshToken);
 
         if (result == null || StringExtensions.IsNullOrBlank(result.getAccessToken())) {
             String errLogInfo = result == null ? "" : result.getErrorLogInfo();
@@ -1817,16 +1816,15 @@ public class AuthenticationContext {
                     + errLogInfo, ADALError.AUTH_FAILED_NO_TOKEN);
 
             // Check error code, only remove token from cache if receiving invalid_grant from server
-            if (AuthenticationConstants.OAuth2ErrorCode.INVALID_GRANT.equals(result.getErrorCode())) {
+            if (result != null &&
+                    AuthenticationConstants.OAuth2ErrorCode.INVALID_GRANT.equalsIgnoreCase(result.getErrorCode())) {
                 Logger.v(TAG, "Removing token cache for invalid_grant error returned from server.");
                 // remove item from cache to avoid same usage of
                 // refresh token in next acquireToken call
                 removeItemFromCache(refreshItem);
             }
 
-            return result;
         } else {
-            Logger.v(TAG, "It finished refresh token request:" + request.getLogInfo());
             if (result.getUserInfo() == null && refreshItem.mUserInfo != null) {
                 Logger.v(TAG, "UserInfo is updated from cached result:" + request.getLogInfo());
                 result.setUserInfo(refreshItem.mUserInfo);
@@ -1837,14 +1835,11 @@ public class AuthenticationContext {
             // it replaces multi resource refresh token as
             // well with the new one since it is not stored
             // with resource.
-            Logger.v(TAG, "Cache is used. It will set item to cache" + request.getLogInfo());
+            Logger.v(TAG, "Saving item into the cache." + request.getLogInfo());
             setItemToCacheFromRefresh(refreshItem, request, result);
-
-            // return result obj which has error code and
-            // error description that is returned from
-            // server response
-            return result;
         }
+
+        return result;
     }
 
     /**
@@ -1874,7 +1869,7 @@ public class AuthenticationContext {
             throw authenticationException;
         }
 
-        AuthenticationResult result = null;
+        final AuthenticationResult result;
         try {
             Oauth2 oauthRequest = new Oauth2(request, mWebRequest, mJWSBuilder);
             result = oauthRequest.refreshToken(refreshToken);
--- a/src/src/com/microsoft/aad/adal/AuthenticationContext.java
+++ b/src/src/com/microsoft/aad/adal/AuthenticationContext.java
@@ -1436,7 +1436,7 @@ public class AuthenticationContext {
                 && !StringExtensions.IsNullOrBlank(refreshItem.mRefreshToken)) {
             Logger.v(TAG, "Refresh token is available and it will attempt to refresh token");
             try {
-                authResult = getTokenWithRefreshToken(activity, useDialog, request, refreshItem, true);
+                authResult = getTokenWithRefreshTokenAndUpdateCache(request, refreshItem);
             } catch (AuthenticationException authenticationException) {
                 callbackHandle.onError(authenticationException);
                 return null;
@@ -1573,45 +1573,33 @@ public class AuthenticationContext {
      * another call to acquireToken. It may try multi resource refresh token for
      * second attempt.
      */
-    private class RefreshItem {
-        String mRefreshToken;
-
-        String mKey;
-
-        boolean mMultiResource;
-
-        UserInfo mUserInfo;
-
-        String mRawIdToken;
-
+    private static class RefreshItem {
+        final String mRefreshToken;
+        final String mKey;
+        final boolean mMultiResource;
+        final UserInfo mUserInfo;
+        final String mRawIdToken;
+        final String mTenantId;
         String mKeyWithUserId;
-
         String mKeyWithDisplayableId;
-        
-        String mTenantId;
 
         public RefreshItem(String keyInCache, AuthenticationRequest request, TokenCacheItem item,
                 boolean multiResource) {
+            if (item == null) {
+                throw new IllegalArgumentException("item");
+            }
             mKey = keyInCache;
             mMultiResource = multiResource;
-
-            if (item != null) {
-                mRefreshToken = item.getRefreshToken();
-                mUserInfo = item.getUserInfo();
-                mRawIdToken = item.getRawIdToken();
-                mTenantId = item.getTenantId();
-                if (item.getUserInfo() != null) {
-                    mKeyWithUserId = CacheKey.createCacheKey(request, item.getUserInfo()
-                            .getUserId());
-                    mKeyWithDisplayableId = CacheKey.createCacheKey(request, item.getUserInfo()
-                            .getDisplayableId());
-                }
+            mRefreshToken = item.getRefreshToken();
+            mUserInfo = item.getUserInfo();
+            mRawIdToken = item.getRawIdToken();
+            mTenantId = item.getTenantId();
+            if (item.getUserInfo() != null) {
+                mKeyWithUserId = CacheKey.createCacheKey(request, item.getUserInfo()
+                        .getUserId());
+                mKeyWithDisplayableId = CacheKey.createCacheKey(request, item.getUserInfo()
+                        .getDisplayableId());
             }
-        }
-
-        public RefreshItem(String refreshToken) {
-            mMultiResource = false;
-            mRefreshToken = refreshToken;
         }
     }
 
@@ -1810,23 +1798,66 @@ public class AuthenticationContext {
     }
 
     /**
-     * refresh token if possible. if it fails, it calls acquire token after
-     * removing refresh token from cache.
-     * 
-     * @param activity Activity to use in case refresh token does not succeed
-     *            and prompt is not set to never.
+     * refresh token if possible and update cache with new token value
+     *
      * @param request incoming request
      * @param refreshItem refresh item info to remove this refresh token from
      *            cache
-     * @param useCache refresh request can be explicit without cache usage.
-     *            Error message should return without trying prompt.
      * @return
      */
-    private AuthenticationResult getTokenWithRefreshToken(final IWindowComponent activity, final boolean useDialog,
-            final AuthenticationRequest request, final RefreshItem refreshItem, final boolean useCache) 
+    private AuthenticationResult getTokenWithRefreshTokenAndUpdateCache(
+            final AuthenticationRequest request, final RefreshItem refreshItem)
+            throws AuthenticationException {
+        AuthenticationResult result = getTokenWithRefreshToken(request, refreshItem.mRefreshToken);
+
+        if (result == null || StringExtensions.IsNullOrBlank(result.getAccessToken())) {
+            String errLogInfo = result == null ? "" : result.getErrorLogInfo();
+            Logger.e(TAG, "Refresh token did not return accesstoken.", request.getLogInfo()
+                    + errLogInfo, ADALError.AUTH_FAILED_NO_TOKEN);
+
+            // Check error code, only remove token from cache if receiving invalid_grant from server
+            if (AuthenticationConstants.OAuth2ErrorCode.INVALID_GRANT.equals(result.getErrorCode())) {
+                Logger.v(TAG, "Removing token cache for invalid_grant error returned from server.");
+                // remove item from cache to avoid same usage of
+                // refresh token in next acquireToken call
+                removeItemFromCache(refreshItem);
+            }
+
+            return result;
+        } else {
+            Logger.v(TAG, "It finished refresh token request:" + request.getLogInfo());
+            if (result.getUserInfo() == null && refreshItem.mUserInfo != null) {
+                Logger.v(TAG, "UserInfo is updated from cached result:" + request.getLogInfo());
+                result.setUserInfo(refreshItem.mUserInfo);
+                result.setIdToken(refreshItem.mRawIdToken);
+                result.setTenantId(refreshItem.mTenantId);
+            }
+
+            // it replaces multi resource refresh token as
+            // well with the new one since it is not stored
+            // with resource.
+            Logger.v(TAG, "Cache is used. It will set item to cache" + request.getLogInfo());
+            setItemToCacheFromRefresh(refreshItem, request, result);
+
+            // return result obj which has error code and
+            // error description that is returned from
+            // server response
+            return result;
+        }
+    }
+
+    /**
+     * refresh token if possible.
+     * 
+     * @param request incoming request
+     * @param refreshToken refresh token
+     * @return
+     */
+    private AuthenticationResult getTokenWithRefreshToken(
+            final AuthenticationRequest request, final String refreshToken)
             throws AuthenticationException {
         Logger.v(TAG, "Process refreshToken for " + request.getLogInfo() + " refreshTokenId:"
-                + getTokenHash(refreshItem.mRefreshToken));
+                + getTokenHash(refreshToken));
 
         // Removes refresh token from cache, when this call is complete. Request
         // may be interrupted, if app is shutdown by user. Detect connection
@@ -1845,10 +1876,10 @@ public class AuthenticationContext {
         AuthenticationResult result = null;
         try {
             Oauth2 oauthRequest = new Oauth2(request, mWebRequest, mJWSBuilder);
-            result = oauthRequest.refreshToken(refreshItem.mRefreshToken);
+            result = oauthRequest.refreshToken(refreshToken);
             if (result != null && StringExtensions.IsNullOrBlank(result.getRefreshToken())) {
                 Logger.v(TAG, "Refresh token is not returned or empty");
-                result.setRefreshToken(refreshItem.mRefreshToken);
+                result.setRefreshToken(refreshToken);
             }
         } catch (IOException | AuthenticationException exc) {
             // Server side error or similar
@@ -1862,49 +1893,7 @@ public class AuthenticationContext {
             throw authException;
         }
 
-        // useCache will be false when calling from acquireTokenUsingRefreshToken, and if false, need to return
-        // the error through callback. If true, just return the result back to localflow. 
-        if (useCache) {
-            if (result == null || StringExtensions.IsNullOrBlank(result.getAccessToken())) {
-                String errLogInfo = result == null ? "" : result.getErrorLogInfo();
-                Logger.e(TAG, "Refresh token did not return accesstoken.", request.getLogInfo()
-                        + errLogInfo, ADALError.AUTH_FAILED_NO_TOKEN);
-
-                // Check error code, only remove token from cache if receiving invalid_grant from server
-                if (AuthenticationConstants.OAuth2ErrorCode.INVALID_GRANT.equals(result.getErrorCode())) {
-                    Logger.v(TAG, "Removing token cache for invalid_grant error returned from server.");
-                    removeItemFromCache(refreshItem);
-                }
-
-                return result;
-            } else {
-                Logger.v(TAG, "It finished refresh token request:" + request.getLogInfo());
-                if (result.getUserInfo() == null && refreshItem.mUserInfo != null) {
-                    Logger.v(TAG, "UserInfo is updated from cached result:" + request.getLogInfo());
-                    result.setUserInfo(refreshItem.mUserInfo);
-                    result.setIdToken(refreshItem.mRawIdToken);
-                    result.setTenantId(refreshItem.mTenantId);
-                }
-
-                // it replaces multi resource refresh token as
-                // well with the new one since it is not stored
-                // with resource.
-                Logger.v(TAG, "Cache is used. It will set item to cache" + request.getLogInfo());
-                setItemToCacheFromRefresh(refreshItem, request, result);
-
-                // return result obj which has error code and
-                // error description that is returned from
-                // server response
-                return result;
-            }
-        } else {
-            // User is not using cache and explicitly
-            // calling with refresh token. User should received
-            // error code and error description in
-            // Authentication result for Oauth errors
-            Logger.v(TAG, "Cache is not used for Request:" + request.getLogInfo());
-            return result;
-        }
+        return result;
     }
 
     private boolean validateAuthority(final URL authorityUrl) {
@@ -2061,8 +2050,6 @@ public class AuthenticationContext {
                 // It is not using cache and refresh is not expected to
                 // show authentication activity.
                 request.setSilent(true);
-                final RefreshItem refreshItem = new RefreshItem(refreshToken);
-
                 if (mValidateAuthority) {
                     Logger.v(TAG, "Validating authority");
 
@@ -2084,7 +2071,7 @@ public class AuthenticationContext {
                 final AuthenticationResult authResult;
                 try
                 {
-                    authResult = getTokenWithRefreshToken(null, false, request, refreshItem, false);
+                    authResult = getTokenWithRefreshToken(request, refreshToken);
                 }
                 catch (final AuthenticationException authException)
                 {
--- a/src/src/com/microsoft/aad/adal/AuthenticationParameters.java
+++ b/src/src/com/microsoft/aad/adal/AuthenticationParameters.java
@@ -23,6 +23,7 @@
 
 package com.microsoft.aad.adal;
 
+import java.io.IOException;
 import java.io.UnsupportedEncodingException;
 import java.net.HttpURLConnection;
 import java.net.URL;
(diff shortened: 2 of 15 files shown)


src/src/com/microsoft/aad/adal/HttpWebRequest.java
```diff
@@ -107,107 +89,103 @@ private void setupConnection() {
...
-    public HttpWebResponse send() {
-
-        Logger.v(TAG, "HttpWebRequest send thread:" + android.os.Process.myTid());
-        setupConnection();
-        HttpWebResponse response = new HttpWebResponse();
+    public HttpWebResponse send() throws IOException {
+        Logger.v(TAG, "HttpWebRequest send thread:" + Process.myTid());
+        final HttpURLConnection connection = setupConnection();
+        if (connection == null) {
+            return null;
```
agamemnum: Hm. why this is not a throw IOException?

src/src/com/microsoft/aad/adal/HttpWebRequest.java
```diff
@@ -107,107 +89,103 @@ private void setupConnection() {
...
-                    Logger.v(TAG, "Setting header: " + header);
-                    mConnection.setRequestProperty(header, mRequestHeaders.get(header));
+                responseStream = connection.getInputStream();
+            } catch (IOException ex) {
+                Logger.e(TAG, "IOException:" + ex.getMessage(), "", ADALError.SERVER_ERROR);
+                // If it does not get the error stream, it will return
+                // exception in the httpresponse
+                responseStream = connection.getErrorStream();
+                if (responseStream == null) {
+                    throw ex;
```
agamemnum: nit: just `throw`

src/src/com/microsoft/aad/adal/HttpWebRequest.java
```diff
@@ -107,107 +89,103 @@ private void setupConnection() {
...
-                try {
-                    responseStream = mConnection.getInputStream();
-                } catch (IOException ex) {
-                    Logger.e(TAG, "IOException:" + ex.getMessage(), "", ADALError.SERVER_ERROR);
-                    // If it does not get the error stream, it will return
-                    // exception in the httpresponse
-                    responseStream = mConnection.getErrorStream();
-                    mException = ex;
+                // Continue to read from stream if not cancelled and not EOF
+                while ((bytesRead = responseStream.read(buffer)) > 0) {
```
agamemnum: it looks like `responseStream` is leaked

src/src/com/microsoft/aad/adal/HttpWebRequest.java
```diff
@@ -107,107 +89,103 @@ private void setupConnection() {
...
-                setRequestBody();
+            // GET request should read status after getInputStream to make
+            // this work for different SDKs
+            final int statusCode = getStatusCode(connection);
+            byte[] responseBody = null;
+            if (responseStream != null) {
 
-                byte[] responseBody = null;
-                InputStream responseStream = null;
+                ByteArrayOutputStream byteStream = new ByteArrayOutputStream();
```
agamemnum: Can we have a helper method for this? Copy it from our code base.

src/src/com/microsoft/aad/adal/HttpWebRequest.java
```diff
@@ -107,107 +89,103 @@ private void setupConnection() {
...
-                mException = e;
-            } finally {
-                mConnection.disconnect();
-                mConnection = null;
             }
+
+            Logger.v(TAG, "Response is received");
+            response = new HttpWebResponse(statusCode, responseBody, connection.getHeaderFields());
+        } finally {
+            connection.disconnect();
```
agamemnum: Hm. I think I know why we setting connection reuse headers :-) Documentation says to avoid calling this method as it may kill other connections. I don't think we should change this right now, but we should consider removing `System.setProperty("http.keepAlive", "false");` as well as `connection.disconnect();`

src/src/com/microsoft/aad/adal/HttpWebRequest.java
```diff
@@ -107,107 +89,103 @@ private void setupConnection() {
...
         }
 
-        response.setResponseException(mException);
         return response;
     }
 
-    private void getStatusCode(HttpWebResponse response) throws IOException {
+    private int getStatusCode(HttpURLConnection connection) throws IOException {
```
agamemnum: static now?

src/src/com/microsoft/aad/adal/HttpWebRequest.java
```diff
@@ -107,107 +89,103 @@ private void setupConnection() {
...
         int statusCode = HttpURLConnection.HTTP_BAD_REQUEST;
 
         try {
-            statusCode = mConnection.getResponseCode();
+            statusCode = connection.getResponseCode();
         } catch (IOException ex) {
 
             if (Build.VERSION.SDK_INT < 16) {
```
agamemnum: Hm. Does it mean that we call this function before we actually called getInputStream?

src/src/com/microsoft/aad/adal/HttpWebRequest.java
```diff
@@ -233,42 +211,23 @@ private void getStatusCode(HttpWebResponse response) throws IOException {
...
-            mConnection.setRequestProperty("Content-Length",
+            connection.setRequestProperty("Content-Length",
                     Integer.toString(mRequestContent.length));
-            mConnection.setFixedLengthStreamingMode(mRequestContent.length);
+            connection.setFixedLengthStreamingMode(mRequestContent.length);
 
-            OutputStream out = mConnection.getOutputStream();
+            OutputStream out = connection.getOutputStream();
             out.write(mRequestContent);
             out.close();
```
agamemnum: finally please

src/src/com/microsoft/aad/adal/HttpWebRequest.java
imakhalova: Actually it will never be null, because initialization will throw IOException and we won't reach this point. Let's remove this check...

src/src/com/microsoft/aad/adal/HttpWebRequest.java
imakhalova: Nope, right after...but parsing what in input streams comes after status code check

src/src/com/microsoft/aad/adal/HttpWebRequest.java
imakhalova: Nope, we should check error stream which contains data why refresh token failed...

src/src/com/microsoft/aad/adal/Discovery.java
```diff
@@ -189,13 +190,13 @@ private boolean sendRequest(final URL queryUrl) throws MalformedURLException, JS
...
         try {
             ClientMetrics.INSTANCE.beginClientMetricsRecord(queryUrl, mCorrelationId, headers);
-            webResponse = mWebrequestHandler.sendGet(queryUrl, headers);
-            if (webResponse.getResponseException() == null) {
+            try {
+                webResponse = mWebrequestHandler.sendGet(queryUrl, headers);
                 ClientMetrics.INSTANCE.setLastError(null);
-            } else {
+            } catch (IOException e) {
                 ClientMetrics.INSTANCE.setLastError(String.valueOf(webResponse.getStatusCode()));
```
imakhalova: I don't understand this ClientMetrics usage, so I'm not sure why we swallowing this exception...

src/src/com/microsoft/aad/adal/Oauth2.java
```diff
@@ -598,16 +593,13 @@ private AuthenticationResult processTokenResponse(HttpWebResponse webResponse) {
         case HttpURLConnection.HTTP_UNAUTHORIZED:
             try {
                 result = parseJsonResponse(webResponse.getBody());
-            } catch (final JSONException jsonException) {
-                Logger.e(TAG, jsonException.getMessage(), "", ADALError.SERVER_INVALID_JSON_RESPONSE, jsonException);
-                result = new AuthenticationResult(JSON_PARSING_ERROR, jsonException.getMessage(), null);
+            } catch (JSONException e) {
+                throw new AuthenticationException(ADALError.SERVER_INVALID_JSON_RESPONSE, "Can't parse server response " + webResponse.getBody(), e);
```
imakhalova: I actually revert changes in this function, since it might end up in discussion and I'll submit it in separate PR

src/src/com/microsoft/aad/adal/Oauth2.java
```diff
@@ -272,6 +273,8 @@ else if (response.containsKey(AuthenticationConstants.OAuth2.ACCESS_TOKEN)) {
             
             //Set family client id on authentication result for TokenCacheItem to pick up
             result.setFamilyClientId(familyClientId);
+        } else {
+            result = null;
```
imakhalova: This actually should throw an AuthenticationException, but I will submit it in next PR

src/src/com/microsoft/aad/adal/HttpWebRequest.java
agamemnum: Why are we doing this magic then?

src/src/com/microsoft/aad/adal/HttpWebRequest.java
```diff
@@ -233,44 +227,45 @@ private void getStatusCode(HttpWebResponse response) throws IOException {
...
+            }
+        }
+    }
+
+    /**
+     * Close the stream safely
+     *
+     * @param stream stream to be closed
+     */
+    private void safeCloseStream(Closeable stream) {
```
agamemnum: static

src/src/com/microsoft/aad/adal/HttpWebRequest.java
```diff
@@ -107,107 +92,116 @@ private void setupConnection() {
...
+        final Iterator<String> headerKeys = mRequestHeaders.keySet().iterator();
+
+        while (headerKeys.hasNext()) {
+            String header = headerKeys.next();
+            Logger.v(TAG, "Setting header: " + header);
+            connection.setRequestProperty(header, mRequestHeaders.get(header));
         }
+
+        // Avoid reuse of existing sockets to avoid random EOF errors
+        System.setProperty("http.keepAlive", "false");
```
agamemnum: Actually this will disabled connection reuse for every client that uses a library. For example for us it would have a significant impact on thumbnail downloading code...

src/src/com/microsoft/aad/adal/HttpWebRequest.java
imakhalova: According GitHub history this function was called before getInputStream for POST methods and after for GET. Since logic was changed, this magic can be removed, but I'll submit it in next PR.

src/src/com/microsoft/aad/adal/HttpWebRequest.java
imakhalova: Agree, will submit in next PR when we wrap up this one

src/src/com/microsoft/aad/adal/Discovery.java
imakhalova: Ok, now I understand ClientMetrics and we need to rethrow this exception after setting last error to ClientMetrics

src/src/com/microsoft/aad/adal/AuthenticationContext.java
```diff
@@ -1574,45 +1574,33 @@ private String getTokenHash(String token) {
...
+        final String mTenantId;
         String mKeyWithUserId;
-
         String mKeyWithDisplayableId;
-        
-        String mTenantId;
 
         public RefreshItem(String keyInCache, AuthenticationRequest request, TokenCacheItem item) {
+            if (item == null) {
+                throw new IllegalArgumentException("item");
```
weijjia: I don't think we should throw an IllegalArgumentException in our code, we should only throw it when validating input from calling app.

imakhalova: Ok, I don't mind removing this validation...

weijjia: I'm saying remove the validation, we can still check, but shouldn't throw IllegalArgumentException.

imakhalova: we shouldn't create this object if this argument is null, so I don't really get what do you mean by check and do not throw...

weijjia: All what I'm saying is, I don't think we should throw unchecked exception in our code, instead, we can throw checked exception.

weijjia: so we reach an agreement here that we can throw here as a defensive programming approach?

src/src/com/microsoft/aad/adal/AuthenticationContext.java
```diff
@@ -1811,23 +1799,66 @@ private void removeItemFromCache(final RefreshItem refreshItem) {
...
+    }
+
+    /**
+     * refresh token if possible.
+     * 
+     * @param request incoming request
+     * @param refreshToken refresh token
+     * @return
+     */
+    private AuthenticationResult getTokenWithRefreshToken(
```
weijjia: So you separate the original getTokenWithRefreshToken into two methods? one with cache update? one without? Then can you please rename this one? It's confusing to me. And also update the java doc.

weijjia: ok, never mind. I think I misunderstood here.

src/src/com/microsoft/aad/adal/AuthenticationContext.java
```diff
@@ -1811,23 +1799,66 @@ private void removeItemFromCache(final RefreshItem refreshItem) {
...
-     * @param useCache refresh request can be explicit without cache usage.
-     *            Error message should return without trying prompt.
      * @return
      */
-    private AuthenticationResult getTokenWithRefreshToken(final IWindowComponent activity, final boolean useDialog,
-            final AuthenticationRequest request, final RefreshItem refreshItem, final boolean useCache) 
+    private AuthenticationResult getTokenWithRefreshTokenAndUpdateCache(
+            final AuthenticationRequest request, final RefreshItem refreshItem)
+            throws AuthenticationException {
+        AuthenticationResult result = getTokenWithRefreshToken(request, refreshItem.mRefreshToken);
```
weijjia: nit:final

src/src/com/microsoft/aad/adal/AuthenticationParameters.java
```diff
@@ -166,17 +167,17 @@ public static void createFromResourceUrl(Context context, final URL resourceUrl,
             public void run() {
                 HashMap<String, String> headers = new HashMap<String, String>();
                 headers.put(WebRequestHandler.HEADER_ACCEPT, WebRequestHandler.HEADER_ACCEPT_JSON);
-                HttpWebResponse webResponse = sWebRequest.sendGet(resourceUrl, headers);
 
-                if (webResponse != null) {
-                    if(webResponse.getResponseException() != null) {
```
weijjia: I like the removal for checking getResponseException, it's just wrapping the IOException, have no idea why we did it that way :disappointed:

src/src/com/microsoft/aad/adal/AuthenticationParameters.java
```diff
@@ -166,17 +167,17 @@ public static void createFromResourceUrl(Context context, final URL resourceUrl,
...
-                        onCompleted(webResponse.getResponseException(), null);
-                    }
+                HttpWebResponse webResponse = null;
+                try {
+                    webResponse = sWebRequest.sendGet(resourceUrl, headers);
                     try {
                         onCompleted(null, parseResponse(webResponse));
                     } catch (ResourceAuthenticationChallengeException exc) {
```
weijjia: nit: final
Sorry, I always insist on making variables final if it's not going to be changed.

src/src/com/microsoft/aad/adal/Discovery.java
```diff
@@ -189,13 +192,14 @@ private boolean sendRequest(final URL queryUrl) throws MalformedURLException, JS
         String errorCodes = "";
         try {
             ClientMetrics.INSTANCE.beginClientMetricsRecord(queryUrl, mCorrelationId, headers);
-            webResponse = mWebrequestHandler.sendGet(queryUrl, headers);
-            if (webResponse.getResponseException() == null) {
+            try {
+                webResponse = mWebrequestHandler.sendGet(queryUrl, headers);
                 ClientMetrics.INSTANCE.setLastError(null);
-            } else {
+            } catch (IOException e) {
```
weijjia: nit: final

src/src/com/microsoft/aad/adal/Discovery.java
```diff
@@ -205,16 +209,7 @@ private boolean sendRequest(final URL queryUrl) throws MalformedURLException, JS
             }
             
             return (discoveryResponse != null && discoveryResponse.containsKey(TENANT_DISCOVERY_ENDPOINT));
-        } catch (IllegalArgumentException exc) {
```
weijjia: :disappointed: Thanks for cleaning this up. I don't understand why we did it this way, catch it and rethrow it, just for logging?!...

src/src/com/microsoft/aad/adal/HashMapExtensions.java
```diff
@@ -95,17 +96,13 @@
      * @throws JSONException
      */
     static final HashMap<String, String> getJsonResponse(HttpWebResponse webResponse) throws JSONException{
-        HashMap<String, String> response = new HashMap<String, String>();
-        if(webResponse != null && webResponse.getBody() != null && webResponse.getBody().length != 0){
-            JSONObject jsonObject = new JSONObject(
-                    new String(webResponse.getBody()));
-
+        HashMap<String, String> response = new HashMap<>();
```
weijjia: I know you don't touch this, but I'm just wondering why not returning the generic Map.

src/src/com/microsoft/aad/adal/HashMapExtensions.java
```diff
@@ -95,17 +96,13 @@
...
+        if(webResponse != null && !TextUtils.isEmpty(webResponse.getBody())) {
+            JSONObject jsonObject = new JSONObject(webResponse.getBody());
```
weijjia: why removing the creation of string from the byte array?

src/src/com/microsoft/aad/adal/HttpWebRequest.java
```diff
@@ -107,107 +92,116 @@ private void setupConnection() {
...
-
-        Logger.v(TAG, "HttpWebRequest send thread:" + android.os.Process.myTid());
-        setupConnection();
-        HttpWebResponse response = new HttpWebResponse();
-
-        if (mConnection != null) {
+    public HttpWebResponse send() throws IOException {
+        Logger.v(TAG, "HttpWebRequest send thread:" + Process.myTid());
+        final HttpURLConnection connection = setupConnection();
+        final HttpWebResponse response;
```
weijjia: final?? You're overriding it later.

src/src/com/microsoft/aad/adal/AuthenticationContext.java
```diff
@@ -1811,23 +1799,66 @@ private void removeItemFromCache(final RefreshItem refreshItem) {
...
+            // it replaces multi resource refresh token as
+            // well with the new one since it is not stored
+            // with resource.
+            Logger.v(TAG, "Cache is used. It will set item to cache" + request.getLogInfo());
+            setItemToCacheFromRefresh(refreshItem, request, result);
+
+            // return result obj which has error code and
+            // error description that is returned from
+            // server response
+            return result;
```
agamemnum: nit: prefer to have only one return. It looks like early return right now, while it's not

src/src/com/microsoft/aad/adal/AuthenticationContext.java
```diff
@@ -1811,23 +1799,66 @@ private void removeItemFromCache(final RefreshItem refreshItem) {
...
+            throws AuthenticationException {
+        AuthenticationResult result = getTokenWithRefreshToken(request, refreshItem.mRefreshToken);
+
+        if (result == null || StringExtensions.IsNullOrBlank(result.getAccessToken())) {
+            String errLogInfo = result == null ? "" : result.getErrorLogInfo();
+            Logger.e(TAG, "Refresh token did not return accesstoken.", request.getLogInfo()
+                    + errLogInfo, ADALError.AUTH_FAILED_NO_TOKEN);
+
+            // Check error code, only remove token from cache if receiving invalid_grant from server
+            if (AuthenticationConstants.OAuth2ErrorCode.INVALID_GRANT.equals(result.getErrorCode())) {
```
agamemnum: nit: consider to make ignore case here

agamemnum: result can be null in here right? U can get NPE.

src/src/com/microsoft/aad/adal/AuthenticationParameters.java
```diff
@@ -166,17 +167,17 @@ public static void createFromResourceUrl(Context context, final URL resourceUrl,
...
                         onCompleted(exc, null);
```
agamemnum: That's quite strange. So callback can be invoked twice for same call? Once with success and once with failure? Is caller even handles it correctly?

src/src/com/microsoft/aad/adal/HttpWebRequest.java
weijjia: Can you elaborate a little bit more, i don't quite understand.

src/src/com/microsoft/aad/adal/AuthenticationContext.java
weijjia: you're right, should check null for result.

src/src/com/microsoft/aad/adal/Oauth2.java
```diff
@@ -193,7 +194,7 @@ public String buildRefreshTokenRequestMessage(String refreshToken)
 
     public static AuthenticationResult processUIResponseParams(HashMap<String, String> response) {
 
-        AuthenticationResult result = null;
+        final AuthenticationResult result;
```
weijjia: final??

src/src/com/microsoft/aad/adal/Oauth2.java
```diff
@@ -598,16 +593,16 @@ private AuthenticationResult processTokenResponse(HttpWebResponse webResponse) {
         case HttpURLConnection.HTTP_UNAUTHORIZED:
             try {
                 result = parseJsonResponse(webResponse.getBody());
-            } catch (final JSONException jsonException) {
+            } catch (JSONException jsonException) {
```
weijjia: why removing?

src/src/com/microsoft/aad/adal/AuthenticationContext.java
```diff
@@ -1574,45 +1574,33 @@ private String getTokenHash(String token) {
...
-                    mKeyWithUserId = CacheKey.createCacheKey(request, item.getUserInfo()
-                            .getUserId());
-                    mKeyWithDisplayableId = CacheKey.createCacheKey(request, item.getUserInfo()
-                            .getDisplayableId());
-                }
+            mRefreshToken = item.getRefreshToken();
+            mUserInfo = item.getUserInfo();
+            mRawIdToken = item.getRawIdToken();
+            mTenantId = item.getTenantId();
+            if (item.getUserInfo() != null) {
```
agamemnum: nit: creating a local for `item.getUserInfo()` would be easer to read

src/src/com/microsoft/aad/adal/WebRequestHandler.java
```diff
@@ -60,40 +61,31 @@ public WebRequestHandler() {
...
-        if (headers != null && !headers.isEmpty()) {
-            request.getRequestHeaders().putAll(headers);
-        }
-    }
-
     private HashMap<String, String> updateHeaders(HashMap<String, String> headers) {
 
         if (headers == null) {
-            headers = new HashMap<String, String>();
+            headers = new HashMap<>();
```
weijjia: Collections.emptyMap?

src/src/com/microsoft/aad/adal/AuthenticationContext.java
```diff
@@ -1811,23 +1799,66 @@ private void removeItemFromCache(final RefreshItem refreshItem) {
...
+                Logger.v(TAG, "UserInfo is updated from cached result:" + request.getLogInfo());
+                result.setUserInfo(refreshItem.mUserInfo);
+                result.setIdToken(refreshItem.mRawIdToken);
+                result.setTenantId(refreshItem.mTenantId);
+            }
+
+            // it replaces multi resource refresh token as
+            // well with the new one since it is not stored
+            // with resource.
+            Logger.v(TAG, "Cache is used. It will set item to cache" + request.getLogInfo());
```
agamemnum: nit: logging line text is quite interesting. Is there a way for "cache is not used"? And why all logs in this function start with "It"? What "It" represent in here? :-) Normally log lines look like "Saving item into the cache"

src/src/com/microsoft/aad/adal/AuthenticationContext.java
agamemnum: Can we call this method `refreshToken`?

src/src/com/microsoft/aad/adal/Discovery.java
```diff
@@ -238,12 +232,9 @@ private boolean sendRequest(final URL queryUrl) throws MalformedURLException, JS
...
      * @param authorizationEndpointUrl
      * @return https://hostname/common
-     * @throws MalformedURLException
      */
-    private String getAuthorizationCommonEndpoint(final URL authorizationEndpointUrl)
-            throws MalformedURLException {
-        return String.format("https://%s%s", authorizationEndpointUrl.getHost(),
-                AUTHORIZATION_COMMON_ENDPOINT);
+    private String getAuthorizationCommonEndpoint(final URL authorizationEndpointUrl) {
+        return String.format("https://%s%s", authorizationEndpointUrl.getHost(), AUTHORIZATION_COMMON_ENDPOINT);
```
agamemnum: please add locale in here

agamemnum: Better - switch to UrlBuilder...

src/src/com/microsoft/aad/adal/HttpWebRequest.java
agamemnum: it would compile - it works because variable initialized only once and in all code paths

src/src/com/microsoft/aad/adal/HttpWebRequest.java
```diff
@@ -107,107 +92,116 @@ private void setupConnection() {
...
+        final Iterator<String> headerKeys = mRequestHeaders.keySet().iterator();
+
+        while (headerKeys.hasNext()) {
+            String header = headerKeys.next();
+            Logger.v(TAG, "Setting header: " + header);
+            connection.setRequestProperty(header, mRequestHeaders.get(header));
+        }
+
+        // Avoid reuse of existing sockets to avoid random EOF errors
+        System.setProperty("http.keepAlive", "false");
```
agamemnum: Don't forget about this one too :-) I know it's going to be part of another PR.

src/src/com/microsoft/aad/adal/HttpWebRequest.java
```diff
@@ -23,81 +23,66 @@
...
     }
 
     /**
      * setupConnection before sending the request.
      */
-    private void setupConnection() {
+    private HttpURLConnection setupConnection() throws IOException {
         Logger.v(TAG, "HttpWebRequest setupConnection thread:" + android.os.Process.myTid());
         if (mUrl == null) {
             throw new IllegalArgumentException("requestURL");
```
agamemnum: I have to ask - why we aren't merged this "if"? ;-)

src/src/com/microsoft/aad/adal/HttpWebRequest.java
weijjia: hm, i see, it can make sure the response will be only initialized once.

src/src/com/microsoft/aad/adal/Oauth2.java
weijjia: ignore it, i see the purpose here.

src/src/com/microsoft/aad/adal/HttpWebRequest.java
imakhalova: I'll explain in next PR, I'm not going to touch it in this PR

src/src/com/microsoft/aad/adal/HashMapExtensions.java
imakhalova: because now getBody returns String

src/src/com/microsoft/aad/adal/HashMapExtensions.java
imakhalova: right, lets return Map

src/src/com/microsoft/aad/adal/Discovery.java
imakhalova: which variable would you prefer to see final? IOException?

src/src/com/microsoft/aad/adal/WebRequestHandler.java
imakhalova: Why do you want me to create immutable map?

src/src/com/microsoft/aad/adal/Oauth2.java
imakhalova: 1) to keep consistent since nowhere we catch final exception
2) and I don't really see a big point in this
Do you see any advantage to have final here?

weijjia: nope, i just think whenever possible, make it final.

src/src/com/microsoft/aad/adal/AuthenticationContext.java
imakhalova: Haha...it was refreshToken before, but Wei renamed it to getTokenWithRefreshToken
Sorry, I'm not going to rename it back and forth )

src/src/com/microsoft/aad/adal/WebRequestHandler.java
weijjia: sorry, my bad, i thought we're returning here, missing the following header we add.

src/src/com/microsoft/aad/adal/AuthenticationContext.java
imakhalova: agree

src/src/com/microsoft/aad/adal/AuthenticationContext.java
imakhalova: adding this check, but with next PR I'm going to exclude possibility of authResut == null

src/src/com/microsoft/aad/adal/AuthenticationParameters.java
imakhalova: If we fail on parseResponse with ResourceAuthenticationChallengeException we won't invoke success callback,
but if we invoke success it means we didn't throw an exception and we won't invoke callback for error

src/src/com/microsoft/aad/adal/Oauth2.java
imakhalova: ok, reverting

weijjia:
:shipit:

PR #593, opened 2016-04-19
Checking private key before doing unwrap

weijjia:
When key data is wiped by android keystore but alias is not, we may get InvalidArgumentException thrown, to avoid crashing the app, check if private key used to do unwrap is empty or now. If it's already empty, shouldn't bother continuing with unwrap.

### DIFF
--- a/src/src/com/microsoft/aad/adal/StorageHelper.java
+++ b/src/src/com/microsoft/aad/adal/StorageHelper.java
@@ -40,6 +40,7 @@ import java.security.KeyStore;
 import java.security.KeyStoreException;
 import java.security.MessageDigest;
 import java.security.NoSuchAlgorithmException;
+import java.security.PrivateKey;
 import java.security.SecureRandom;
 import java.security.UnrecoverableKeyException;
 import java.security.cert.CertificateException;
@@ -452,7 +453,8 @@ public class StorageHelper {
             // the key data is wiped, if this is the case, the retrieved keypair will
             // be empty, and when we use the private key to do unwrap, we'll encounter 
             // IllegalArgumentException
-            if (mKeyPair.getPrivate() == null || mKeyPair.getPrivate().getEncoded().length == 0) {
+            final PrivateKey privateKey = mKeyPair.getPrivate();
+            if (privateKey == null || privateKey.getEncoded() == null || privateKey.getEncoded().length == 0) {
                 throw new UnrecoverableKeyException("Retrieved private key is empty.");
             }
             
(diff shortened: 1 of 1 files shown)


src/src/com/microsoft/aad/adal/StorageHelper.java
```diff
@@ -450,6 +446,16 @@ final synchronized private SecretKey getSecretKeyFromAndroidKeyStore() throws IO
             if (encryptedKey == null || encryptedKey.length == 0) {
                 throw new UnrecoverableKeyException("Couldn't find encrypted key in file");
             }
+            
+            // Check if the retrieved keypair is empty. With the current limitation of 
+            // AndroidKeyStore, there is possibility that the alias is not wiped but 
+            // the key data is wiped, if this is the case, the retrieved keypair will
+            // be empty, and when we use the private key to do unwrap, we'll encounter 
+            // IllegalArgumentException
+            if (mKeyPair.getPrivate() == null || mKeyPair.getPrivate().getEncoded().length == 0) {
```
imakhalova: + check mKeyPair.getPrivate().getEncoded() != null

imakhalova:
fix potential NPE and good to go

src/src/com/microsoft/aad/adal/StorageHelper.java
```diff
@@ -450,6 +447,17 @@ final synchronized private SecretKey getSecretKeyFromAndroidKeyStore() throws IO
             if (encryptedKey == null || encryptedKey.length == 0) {
                 throw new UnrecoverableKeyException("Couldn't find encrypted key in file");
             }
+            
+            // Check if the retrieved keypair is empty. With the current limitation of 
+            // AndroidKeyStore, there is possibility that the alias is not wiped but 
+            // the key data is wiped, if this is the case, the retrieved keypair will
+            // be empty, and when we use the private key to do unwrap, we'll encounter 
+            // IllegalArgumentException
+            final PrivateKey privateKey = mKeyPair.getPrivate();
```
imakhalova: ideally this should be checked even before wrap operation

agamemnum: I agree it looks like it would be cleaner if `getKeyPairFromAndroidKeyStore` would perform this verification.

src/src/com/microsoft/aad/adal/StorageHelper.java
```diff
@@ -450,6 +447,17 @@ final synchronized private SecretKey getSecretKeyFromAndroidKeyStore() throws IO
...
+            if (privateKey == null || privateKey.getEncoded() == null || privateKey.getEncoded().length == 0) {
+                throw new UnrecoverableKeyException("Retrieved private key is empty.");
```
agamemnum: Can user recover now? we aren't reseting `mKeyPair`

PR #603, opened 2016-04-24
fix tests that was broken by recent changes

imakhalova:
Signature of beginClientMetricsRecord was change, argument switched from HashMap to Map
Number of validHosts increased since Germany gateway was added


weijjia:
:shipit:

PR #604, opened 2016-04-24
reuse of existing sockets and do not disconnect connection due to bug…

imakhalova:
…gy android implementation
1) http.keepAlive was set to false because there is a bug in connection.disconnect method() which can close socket even if there is another connection is connected to this socket. To avoid that it's better to rely on GC to release connection.
It's very bad to set http.keepAlive property to false, because it's global system property and if application really makes lots of network calls to one socket, it improves his performance to reuse existing sockets

2)  When you call getStatusCode before getInputStream HTTP stack behaves really unpredictable and there were some logic added to support that case. 
Back in the past getStatusCode function was called before getInputStream for POST methods and after for GET. 
But now  logic is different and getStatusCode always invoked after getInputStream, so we don't need this magic anymore.

### DIFF
--- a/src/src/com/microsoft/aad/adal/HttpWebRequest.java
+++ b/src/src/com/microsoft/aad/adal/HttpWebRequest.java
@@ -109,8 +109,6 @@ class HttpWebRequest {
             connection.setRequestProperty(header, mRequestHeaders.get(header));
         }
 
-        // Avoid reuse of existing sockets to avoid random EOF errors
-        System.setProperty("http.keepAlive", "false");
         connection.setReadTimeout(READ_TIME_OUT);
         connection.setInstanceFollowRedirects(mInstanceRedirectsFollow);
         connection.setUseCaches(mUseCaches);
@@ -144,7 +142,7 @@ class HttpWebRequest {
             }
             // GET request should read status after getInputStream to make
             // this work for different SDKs
-            final int statusCode = getStatusCode(connection);
+            final int statusCode = connection.getResponseCode();
             final String responseBody = convertStreamToString(responseStream);
 
             // It will only run in debugger and set from outside for testing
@@ -162,7 +160,12 @@ class HttpWebRequest {
             response = new HttpWebResponse(statusCode, responseBody, connection.getHeaderFields());
         } finally {
             safeCloseStream(responseStream);
-            connection.disconnect();
+            // We are not disconnecting from network to allow connection to be returned into the
+            // connection pool. If we call disconnect due to buggy implementation we are not reusing
+            // connections.
+            //if (connection != null) {
+            //	connection.disconnect();
+            //}
         }
 
         return response;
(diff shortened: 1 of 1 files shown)


src/src/com/microsoft/aad/adal/HttpWebRequest.java
```diff
@@ -198,40 +201,6 @@ private static String convertStreamToString(InputStream is) throws IOException {
...
     }
 
-    private static int getStatusCode(HttpURLConnection connection) throws IOException {
-        int statusCode = HttpURLConnection.HTTP_BAD_REQUEST;
-
-        try {
-            statusCode = connection.getResponseCode();
-        } catch (IOException ex) {
-
-            if (Build.VERSION.SDK_INT < 16) {
```
weijjia: Can you remind me why we're removing this?

imakhalova: It's actually in description to this PR.
We don't need to getResponseCode() twice anymore, since we call this method only after getInputStream90

src/src/com/microsoft/aad/adal/HttpWebRequest.java
```diff
@@ -109,8 +109,6 @@ private HttpURLConnection setupConnection() throws IOException {
             connection.setRequestProperty(header, mRequestHeaders.get(header));
         }
 
-        // Avoid reuse of existing sockets to avoid random EOF errors
-        System.setProperty("http.keepAlive", "false");
```
sdabbiru: I would recommend that this change be put in selfhost and tested before releasing to the public.

src/src/com/microsoft/aad/adal/HttpWebRequest.java
```diff
@@ -144,7 +142,7 @@ public HttpWebResponse send() throws IOException {
             }
             // GET request should read status after getInputStream to make
             // this work for different SDKs
-            final int statusCode = getStatusCode(connection);
+            final int statusCode = connection.getResponseCode();
```
sdabbiru: Should we not handle IOException here?

imakhalova: ? what do you mean by "not handle", do you think we should crash?
this method throws IOException in multiple places, we handle them all

src/src/com/microsoft/aad/adal/HttpWebRequest.java
imakhalova: Then do it please

agamemnum: I would strongly advice to remove this ASAP from the ADAL library. I hope there is an understanding of impact on consumers of a library by writing this code. In short this performed a global disable of the connection pipe-lining for every consumer of the ADAL library. For instance when client app downloads thumbnails it make new socket for each of them. Even if there is an EOF exception (I strongly it would be resolved by proposed fix), it still doesn't justify this code being written in the shared library.

src/src/com/microsoft/aad/adal/HttpWebRequest.java
weijjia: I need you to help me understand what's changed? I may miss something, the only place I saw getStatusCode is called is in the send method, and it's called after getInputStream. The code comment here clearly shows that if the returned 401 challenge is empty, and when you call getResponseCode, you'll get exception when reading the code.

weijjia:
Ok, I see the change you mentioned. Merging then.

PR #605, opened 2016-04-24
If server responded with some unknown error do not force user to sign…

imakhalova:
… in again

All unknown server responses should be recoverable and we shouldn't require UI prompt in case of server failures

What about specific server responses it's questionable how library should behave. Added todo as we need to discuss it with server team and decide what we want to do.


weijjia:
looks good to me.

PR #606, opened 2016-04-24
HOTFIX: revert Checking private key before doing unwrap #593

imakhalova:
This exception happens too frequently.
I fixed UTs for keyStore and test around keystore actually fails because of this exception, so 
privateKey.getEncoded() might be null, but  unwrap method works just fine with that
It means that we currently removing good tokens that were encrypted with keystore key
I think we need to revert and hotfix it

### DIFF
--- a/src/src/com/microsoft/aad/adal/StorageHelper.java
+++ b/src/src/com/microsoft/aad/adal/StorageHelper.java
@@ -448,16 +448,6 @@ public class StorageHelper {
                 throw new UnrecoverableKeyException("Couldn't find encrypted key in file");
             }
             
-            // Check if the retrieved keypair is empty. With the current limitation of 
-            // AndroidKeyStore, there is possibility that the alias is not wiped but 
-            // the key data is wiped, if this is the case, the retrieved keypair will
-            // be empty, and when we use the private key to do unwrap, we'll encounter 
-            // IllegalArgumentException
-            final PrivateKey privateKey = mKeyPair.getPrivate();
-            if (privateKey == null || privateKey.getEncoded() == null || privateKey.getEncoded().length == 0) {
-                throw new UnrecoverableKeyException("Retrieved private key is empty.");
-            }
-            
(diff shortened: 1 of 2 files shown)


src/src/com/microsoft/aad/adal/StorageHelper.java
```diff
@@ -448,16 +448,6 @@ final synchronized private SecretKey getSecretKeyFromAndroidKeyStore() throws IO
                 throw new UnrecoverableKeyException("Couldn't find encrypted key in file");
             }
             
-            // Check if the retrieved keypair is empty. With the current limitation of 
-            // AndroidKeyStore, there is possibility that the alias is not wiped but 
-            // the key data is wiped, if this is the case, the retrieved keypair will
-            // be empty, and when we use the private key to do unwrap, we'll encounter 
-            // IllegalArgumentException
-            final PrivateKey privateKey = mKeyPair.getPrivate();
```
weijjia: How does it perform unwrap if the key is null?

src/src/com/microsoft/aad/adal/StorageHelper.java
```diff
@@ -448,16 +448,6 @@ final synchronized private SecretKey getSecretKeyFromAndroidKeyStore() throws IO
...
-            if (privateKey == null || privateKey.getEncoded() == null || privateKey.getEncoded().length == 0) {
```
weijjia: we should still check if privateKey == null, but not checking privateKey.getEncoded() and length. 
Agree, this needs to be hot fixed.

weijjia:
Talk with @imakhalova, to reduce the risk, revert the change now. And we'll think of including back the privatekey check with more testing around it.

src/src/com/microsoft/aad/adal/StorageHelper.java
imakhalova: if it was ever null we would get InvalidKeyException("Need RSA private or public key"); 
But we've never had this crash

PR #607, opened 2016-04-24
Refactor storageHelper: Decryption only read keys not writing

weijjia:
1. Separate the key reading for encryption/decryption
2. Decryption should only read keys not writing.
3. Adding telemetry event logging for decryption/encryption failure, also add the logging for keystore failure.


weijjia:
@imakhalova ^^^

src/src/com/microsoft/aad/adal/StorageHelper.java
```diff
@@ -123,124 +129,94 @@
...
+        // in the digest. It defines what to use for Mac Key.
+        mac.init(mMacKey);
+        mac.update(blobVersion);
+        mac.update(encrypted);
+        mac.update(iv);
+        byte[] macDigest = mac.doFinal();
 
-        throw new IllegalArgumentException("keyVersion = " + keyVersion);
-    }
+        // Init array to store blobVersion, encrypted data, iv, macdigest
```
weijjia: Plan to put this into a object. i.e. EncrpyBlobContainer. But will be in a separate PR.

src/src/com/microsoft/aad/adal/DefaultTokenCacheStore.java
```diff
@@ -98,12 +98,6 @@ public DefaultTokenCacheStore(Context context) {
         if (mPrefs == null) {
             throw new IllegalStateException(ADALError.DEVICE_SHARED_PREF_IS_NOT_AVAILABLE.getDescription());
         }
-
-        try {
-            getStorageHelper().loadSecretKeyForAPI();
```
imakhalova: I would leave loadSecretKeyForEncryption() here, because if secret key wasn't set dev should know it before sign in process. And this load can be removed from encryption

src/src/com/microsoft/aad/adal/StorageHelper.java
```diff
@@ -264,28 +240,27 @@ public String decrypt(String value) throws GeneralSecurityException, IOException
 
         // get key version used for this data. If user upgraded to different
         // API level, data needs to be updated
-        String keyVersionCheck = new String(bytes, 0, KEY_VERSION_BLOB_LENGTH,
+        String encryptedDataKeyVersion = new String(bytes, 0, KEY_VERSION_BLOB_LENGTH,
```
imakhalova: nit: name it just keyVersion

tests/Functional/src/com/microsoft/aad/adal/test/StorageHelperTests.java
```diff
@@ -54,18 +57,12 @@ protected void setUp() throws Exception {
...
         Log.d(TAG, "setup key at settings");
-        if (AuthenticationSettings.INSTANCE.getSecretKeyData() == null) {
-            // use same key for tests
-            SecretKeyFactory keyFactory = SecretKeyFactory
-                    .getInstance("PBEWithSHA256And256BitAES-CBC-BC");
-            SecretKey tempkey = keyFactory.generateSecret(new PBEKeySpec("test".toCharArray(),
-                    "abcdedfdfd".getBytes("UTF-8"), 100, 256));
-            SecretKey secretKey = new SecretKeySpec(tempkey.getEncoded(), "AES");
-            AuthenticationSettings.INSTANCE.setSecretKey(secretKey.getEncoded());
+        if (AuthenticationSettings.INSTANCE.getSecretKeyData() == null && Build.VERSION.SDK_INT < 18) {
```
imakhalova: I think we should actually have 2 different set of tests: 1) if secretkey is null 2) when it's set 
Because I don' think that anyone will launch tests on older versions.

src/src/com/microsoft/aad/adal/DefaultTokenCacheStore.java
```diff
@@ -139,7 +133,7 @@ private String decrypt(final String key, final String value) {
         try {
             return getStorageHelper().decrypt(value);
         } catch (GeneralSecurityException | IOException e) {
-            Logger.e(TAG, "Decryption failure", "", ADALError.ENCRYPTION_FAILED, e);
+            Logger.e(TAG, "Decryption failure", "", ADALError.DECRYPTION_FAILED, e);
```
imakhalova: If you'd like to get telemetry how frequently this error occurs use ClientAnalytics to log this event and information that might help you to investigate what happened

src/src/com/microsoft/aad/adal/StorageHelper.java
```diff
@@ -123,124 +129,94 @@
     private static final String ENCODE_VERSION = "E1";
 
     private static final Object LOCK_OBJ = new Object();
+    
+    private final Context mContext;
+    private final SecureRandom mRandom;
 
+    /**
+     * it is needed for AndroidKeyStore.
```
imakhalova: private and public keys that generated in keystore

src/src/com/microsoft/aad/adal/StorageHelper.java
```diff
@@ -305,71 +280,204 @@ public String decrypt(String value) throws GeneralSecurityException, IOException
...
     }
+    
+    /**
+     * Get Secret Key based on API level to use in encryption. Decryption key
+     * depends on version# since user can migrate to new Android.OS
+     *
+     * @return
+     * @throws NoSuchAlgorithmException
+     */
+    public synchronized SecretKey loadSecretKeyForEncryption() throws IOException, GeneralSecurityException {
```
imakhalova: private?

src/src/com/microsoft/aad/adal/StorageHelper.java
```diff
@@ -305,71 +280,204 @@ public String decrypt(String value) throws GeneralSecurityException, IOException
...
+        return mKey;
     }
+    
+    /**
+     * load key based on version for migration
+     * 
+     * @throws GeneralSecurityException
+     * @throws IOException
+     */
+    public synchronized SecretKey loadSecretKeyForDecryption(String keyVersion) throws GeneralSecurityException, IOException {
```
imakhalova: private

src/src/com/microsoft/aad/adal/StorageHelper.java
```diff
@@ -401,69 +509,44 @@ final private SecretKey generateSecretKey() throws NoSuchAlgorithmException {
...
+            Cipher wrapCipher = Cipher.getInstance(WRAP_ALGORITHM);
+            unwrappedSecretKey = unwrap(wrapCipher, wrappedSecretKey);
             Logger.v(TAG, "Finished reading SecretKey");
         } catch (GeneralSecurityException | IOException ex) {
             // Reset KeyPair info so that new request will generate correct KeyPairs.
             // All tokens with previous SecretKey are not possible to decrypt.
             Logger.e(TAG, "Unwrap failed for AndroidKeyStore", "", ADALError.ANDROIDKEYSTORE_FAILED, ex);
             mKeyPair = null;
-            sSecretKeyFromAndroidKeyStore = null;
+            unwrappedSecretKey = null;
```
imakhalova: nit: can be removed

src/src/com/microsoft/aad/adal/StorageHelper.java
```diff
@@ -401,69 +509,44 @@ final private SecretKey generateSecretKey() throws NoSuchAlgorithmException {
...
         try {
-            final byte[] encryptedKey = readKeyData(keyFile);
-            if (encryptedKey == null || encryptedKey.length == 0) {
+            final byte[] wrappedSecretKey = readKeyData(keyFile);
+            if (wrappedSecretKey == null || wrappedSecretKey.length == 0) {
                 throw new UnrecoverableKeyException("Couldn't find encrypted key in file");
             }
             
-            sSecretKeyFromAndroidKeyStore = unwrap(wrapCipher, encryptedKey);
+            if (privateKey == null) {
```
imakhalova: 1) I don't see initialization of privateKey
2) I don't think that this error should be recoverable, if this happened it means something went wrong and we swallowed some keystore exception. Do not make it recoverable

src/src/com/microsoft/aad/adal/StorageHelper.java
```diff
@@ -123,124 +129,94 @@
...
     }
 
     /**
-     * Get Secret Key based on API level to use in encryption. Decryption key
-     * depends on version# since user can migrate to new Android.OS
+     * encrypt text with current key based on API level
      *
-     * @return
-     * @throws NoSuchAlgorithmException
+     * @param clearText Clar text to encrypt.
```
sdabbiru: NIT Typo: Clar

src/src/com/microsoft/aad/adal/StorageHelper.java
```diff
@@ -123,124 +129,94 @@
...
+        return getEncodeVersionLengthPrefix() + ENCODE_VERSION + encryptedText;
     }
 
     /**
-     * Derive mac key from given key
-     * 
-     * @param key
-     * @return
-     * @throws NoSuchAlgorithmException
+     * Decrypt then blob with either user provided key or key persisted in AndroidKeyStore.
```
sdabbiru: Please update the comment.

src/src/com/microsoft/aad/adal/StorageHelper.java
```diff
@@ -264,28 +240,27 @@ public String decrypt(String value) throws GeneralSecurityException, IOException
 
         // get key version used for this data. If user upgraded to different
```
sdabbiru: Please refrain from throwing illegalArgumentExceptions when encodedVersionLength etc do not appear to be valid. We should throw a non-runtime exception.

src/src/com/microsoft/aad/adal/StorageHelper.java
```diff
@@ -264,28 +240,27 @@ public String decrypt(String value) throws GeneralSecurityException, IOException
...
+        SecretKey macKey = getMacKey(secretKey);
 
         // byte input array: encryptedData-iv-macDigest
         int ivIndex = bytes.length - DATA_KEY_LENGTH - MAC_LENGTH;
         int macIndex = bytes.length - MAC_LENGTH;
         int encryptedLength = ivIndex - KEY_VERSION_BLOB_LENGTH;
         if (ivIndex < 0 || macIndex < 0 || encryptedLength < 0) {
-            throw new IllegalArgumentException(
-                    "Given value is smaller than the IV vector and MAC length");
+            throw new IllegalArgumentException("Invalid byte array input for decryption.");
```
sdabbiru: There need not be a consumer developer 'error' that can cause this. So, we should not be throwing IllegalArgExceptions here... Let's reserve the IllegalArgExceptions only for conveying direct ADAL user errors.

src/src/com/microsoft/aad/adal/StorageHelper.java
```diff
@@ -305,71 +280,204 @@ public String decrypt(String value) throws GeneralSecurityException, IOException
         Logger.v(TAG, "Finished decryption");
         return decrypted;
     }
+    
+    /**
+     * Get Secret Key based on API level to use in encryption. Decryption key
```
sdabbiru: NIT: Decryption key?

src/src/com/microsoft/aad/adal/StorageHelper.java
```diff
@@ -305,71 +280,204 @@ public String decrypt(String value) throws GeneralSecurityException, IOException
...
+            mKey = getSecretKeyFromAndroidKeyStore();
+            mMacKey = getMacKey(mKey);
+            mBlobVersion = VERSION_ANDROID_KEY_STORE;
+        }
+
+        return mKey;
     }
+    
+    /**
+     * load key based on version for migration
```
sdabbiru: migration?

src/src/com/microsoft/aad/adal/StorageHelper.java
```diff
@@ -305,71 +280,204 @@ public String decrypt(String value) throws GeneralSecurityException, IOException
...
+        }
 
+        if (keyVersion.equals(VERSION_ANDROID_KEY_STORE)) {
+            if (Build.VERSION.SDK_INT >= 18) {
+                // androidKeyStore can store app specific self signed cert.
+                // Asymmetric cryptography is used to protect the session key
+                // used for Encryption and HMac
+                return readSecretKeyFromKeyStoreForDecryption();
+            } else {
+                throw new IllegalArgumentException(
```
sdabbiru: Why not throw this:

   +            throw new IllegalStateException("Secret key must be provided for API < 18. " +  
 300   +                    "Use AuthenticationSettings.INSTANCE.setSecretKey()");

src/src/com/microsoft/aad/adal/StorageHelper.java
```diff
@@ -305,71 +280,204 @@ public String decrypt(String value) throws GeneralSecurityException, IOException
...
+        // Loading file and unwrapping this key is causing performance issue.
+        if (mSecretKeyFromAndroidKeyStore != null) {
+            return mSecretKeyFromAndroidKeyStore;
+        }
 
-        if (StringExtensions.IsNullOrBlank(clearText)) {
-            throw new IllegalArgumentException("Input is empty or null");
+        if (mKeyPair == null) {
+            mKeyPair = getKeyPairFromAndroidKeyStoreForEncryption();
+            Logger.v(TAG, "Retrived keypair from androidKeyStore");
```
sdabbiru: In this entire file, it is worth logging the method name.

src/src/com/microsoft/aad/adal/StorageHelper.java
```diff
@@ -305,71 +280,204 @@ public String decrypt(String value) throws GeneralSecurityException, IOException
...
+                ADALKS);
+        if (mKeyPair == null) {
+            final KeyStore keyStore = KeyStore.getInstance("AndroidKeyStore");
+            keyStore.load(null);
+            if (!keyStore.containsAlias(KEY_STORE_CERT_ALIAS)) {
+                Logger.v(TAG, "Key store alias is not avaliable.");
+                // Due to an issue with android keystore, when user change lockscreen type, 
+                // key data and alias could be wiped. Key data will be unrecoverable if being 
+                // wiped out.
+                throw new UnrecoverableKeyException("AndroidKeyStore does not contain the alias.");
```
sdabbiru: How is this exception surfaced to the app developer? What are they supposed to do when they see this exception?

src/src/com/microsoft/aad/adal/StorageHelper.java
```diff
@@ -305,71 +280,204 @@ public String decrypt(String value) throws GeneralSecurityException, IOException
...
-        System.arraycopy(macDigest, 0, blobVerAndEncryptedDataAndIVAndMacDigest, blobVersion.length
-                + encrypted.length + iv.length, macDigest.length);
+    private SecretKey getSecretKey(byte[] rawBytes) {
+        if (rawBytes != null)
+            return new SecretKeySpec(rawBytes, KEYSPEC_ALGORITHM);
 
-        String encryptedText = new String(Base64.encode(blobVerAndEncryptedDataAndIVAndMacDigest,
-                Base64.NO_WRAP), AuthenticationConstants.ENCODING_UTF8);
-        Logger.v(TAG, "Finished encryption");
+        throw new IllegalArgumentException("rawBytes");
```
sdabbiru: Nope.

src/src/com/microsoft/aad/adal/StorageHelper.java
```diff
@@ -401,69 +509,45 @@ final private SecretKey generateSecretKey() throws NoSuchAlgorithmException {
...
-            if (encryptedKey == null || encryptedKey.length == 0) {
+            final byte[] wrappedSecretKey = readKeyData(keyFile);
+            if (wrappedSecretKey == null || wrappedSecretKey.length == 0) {
                 throw new UnrecoverableKeyException("Couldn't find encrypted key in file");
             }
             
-            sSecretKeyFromAndroidKeyStore = unwrap(wrapCipher, encryptedKey);
+            final PrivateKey privateKey = mKeyPair.getPrivate();
+            if (privateKey == null) {
+                throw new UnrecoverableKeyException("Retrieved private key is empty.");
```
sdabbiru: I want to know how this exception is surfaced to the developer. I wish we handle this internally (may be that's what is already happening).

src/src/com/microsoft/aad/adal/StorageHelper.java
```diff
@@ -123,124 +129,94 @@
...
+     * @return Encrypted blob.
+     * @throws GeneralSecurityException
+     * @throws UnsupportedEncodingException
      */
-    public synchronized SecretKey loadSecretKeyForAPI() throws IOException, GeneralSecurityException {
-        // Loading key only once for performance. If API is upgraded, it will
-        // restart the device anyway. It will load the correct key for new API.
-        if (mKey != null && mMacKey != null)
-            return mKey;
+    public String encrypt(String clearText)
```
imakhalova: private

src/src/com/microsoft/aad/adal/StorageHelper.java
```diff
@@ -401,69 +509,45 @@ final private SecretKey generateSecretKey() throws NoSuchAlgorithmException {
...
         }
-
-        // Read from file again
-        Logger.v(TAG, "Reading SecretKey");
+        
         try {
-            final byte[] encryptedKey = readKeyData(keyFile);
-            if (encryptedKey == null || encryptedKey.length == 0) {
+            final byte[] wrappedSecretKey = readKeyData(keyFile);
+            if (wrappedSecretKey == null || wrappedSecretKey.length == 0) {
```
imakhalova: remove this check...
and the one below

src/src/com/microsoft/aad/adal/StorageHelper.java
```diff
@@ -123,124 +129,94 @@
...
-                mMacKey = getMacKey(mKey);
-                mBlobVersion = VERSION_ANDROID_KEY_STORE;
-            } catch (IOException | GeneralSecurityException e) {
-                Logger.e(TAG, "Failed to get private key from AndroidKeyStore", "",
-                        ADALError.ANDROIDKEYSTORE_FAILED, e);
-                throw e;
-            }
-        }
+        // load key for encryption if not loaded
+        loadSecretKeyForEncryption();
```
agamemnum: If this function returns keys - you shouldn't use `mKey`. You should use what function returns

src/src/com/microsoft/aad/adal/StorageHelper.java
```diff
@@ -305,71 +280,204 @@ public String decrypt(String value) throws GeneralSecurityException, IOException
...
+     * Get Secret Key based on API level to use in encryption. Decryption key
+     * depends on version# since user can migrate to new Android.OS
+     *
+     * @return
+     * @throws NoSuchAlgorithmException
+     */
+    public synchronized SecretKey loadSecretKeyForEncryption() throws IOException, GeneralSecurityException {
+        // Loading key only once for performance. If API is upgraded, it will
+        // restart the device anyway. It will load the correct key for new API.
+        if (mKey != null && mMacKey != null)
```
agamemnum: Why this function doesn't invoke `loadSecretKeyForDecryption` and if it fails fallback to generate new keys

src/src/com/microsoft/aad/adal/StorageHelper.java
```diff
@@ -305,71 +280,204 @@ public String decrypt(String value) throws GeneralSecurityException, IOException
...
+    /**
+     * Get key pair from AndroidKeyStore.
+     * 
+     * @return
+     * @throws GeneralSecurityException
+     * @throws IOException
+     */
+    @TargetApi(Build.VERSION_CODES.JELLY_BEAN_MR2)
+    private synchronized KeyPair getKeyPairFromAndroidKeyStoreForEncryption() throws GeneralSecurityException, IOException {
+        final KeyStore keyStore = KeyStore.getInstance("AndroidKeyStore");
```
agamemnum: same comment here. It would be better to call `readSecretKeyFromKeyStoreForDecryption` if it fails fallback to generate new key

agamemnum:
Overall on this CR - what would it fix? It complicates code quite a bit, but I don't see clear benefit. It's super rare when we do a read without a write right after it.

weijjia:
Read can follow by write, but decryption should only read keys. Now, if we fail to read keypair from keystore, we generate a new keypair, then use it. It just adds one more layer logic here, the newly generated keypair won't work with the existing items, then why bother generating and using it?

agamemnum:
Code quite a bit more complicated and I don't see what it really fix/improve....

agamemnum:
On design - you should create two methods `getKey` or `getKeyOrCreate`. `getKeyOrCreate` should call `getKey` if generation fails should perform key generation. 
`encrypt` should always call `getKey` and use it's result i.e. never use `mKey` and so on.
`decrypt` should call `getKeyOrCreate`

weijjia:
@agamemnum why decrypt should call getKeyOrCreate? One of the other reason I want to avoid writing keys for decryption is, we currently advise apps to supply us the symmetric key all the time despite of the API version. However, there are still old items encrypted in the cache with key persisted in keystore. When we fail to read the keys, we to generate a new one, I've seen IllegalStateException thrown during the key generation process reported by some apps. 
I agree we can create a getKey and getKeyOrCreate to make the logic clearer, but I don't follow why decryption needs to call getKeyOrCreate? why am I creating a new one during decryption?

src/src/com/microsoft/aad/adal/StorageHelper.java
weijjia: Cannot make it package visible, broker is implementing its own cache, but also use the encrypt/decrypt provided in the helper.

src/src/com/microsoft/aad/adal/StorageHelper.java
```diff
@@ -305,70 +283,241 @@ public String decrypt(String value) throws GeneralSecurityException, IOException
...
 
-        // Mac output to sign encryptedData+IV. Keyversion is not included
-        // in the digest. It defines what to use for Mac Key.
-        mac.init(mMacKey);
-        mac.update(blobVersion);
-        mac.update(encrypted);
-        mac.update(iv);
-        byte[] macDigest = mac.doFinal();
+    private SecretKey getSecretKey(byte[] rawBytes) {
+        if (rawBytes != null)
```
addev-ashish: nit: braces around single line conditional seems to be the convention in this file

src/src/com/microsoft/aad/adal/StorageHelper.java
```diff
@@ -68,18 +69,31 @@
  */
 public class StorageHelper {
 
+    /**
+     * Mac key hashing alogirighm.
+     */
     private static final String MAC_KEY_HASH_ALGORITHM = "SHA256";
```
weijjia: rename it to HMAC_KEY_****

src/src/com/microsoft/aad/adal/StorageHelper.java
```diff
@@ -68,18 +69,31 @@
  */
 public class StorageHelper {
```
weijjia: Check if we can make this class package visible.

src/src/com/microsoft/aad/adal/StorageHelper.java
```diff
@@ -305,70 +283,241 @@ public String decrypt(String value) throws GeneralSecurityException, IOException
...
     /**
-     * encrypt text with current key based on API level
+     * Get Secret Key based on API level to use in encryption. Decryption key
+     * depends on version# since user can migrate to new Android.OS
      *
-     * @param clearText
      * @return
+     * @throws NoSuchAlgorithmException
+     */
+    public synchronized SecretKey loadSecretKeyForEncryption() throws IOException, GeneralSecurityException {
```
weijjia: package visible

src/src/com/microsoft/aad/adal/StorageHelper.java
```diff
@@ -122,124 +129,96 @@
...
-                mBlobVersion = VERSION_ANDROID_KEY_STORE;
-            } catch (IOException | GeneralSecurityException e) {
-                Logger.e(TAG, "Failed to get private key from AndroidKeyStore", "",
-                        ADALError.ANDROIDKEYSTORE_FAILED, e);
-                throw e;
-            }
-        }
+        // load key for encryption if not loaded
+        mKey = loadSecretKeyForEncryption();
+        mMacKey = getMacKey(mKey);
```
weijjia: rename it mHMACKey

src/src/com/microsoft/aad/adal/StorageHelper.java
weijjia: Put this in telemetry. 
Throw IOException.

src/src/com/microsoft/aad/adal/StorageHelper.java
weijjia: add parentheses for single line.

src/src/com/microsoft/aad/adal/DefaultTokenCacheStore.java
```diff
@@ -98,11 +98,14 @@ public DefaultTokenCacheStore(Context context) {
...
+        // Check upfront when initializing DefaultTokenCacheStore. 
+        // If it's under API 18 and secretkey is not provided, we should fail upfront to inform 
+        // notify developers. 
         try {
-            getStorageHelper().loadSecretKeyForAPI();
-        } catch (IOException | GeneralSecurityException e) {
-            throw new IllegalStateException("Failed to get private key from AndroidKeyStore", e);
+            getStorageHelper().loadSecretKeyForEncryption();
+        } catch (final IOException | GeneralSecurityException e) {
+            Logger.v(TAG, "Fail to create keys from android keystore.");
```
weijjia: only check if the key is supplied for API < 18, also extract it into a separate separate method. validatekeySetting

src/src/com/microsoft/aad/adal/StorageHelper.java
```diff
@@ -305,70 +283,241 @@ public String decrypt(String value) throws GeneralSecurityException, IOException
...
+     */
+    public synchronized SecretKey loadSecretKeyForEncryption() throws IOException, GeneralSecurityException {
+        // Loading key only once for performance. If API is upgraded, it will
+        // restart the device anyway. It will load the correct key for new API.
+        if (mKey != null && mMacKey != null)
+            return mKey;
+
+        final byte[] secretKeyData = AuthenticationSettings.INSTANCE.getSecretKeyData();
+        if(secretKeyData == null && Build.VERSION.SDK_INT < Build.VERSION_CODES.JELLY_BEAN_MR2) {
+            throw new IllegalStateException("Secret key must be provided for API < 18. " +
```
weijjia: remove the throw here.

src/src/com/microsoft/aad/adal/StorageHelper.java
```diff
@@ -305,70 +283,241 @@ public String decrypt(String value) throws GeneralSecurityException, IOException
...
+     * Supported API >= 18 PrivateKey is stored in AndroidKeyStore. Loads key
+     * from the file if it exists. If not exist, it will generate one.
+     * @param keyVersion
+     * @return
+     * @throws GeneralSecurityException
+     * @throws IOException
+     */
+    private synchronized SecretKey getKeyOrCreate(final String keyVersion) throws GeneralSecurityException, IOException {
+        if (VERSION_USER_DEFINED.equals(keyVersion)) {
+            return getKey(keyVersion);
```
weijjia: getSecretKey(AuthenticationSettings.INSTANCE.getSecretKeyData());

src/src/com/microsoft/aad/adal/StorageHelper.java
```diff
@@ -305,70 +283,241 @@ public String decrypt(String value) throws GeneralSecurityException, IOException
...
+        }
+        
+        try {
+            mSecretKeyFromAndroidKeyStore = getKey(keyVersion);
+        } catch (final IOException | GeneralSecurityException exception) {
+            // If encountering exception for reading keys, try to generate new keys
```
weijjia: Move this out of the catch block. 
Check if mSecretKeyFromAndroidKeyStore is null, and generate the key.

src/src/com/microsoft/aad/adal/StorageHelper.java
```diff
@@ -305,70 +283,241 @@ public String decrypt(String value) throws GeneralSecurityException, IOException
...
+     * @throws GeneralSecurityException
+     * @throws IOException
+     */
+    private synchronized SecretKey getKey(final String keyVersion) throws GeneralSecurityException, IOException {
+        if (VERSION_USER_DEFINED.equals(keyVersion)) {
+            return getSecretKey(AuthenticationSettings.INSTANCE.getSecretKeyData());
+        }
+        
+        if (VERSION_ANDROID_KEY_STORE.equals(keyVersion)) {
+            if (Build.VERSION.SDK_INT >= 18) {
```
addev-ashish: Flip this check and throw

src/src/com/microsoft/aad/adal/StorageHelper.java
```diff
@@ -305,70 +283,241 @@ public String decrypt(String value) throws GeneralSecurityException, IOException
...
+        try {
+            mSecretKeyFromAndroidKeyStore = getKey(keyVersion);
+        } catch (final IOException | GeneralSecurityException exception) {
+            // If encountering exception for reading keys, try to generate new keys
+            mKeyPair = generateKeyPairFromAndroidKeyStore();
+            
+            // Store secret key in a file after wrapping
+            File keyFile = new File(mContext.getDir(mContext.getPackageName(), Context.MODE_PRIVATE),
+                    ADALKS);
+            Cipher wrapCipher = Cipher.getInstance(WRAP_ALGORITHM);
```
weijjia: move this wrapCipher inside the wrap method.

src/src/com/microsoft/aad/adal/StorageHelper.java
```diff
@@ -305,70 +283,241 @@ public String decrypt(String value) throws GeneralSecurityException, IOException
...
+        }
+        
+        try {
+            mSecretKeyFromAndroidKeyStore = getKey(keyVersion);
+        } catch (final IOException | GeneralSecurityException exception) {
+            // If encountering exception for reading keys, try to generate new keys
+            mKeyPair = generateKeyPairFromAndroidKeyStore();
+            
+            // Store secret key in a file after wrapping
+            File keyFile = new File(mContext.getDir(mContext.getPackageName(), Context.MODE_PRIVATE),
```
weijjia: move the keyFile into writeKeyData

src/src/com/microsoft/aad/adal/StorageHelper.java
```diff
@@ -305,70 +283,241 @@ public String decrypt(String value) throws GeneralSecurityException, IOException
...
-        final byte[] bytes = clearText.getBytes(AuthenticationConstants.ENCODING_UTF8);
+    /**
+     * Get the saved key. Will only do read operation. 
+     * @param keyVersion
+     * @return
+     * @throws GeneralSecurityException
+     * @throws IOException
+     */
+    private synchronized SecretKey getKey(final String keyVersion) throws GeneralSecurityException, IOException {
+        if (VERSION_USER_DEFINED.equals(keyVersion)) {
```
weijjia: throw in the beginning.

src/src/com/microsoft/aad/adal/StorageHelper.java
```diff
@@ -305,70 +283,241 @@ public String decrypt(String value) throws GeneralSecurityException, IOException
...
+            return getSecretKey(AuthenticationSettings.INSTANCE.getSecretKeyData());
+        }
+        
+        if (VERSION_ANDROID_KEY_STORE.equals(keyVersion)) {
+            if (Build.VERSION.SDK_INT >= 18) {
+                // androidKeyStore can store app specific self signed cert.
+                // Asymmetric cryptography is used to protect the session key
+                // used for Encryption and HMac
+                final File keyFile = new File(mContext.getDir(mContext.getPackageName(), Context.MODE_PRIVATE),
```
weijjia: move the keyFile into getUnwrappedSecretKey

src/src/com/microsoft/aad/adal/DefaultTokenCacheStore.java
sdabbiru: Does ClientAnalytics record the calling app details such as name, version etc?

---
In reply to: [60865909](https://github.com/AzureAD/azure-activedirectory-library-for-android/pull/607#discussion_r60865909) [](ancestors = 60865909)

src/src/com/microsoft/aad/adal/StorageHelper.java
```diff
@@ -103,13 +117,6 @@
...
     public static final int MAC_LENGTH = 32;
 
-    /**
-     * it is needed for AndroidKeyStore.
-     */
-    private KeyPair mKeyPair;
-
-    private Context mContext;
-
     public static final String VERSION_ANDROID_KEY_STORE = "A001";
```
weijjia: Add comment here explaining what the version is.

src/src/com/microsoft/aad/adal/StorageHelper.java
weijjia: Change it to a switch/case, throw IOException.

weijjia: When throwing IOException, also add ClientAnalystics

src/src/com/microsoft/aad/adal/StorageHelper.java
```diff
@@ -305,70 +283,241 @@ public String decrypt(String value) throws GeneralSecurityException, IOException
...
-        byte[] iv = new byte[DATA_KEY_LENGTH];
-        mRandom.nextBytes(iv);
-        IvParameterSpec ivSpec = new IvParameterSpec(iv);
+    @TargetApi(Build.VERSION_CODES.JELLY_BEAN_MR2)
+    private synchronized KeyPair generateKeyPairFromAndroidKeyStore() throws GeneralSecurityException, IOException {
+        final KeyStore keyStore = KeyStore.getInstance("AndroidKeyStore");
+        keyStore.load(null);
+        
+        Logger.v(TAG, "Generate KeyPair from AndroidKeyStore");
+        Calendar start = Calendar.getInstance();
```
weijjia: final

src/src/com/microsoft/aad/adal/StorageHelper.java
```diff
@@ -305,70 +283,241 @@ public String decrypt(String value) throws GeneralSecurityException, IOException
...
-        cipher.init(Cipher.ENCRYPT_MODE, mKey, ivSpec);
+    /**
+     * Read KeyPair from AndroidKeyStore. 
+     */
+    private synchronized KeyPair readKeyPair() throws GeneralSecurityException, IOException {
+        if (doesKeyPairExist()) {
+            // Read key pair again
+            Logger.v(TAG, "Reading Key entry");
+            try {
+                KeyStore keyStore = KeyStore.getInstance("AndroidKeyStore");
```
weijjia: final

src/src/com/microsoft/aad/adal/StorageHelper.java
```diff
@@ -305,70 +283,241 @@ public String decrypt(String value) throws GeneralSecurityException, IOException
...
-        // Set to encrypt mode
-        Cipher cipher = Cipher.getInstance(CIPHER_ALGORITHM);
-        Mac mac = Mac.getInstance(MAC_ALGORITHM);
-        cipher.init(Cipher.ENCRYPT_MODE, mKey, ivSpec);
+    /**
+     * Read KeyPair from AndroidKeyStore. 
+     */
+    private synchronized KeyPair readKeyPair() throws GeneralSecurityException, IOException {
+        if (doesKeyPairExist()) {
+            // Read key pair again
```
weijjia: remove the comment

src/src/com/microsoft/aad/adal/StorageHelper.java
```diff
@@ -305,70 +283,241 @@ public String decrypt(String value) throws GeneralSecurityException, IOException
...
+                // in this case getEntry throws
+                // java.lang.RuntimeException: error:0D07207B:asn1 encoding routines:ASN1_get_object:header too long
+                // handle it as regular KeyStoreException
+                ClientAnalytics.logEvent(new AndroidKeyStoreFailureEvent(
+                        new InstrumentationPropertiesBuilder(e)));
+                throw new KeyStoreException(e);
+            }
+        }
+        
+        throw new KeyStoreException("KeyPair entry does not exist.");
```
weijjia: move up the throw.

sdabbiru:
>             throw new IllegalArgumentException(String.format(

NIT: I know it is not related to the change you made. But, I think we should not be throwing IllegalArgumentException here. We should be throwing IllegalStateException.

---
Refers to: src/src/com/microsoft/aad/adal/StorageHelper.java:230 in 4f5f475. [](commit_id = 4f5f4752b65e633f7f06196502a7450da21b861f, deletion_comment = False)

src/src/com/microsoft/aad/adal/StorageHelper.java
```diff
@@ -305,70 +283,241 @@ public String decrypt(String value) throws GeneralSecurityException, IOException
...
+        if (doesKeyPairExist()) {
+            // Read key pair again
+            Logger.v(TAG, "Reading Key entry");
+            try {
+                KeyStore keyStore = KeyStore.getInstance("AndroidKeyStore");
+                keyStore.load(null);
+                final KeyStore.PrivateKeyEntry entry = (KeyStore.PrivateKeyEntry)keyStore.getEntry(
+                        KEY_STORE_CERT_ALIAS, null);
+                return new KeyPair(entry.getCertificate().getPublicKey(), entry.getPrivateKey());
+            } catch (RuntimeException e) {
```
weijjia: should only wrap around getEntry

src/src/com/microsoft/aad/adal/StorageHelper.java
```diff
@@ -305,70 +283,241 @@ public String decrypt(String value) throws GeneralSecurityException, IOException
...
+        }
+        
+        throw new KeyStoreException("KeyPair entry does not exist.");
+    }
+    
+    /**
+     * Check if KeyPair exists on AndroidKeyStore. 
+     */
+    private synchronized boolean doesKeyPairExist() throws GeneralSecurityException, IOException {
+        KeyStore keyStore = KeyStore.getInstance("AndroidKeyStore");
```
weijjia: final

sdabbiru:
>             throw new IllegalArgumentException("Unexpected MAC length");

Should not be an illegalArg exception.

---
Refers to: src/src/com/microsoft/aad/adal/StorageHelper.java:526 in 4f5f475. [](commit_id = 4f5f4752b65e633f7f06196502a7450da21b861f, deletion_comment = False)

src/src/com/microsoft/aad/adal/StorageHelper.java
```diff
@@ -305,70 +283,241 @@ public String decrypt(String value) throws GeneralSecurityException, IOException
...
+     * Read KeyPair from AndroidKeyStore. 
+     */
+    private synchronized KeyPair readKeyPair() throws GeneralSecurityException, IOException {
+        if (doesKeyPairExist()) {
+            // Read key pair again
+            Logger.v(TAG, "Reading Key entry");
+            try {
+                KeyStore keyStore = KeyStore.getInstance("AndroidKeyStore");
+                keyStore.load(null);
+                final KeyStore.PrivateKeyEntry entry = (KeyStore.PrivateKeyEntry)keyStore.getEntry(
```
weijjia: Check to see if getEntry can call containsAlias, if it could call containAlias, we still need to catch the NullPointerException that could be thrown.

src/src/com/microsoft/aad/adal/StorageHelper.java
weijjia: Double check the whole file, make sure exception is thrown in the beginning.

src/src/com/microsoft/aad/adal/StorageHelper.java
weijjia: If we're throwing IOExceptiion, it will be bubbled up to the decrypt method where we've added the telemetry events, no need to add it here.

src/src/com/microsoft/aad/adal/StorageHelper.java
weijjia: Same as above, the IOException will be bubbled up, clientAnalystics is added in the caller.

src/src/com/microsoft/aad/adal/StorageHelper.java
weijjia: getEntry does also call containAlias, and the nullpointerexception could also potentially be thrown. We have a broader RuntimeException catch around getEntry, the benefit of doing the containAlias check before getEntry is we can catch those exceptions separately.

src/src/com/microsoft/aad/adal/StorageHelper.java
```diff
@@ -369,34 +367,31 @@ private synchronized SecretKey getKeyOrCreate(final String keyVersion) throws Ge
...
     private synchronized SecretKey getKey(final String keyVersion) throws GeneralSecurityException, IOException {
-        if (VERSION_USER_DEFINED.equals(keyVersion)) {
+        switch (keyVersion) {
+        case VERSION_USER_DEFINED : 
             return getSecretKey(AuthenticationSettings.INSTANCE.getSecretKeyData());
-        }
-        
-        if (VERSION_ANDROID_KEY_STORE.equals(keyVersion)) {
+        case VERSION_ANDROID_KEY_STORE :
             if (Build.VERSION.SDK_INT >= 18) {
```
RandalliLama: @weijjia I thought that we decided we didn't need this check.  We should never have a case where there is a key in the key store on a device with an OS below 18.

src/src/com/microsoft/aad/adal/StorageHelper.java
```diff
@@ -426,36 +421,36 @@ private synchronized KeyPair generateKeyPairFromAndroidKeyStore() throws General
...
+        }
+
+        Logger.v(TAG, "Reading Key entry");
+        final KeyStore keyStore = KeyStore.getInstance("AndroidKeyStore");
+        keyStore.load(null);
+
+        try {
+            final KeyStore.PrivateKeyEntry entry = (KeyStore.PrivateKeyEntry)keyStore.getEntry(
+                    KEY_STORE_CERT_ALIAS, null);
+            return new KeyPair(entry.getCertificate().getPublicKey(), entry.getPrivateKey());
```
RandalliLama: @weijjia Does this line throw RuntimeException as well?  If so, exactly which api can throw?

weijjia: ah, right. Only getEntry should can throw, the new Keypair should be moved out of try/catch.

src/src/com/microsoft/aad/adal/StorageHelper.java
```diff
@@ -632,23 +618,30 @@ private SecretKey unwrap(Cipher wrapCipher, byte[] keyBlob) throws GeneralSecuri
...
+        
         Logger.v(TAG, "Reading key data from a file");
-        final InputStream in = new FileInputStream(file);
-
+        final InputStream in = new FileInputStream(keyFile);
         try {
-            ByteArrayOutputStream bytes = new ByteArrayOutputStream();
-            byte[] buffer = new byte[1024];
+            final ByteArrayOutputStream bytes = new ByteArrayOutputStream();
+            final byte[] buffer = new byte[1024];
```
RandalliLama: @weijjia By the way, why do we have the magic number 1024 here?  Could we make this a class local constant with a comment perhaps?

src/src/com/microsoft/aad/adal/StorageHelper.java
weijjia: ah ok. Sorry, I missed it. Yeah, we talked about it, it's not possible for user to downgrade their device, so this is not likely to happen. Will update.

src/src/com/microsoft/aad/adal/StorageHelper.java
weijjia: Well, it's assuming the byte array size is never more than 1kb, This is a reasonable assume, the wrapped key we store in the file shouldn't be over 1kb, and it saves us from reading the file size.

RandalliLama: @weijjia I am ok with this, but I think it would still be good to make it a class local constant.

weijjia: Sure, will update.

RandalliLama:
@weijjia :shipit:

addev-ashish:
:shipit:

weijjia:
#624

PR #608, opened 2016-04-24
update datetime format for pasing datestring if it does not contaim A…

weijjia:
1. Current datetime string parsing fails when it date string doesn't contain AM or PM
2. fix the test compilation with recent changes in dev

### DIFF
--- a/src/src/com/microsoft/aad/adal/DateTimeAdapter.java
+++ b/src/src/com/microsoft/aad/adal/DateTimeAdapter.java
@@ -84,6 +84,16 @@ public final class DateTimeAdapter implements JsonDeserializer<Date>, JsonSerial
             JsonDeserializationContext context) throws JsonParseException {
         String jsonString = json.getAsString();
 
+        // Datetime string is serialized with iso8601 format by default, should
+        // always try to deserialize with iso8601. But to support the backward
+        // compatibility, we also need to deserialize with old format if failing
+        // with iso8601. 
+        try {
+            return iso8601Format.parse(jsonString);
+        } catch (final ParseException ignored) {
+        }
+        
+        // fallback logic for old date format parsing. 
         try {
             return localFormat.parse(jsonString);
         } catch (final ParseException ignored) {
@@ -101,12 +111,7 @@ public final class DateTimeAdapter implements JsonDeserializer<Date>, JsonSerial
 
         try {
             return enUs24HourFormat.parse(jsonString);
-        } catch (final ParseException ignored) {
-        }
-
-        try {
-            return iso8601Format.parse(jsonString);
-        } catch (ParseException e) {
+        } catch (final ParseException e) {
             Logger.e(TAG, "Could not parse date: " + e.getMessage(), "",
                     ADALError.DATE_PARSING_FAILURE, e);
         }
--- a/src/src/com/microsoft/aad/adal/DateTimeAdapter.java
+++ b/src/src/com/microsoft/aad/adal/DateTimeAdapter.java
@@ -50,11 +50,30 @@ public final class DateTimeAdapter implements JsonDeserializer<Date>, JsonSerial
             DateFormat.DEFAULT);
 
     private final DateFormat iso8601Format = buildIso8601Format();
+    private final DateFormat enUs24HourFormat = buildEnUs24HourDateFormat();
+    private final DateFormat local24HourFormat = buildLocal24HourDateFormat();
 
     private static DateFormat buildIso8601Format() {
         DateFormat iso8601Format = new SimpleDateFormat("yyyy-MM-dd'T'HH:mm:ss'Z'", Locale.US);
         iso8601Format.setTimeZone(TimeZone.getTimeZone("UTC"));
         return iso8601Format;
+    }
+    
+    /**
+     * Add new en-us date format for parsing date string if it doesn't contain AM/PM.
+     */
+    private static DateFormat buildEnUs24HourDateFormat() {
+        DateFormat newformat = new SimpleDateFormat("MMM dd, yyyy HH:mm:ss", Locale.US);
+        return newformat;
+    }
+    
+    /**
+     * Add new local date format when parsing date string if it doesn't contain AM/PM.
+     * @return
+     */
+    private static DateFormat buildLocal24HourDateFormat() {
+        DateFormat newformat = new SimpleDateFormat("MMM dd, yyyy HH:mm:ss", Locale.getDefault());
+        return newformat;
     }
 
     public DateTimeAdapter() {
@@ -67,12 +86,22 @@ public final class DateTimeAdapter implements JsonDeserializer<Date>, JsonSerial
 
         try {
             return localFormat.parse(jsonString);
-        } catch (ParseException ignored) {
+        } catch (final ParseException ignored) {
+        }
+        
+        try {
+            return local24HourFormat.parse(jsonString);
+        } catch (final ParseException ignored) {
         }
 
         try {
             return enUsFormat.parse(jsonString);
-        } catch (ParseException ignored) {
+        } catch (final ParseException ignored) {
+        }
+
+        try {
+            return enUs24HourFormat.parse(jsonString);
+        } catch (final ParseException ignored) {
         }
 
         try {
--- a/tests/Functional/src/com/microsoft/aad/adal/test/DefaultTokenCacheStoreTests.java
+++ b/tests/Functional/src/com/microsoft/aad/adal/test/DefaultTokenCacheStoreTests.java
@@ -27,7 +27,6 @@ import static org.mockito.Mockito.mock;
 import static org.mockito.Mockito.when;
 
 import java.io.IOException;
-import java.lang.reflect.Field;
 import java.security.GeneralSecurityException;
 import java.security.NoSuchAlgorithmException;
 import java.util.Calendar;
@@ -40,6 +39,8 @@ import java.util.TimeZone;
 
 import javax.crypto.NoSuchPaddingException;
 
+import org.mockito.Mockito;
+
 import com.microsoft.aad.adal.AuthenticationSettings;
 import com.microsoft.aad.adal.CacheKey;
 import com.microsoft.aad.adal.DefaultTokenCacheStore;
@@ -51,9 +52,6 @@ import com.microsoft.aad.adal.TokenCacheItem;
 import android.app.Activity;
 import android.content.Context;
 import android.content.SharedPreferences;
-import android.content.pm.PackageManager.NameNotFoundException;
-
-import org.mockito.Mockito;
 
 public class DefaultTokenCacheStoreTests extends BaseTokenStoreTests {
 
@@ -73,7 +71,7 @@ public class DefaultTokenCacheStoreTests extends BaseTokenStoreTests {
         super.tearDown();
     }
 
-    public void testSharedCache() throws GeneralSecurityException, IOException {
+    public void testCacheItemRetrieval() throws GeneralSecurityException, IOException {
         TokenCacheItem item = mockDefaultCacheStore("Apr 28, 2015 1:09:57 PM").getItem("testkey");
 
         // Verify returned item
@@ -104,7 +102,7 @@ public class DefaultTokenCacheStoreTests extends BaseTokenStoreTests {
         assertNotNull(item.getExpiresOn().after(new Date()));
     }
     
-    public void testDateTimeFormatterOldFormatWithoutAMOrPm() throws GeneralSecurityException, IOException {
+    public void testDateTimeFormatterOldFormat24hourFormat() throws GeneralSecurityException, IOException {
         TokenCacheItem item = mockDefaultCacheStore("Apr 28, 2015 13:09:57").getItem("testkey");
 
         // Verify returned item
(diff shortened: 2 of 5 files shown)


weijjia:
@imakhalova FYI

src/src/com/microsoft/aad/adal/DateTimeAdapter.java
```diff
@@ -50,12 +50,31 @@
             DateFormat.DEFAULT);
 
     private final DateFormat iso8601Format = buildIso8601Format();
+    private final DateFormat enUs24HourFormat = buildEnUs24HourDateFormat();
+    private final DateFormat local24HourFormat = buildLocal24HourDateFormat();
 
     private static DateFormat buildIso8601Format() {
         DateFormat iso8601Format = new SimpleDateFormat("yyyy-MM-dd'T'HH:mm:ss'Z'", Locale.US);
```
imakhalova: Should we consider this format for default locale?

src/src/com/microsoft/aad/adal/DateTimeAdapter.java
```diff
@@ -50,12 +50,31 @@
...
         iso8601Format.setTimeZone(TimeZone.getTimeZone("UTC"));
         return iso8601Format;
     }
+    
+    /**
+     * Add new en-us date format for parsing date string if it doesn't contain AM/PM.
+     */
+    private static DateFormat buildEnUs24HourDateFormat() {
+        DateFormat newformat = new SimpleDateFormat("MMM dd, yyyy HH:mm:ss", Locale.US);
```
imakhalova: Have you asked server guys which formats do they support and return?

tests/Functional/src/com/microsoft/aad/adal/test/DefaultTokenCacheStoreTests.java
```diff
@@ -74,7 +74,7 @@ protected void tearDown() throws Exception {
     }
 
     public void testSharedCache() throws GeneralSecurityException, IOException {
```
imakhalova: how does it test shared cache?

tests/Functional/src/com/microsoft/aad/adal/test/WebRequestHandlerTests.java
```diff
@@ -106,7 +106,11 @@ public void testNullUrl() {
         assertThrowsException(IllegalArgumentException.class, "url", new Runnable() {
```
imakhalova: Use ThrowableRunnable, then no need in try-catch block

tests/Functional/src/com/microsoft/aad/adal/test/MockWebRequestHandler.java
```diff
@@ -68,9 +71,8 @@ public HttpWebResponse sendPost(URL url, Map<String, String> headers, byte[] con
             try {
                 mRequestContent = new String(content, "UTF-8");
             } catch (UnsupportedEncodingException e) {
```
imakhalova: this method throws IOException, no need to catch  UnsupportedEncodingException

src/src/com/microsoft/aad/adal/DateTimeAdapter.java
weijjia: server return both expires in and expires on(will be milliseconds since January 1, 1970, 00:00:00 GMT)

tests/Functional/src/com/microsoft/aad/adal/test/DefaultTokenCacheStoreTests.java
weijjia: I think what this test intends to test is, if the cache is shared, then the returned client id should be as expected. It doesn't actually test cache is shared.

tests/Functional/src/com/microsoft/aad/adal/test/WebRequestHandlerTests.java
```diff
@@ -162,26 +162,25 @@ public void testClientTraceInHeaders() {
...
+        } catch (final IOException e) {
+            assertTrue(e instanceof UnknownHostException);
+            assertTrue(e.getMessage().toLowerCase(Locale.US).contains("unable to resolve host"));
+        }
     }
 
-    public void testGetWithIdRequest() {
+    public void testGetWithIdRequest() throws IOException {
         WebRequestHandler request = new WebRequestHandler();
         HttpWebResponse httpResponse = request.sendGet(getUrl(TEST_WEBAPI_URL + "/1"), null);
```
weijjia: All the tests related to this TEST_WEBAPI_URL is failing now since the webapi is no longer reachable. I'm not going to update those tests in this PR, will send a separate one to address them.

src/src/com/microsoft/aad/adal/DateTimeAdapter.java
weijjia: why?

src/src/com/microsoft/aad/adal/DateTimeAdapter.java
```diff
@@ -50,12 +50,31 @@
             DateFormat.DEFAULT);
 
     private final DateFormat iso8601Format = buildIso8601Format();
```
imakhalova: Did you figure out why serialize works with iso8601Format , but produces different format (with AM/PM and w/o) ?

src/src/com/microsoft/aad/adal/DateTimeAdapter.java
imakhalova: It seems like other formats has 2 different locales, but this one works only with US locale?

src/src/com/microsoft/aad/adal/DateTimeAdapter.java
weijjia: iso8601 has the format yyyy-MM-dd'T'HH:mm:ss'Z', it use 24 hour, UTC Z.

tests/Functional/src/com/microsoft/aad/adal/test/DefaultTokenCacheStoreTests.java
imakhalova: Can you rename it to something more appropriate, for me it looks like testGetItem

src/src/com/microsoft/aad/adal/DateTimeAdapter.java
imakhalova: I mean all dates for tokens in cache should have this format yyyy-MM-dd'T'HH:mm:ss'Z' (because we serialize them with this format only)
How do we get tokens with date format like "MMM dd, yyyy HH:mm:ss" ?

tests/Functional/src/com/microsoft/aad/adal/test/WebRequestHandlerTests.java
```diff
@@ -190,11 +189,10 @@ public void testPostRequest() {
         try {
             httpResponse = request.sendPost(getUrl(TEST_WEBAPI_URL), null,
                     json.getBytes(ENCODING_UTF8), "application/json");
-        } catch (UnsupportedEncodingException e) {
-            Assert.fail("Encoding exception is not expected");
+        } catch (final IOException e) {
```
imakhalova: just do not catch anything, you test throws exception

tests/Functional/src/com/microsoft/aad/adal/test/WebRequestHandlerTests.java
```diff
@@ -162,26 +162,25 @@ public void testClientTraceInHeaders() {
...
-                getUrl("http://www.somethingabcddnotexists.com"), null);
-        assertNotNull(httpResponse.getResponseException());
-        assertTrue("Unknown host exception",
-                httpResponse.getResponseException() instanceof UnknownHostException);
-        assertTrue("Unable to resolve host", httpResponse.getResponseException().getMessage()
-                .toLowerCase(Locale.US).contains("unable to resolve host"));
+        try {
+            request.sendGet(
+                    getUrl("http://www.somethingabcddnotexists.com"), null);
+        } catch (final IOException e) {
```
imakhalova: assertfail (if exception wasn't thrown)

tests/Functional/src/com/microsoft/aad/adal/test/WebRequestHandlerTests.java
imakhalova: Sorry, I excluded this file from my project since these tests fail anyway

src/src/com/microsoft/aad/adal/DateTimeAdapter.java
imakhalova: And why don't we use locale Root to serialize Date?

tests/Functional/src/com/microsoft/aad/adal/test/DefaultTokenCacheStoreTests.java
weijjia: Sure, will update it.

src/src/com/microsoft/aad/adal/DateTimeAdapter.java
weijjia: This also bothers me. I can see why it happens in broker, broker always serialize datetime using default dateformat, different locale may cause the problem. But for DefaultTokenCache, the dateTimeAdapter is already introduced since 1.1.4. The only reason I can see why it happens is, user is only an old version of i.e. one drive, so they have existing tokens with old format. Then they update to an newer version, we start seeing the failures, does it sound reasonable?

tests/Functional/src/com/microsoft/aad/adal/test/WebRequestHandlerTests.java
weijjia: yeah, my bad, should already have it removed.

tests/Functional/src/com/microsoft/aad/adal/test/WebRequestHandlerTests.java
weijjia: good catch, will add.

tests/Functional/src/com/microsoft/aad/adal/test/WebRequestHandlerTests.java
weijjia: I'll send a separate PR to mock the response.

src/src/com/microsoft/aad/adal/DateTimeAdapter.java
weijjia: I actually don't think it matters here. What we eventually return back to caller is Date, specific moment in time, with milliseconds in precision, and always UTC. We we need here is just a consistent way to persist this data when serializing/deserializing it. Before 1.1.14, i think we simply use the default format, and default locale. The two different locals is to support backward compatibility.

imakhalova: Please notify broker app about issues that caused by non providing DateTimeAdapter,
it means that if user switches his default date format, he will be signed out
This must be fixed, this DateTimeAdapter is public, so they can reuse it

src/src/com/microsoft/aad/adal/DateTimeAdapter.java
```diff
@@ -67,12 +86,22 @@ public synchronized Date deserialize(JsonElement json, Type typeOfT,
...
-        } catch (ParseException ignored) {
+        } catch (final ParseException ignored) {
+        }
+
+        try {
+            return enUs24HourFormat.parse(jsonString);
+        } catch (final ParseException ignored) {
         }
 
         try {
```
imakhalova: please change the order and start with iso8601Format, because it should be the most popular

src/src/com/microsoft/aad/adal/DateTimeAdapter.java
weijjia: It's not app team actually. Broker sdk is part of wpj library, Satheesh is already aware of it. And we're fixing it.

src/src/com/microsoft/aad/adal/DateTimeAdapter.java
weijjia: Yup, I agree. It should be changed.

src/src/com/microsoft/aad/adal/DateTimeAdapter.java
imakhalova: Users have tokens that are older than 1 year?)) Then do we really care if they will be able to parse date or not? we will remove this token anyway because it's expired
anyway you won't be able to cover all possible date formats that user could set to default, but I think no harm going to be done with adding this couple date formats

imakhalova:
signed off

src/src/com/microsoft/aad/adal/DateTimeAdapter.java
weijjia: Users could stay on old version of onedrive, a version that's one year old. I don't expect so, but it could happen, and they update it, i.e. recently, Then I can understand that the exception is thrown. Otherwise, i don't know why we get the exception in onedrive. If you get more data around it, feel free to let me know. Or I can play with your telemetry data.

weijjia:
@heidijinxujia Your turn :)

src/src/com/microsoft/aad/adal/DateTimeAdapter.java
imakhalova: But the dates that failed on parsing are very recent (t least all that are in telemetry which we added recently are pretty fresh). I'll send you the query for this type of error

weijjia: hmm, sure, send me the query.

PR #610, opened 2016-04-28
Avoid dead lock if sync call is invoked in main thread

imakhalova:
There is a problem in CallbackHandler: currently it takes handler only for main thread and it means it invokes callbacks always on Main Thread (Of cause it's totally incorrect and needs to be fixed). 
But if we use the same interface for sync call it actually caused deadlock on main thread. We can’t invoke callback on main thread which waits for the callback to be invoked.

Of cause we shouldn't allow to call any sync methods in main thread because it's doing network call and it's not allowed to make network call on Main thread (it must throw NetworkOnMainThreadException).
But it was allowed in prev adal versions and needs to be supported til next major release.

### DIFF
--- a/src/src/com/microsoft/aad/adal/AuthenticationContext.java
+++ b/src/src/com/microsoft/aad/adal/AuthenticationContext.java
@@ -61,7 +61,10 @@ import android.net.ConnectivityManager;
 import android.net.NetworkInfo;
 import android.os.Bundle;
 import android.os.Handler;
+import android.os.Looper;
+import android.os.NetworkOnMainThreadException;
 import android.support.v4.content.LocalBroadcastManager;
+import android.util.Log;
 import android.util.SparseArray;
 
 /**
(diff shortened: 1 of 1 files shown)


src/src/com/microsoft/aad/adal/AuthenticationContext.java
```diff
@@ -605,7 +608,25 @@ public AuthenticationResult acquireTokenSilentSync(String resource, String clien
...
+            throw new IllegalArgumentException("resource");
+        }
+        if (StringExtensions.IsNullOrBlank(clientId)) {
+            throw new IllegalArgumentException("clientId");
+        }
+
+        final AuthenticationRequest request = new AuthenticationRequest(mAuthority, resource,
+                clientId, userId, getRequestCorrelationId());
+        request.setSilent(true);
+        request.setPrompt(PromptBehavior.Auto);
```
agamemnum: We can't allow prompt right?

imakhalova: When it's silent we can't. But there is no PromptBehavior.NoPrompt, only auto or always

weijjia:
Have an offline talk with @imakhalova, she already tested this. @addev-ashish and @RandalliLama I'll merge the change then.

PR #612, opened 2016-05-01
Fixing_SSLException_Handling

sathyas38:
Fixing the SSL Hand Shake Failure Exception fields.

When there is a SSL_HANDSHAKE_FAILURE in the AuthenticationActivity the library is throwing an ADALException with error code set to SERVER_INVALID_REQUEST and the actual error code of (WebViewClient.ERROR_FAILED_SSL_HANDSHAKE) being put in a formatted string in the exception message. Since the ERROR_FAILED_SSL_HANDSHAKE is a defined error in the list of ADALError, making the change to set the ADALException error code to ADALError.ERROR_FAILED_SSL_HANDSHAKE so that it can be better handled at the Application code consuming it.

### DIFF
--- a/src/src/com/microsoft/aad/adal/AuthenticationContext.java
+++ b/src/src/com/microsoft/aad/adal/AuthenticationContext.java
@@ -51,6 +51,7 @@ import android.os.Bundle;
 import android.os.Handler;
 import android.support.v4.content.LocalBroadcastManager;
 import android.util.SparseArray;
+import android.webkit.WebViewClient;
 
 /**
  * ADAL context to get access token, refresh token, and lookup from cache.
@@ -755,8 +756,14 @@ public class AuthenticationContext {
                             .getString(AuthenticationConstants.Browser.RESPONSE_ERROR_MESSAGE);
                     Logger.v(TAG, "Error info:" + errCode + " " + errMessage + " for requestId: "
                             + requestId + correlationInfo);
+                    
+                    ADALError adalError = ADALError.SERVER_INVALID_REQUEST;
+                    if (errCode.equalsIgnoreCase("Code:" + String.valueOf(WebViewClient.ERROR_FAILED_SSL_HANDSHAKE))) {
+                    	adalError = ADALError.ERROR_FAILED_SSL_HANDSHAKE;
+                    }
+                    
                     waitingRequestOnError(waitingRequest, requestId, new AuthenticationException(
-                            ADALError.SERVER_INVALID_REQUEST, errCode + " " + errMessage));
+                            adalError, errCode + " " + errMessage));
                 } else if (resultCode == AuthenticationConstants.UIResponse.BROWSER_CODE_COMPLETE) {
                     final AuthenticationRequest authenticationRequest = (AuthenticationRequest)extras
                             .getSerializable(AuthenticationConstants.Browser.RESPONSE_REQUEST_INFO);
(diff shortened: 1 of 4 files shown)


src/src/com/microsoft/aad/adal/AuthenticationContext.java
```diff
@@ -755,8 +756,14 @@ public void onActivityResult(int requestCode, int resultCode, Intent data) {
                             .getString(AuthenticationConstants.Browser.RESPONSE_ERROR_MESSAGE);
                     Logger.v(TAG, "Error info:" + errCode + " " + errMessage + " for requestId: "
                             + requestId + correlationInfo);
+                    
+                    ADALError adalError = ADALError.SERVER_INVALID_REQUEST;
+                    if (errCode.equalsIgnoreCase("Code:" + String.valueOf(WebViewClient.ERROR_FAILED_SSL_HANDSHAKE))) {
```
imakhalova: put ADAL error code into bundle and just use it,
no need to parse and compare strings, when enums are serializable

sathyas38:
Incorporated comments.

src/src/com/microsoft/aad/adal/AuthenticationContext.java
```diff
@@ -755,8 +756,15 @@ public void onActivityResult(int requestCode, int resultCode, Intent data) {
                             .getString(AuthenticationConstants.Browser.RESPONSE_ERROR_MESSAGE);
                     Logger.v(TAG, "Error info:" + errCode + " " + errMessage + " for requestId: "
                             + requestId + correlationInfo);
+                    
+                    ADALError adalError = ADALError.SERVER_INVALID_REQUEST;
```
imakhalova: put this error for other resultIntent for BROWSER_CODE_ERROR
And you won't need to check it for null, because it will be required extras as authRequest

src/src/com/microsoft/aad/adal/AuthenticationContext.java
```diff
@@ -756,7 +757,7 @@ public void onActivityResult(int requestCode, int resultCode, Intent data) {
                     Logger.v(TAG, "Error info:" + errCode + " " + errMessage + " for requestId: "
                             + requestId + correlationInfo);
                     waitingRequestOnError(waitingRequest, requestId, new AuthenticationException(
-                            ADALError.SERVER_INVALID_REQUEST, errCode + " " + errMessage));
+                    		(ADALError)data.getSerializableExtra(AuthenticationConstants.ADAL_ERROR_CODE), errCode + " " + errMessage));
```
addev-ashish: There are tabs here, could you please change it to spaces

weijjia:
@sathyas38 Thanks for the contribution. 
We understand that today we simply send back the top level error code SERVER_INVALID_REQUEST it's hard for developer to tell what's actually going on. Even though the error message itself is containing the original error code returned, we shouldn't have developers to parse the error message and get the original error code. 

However this has been an agreement we have with client app, there may be developers already taking dependency on the specific error code, we cannot simply replace it with another error code. It will be a breaking change for the sdk then. The way we propose to fix the problem is to introduce the inner exception, which will be added as the cause of the AuthenticationException we throw. The "cause" will contain the originally thrown error code and error message. Developers can easily get the "cause" of the exception we send back today, and then get the un-interpreted error code. This can also help us better categorize errors, i.e. when SERVER_INVALID_REQUEST is sent back, it conveys the info that "something wrong is happening when interacting with webview", and inner cause will tell the detailed reason.

PR #484, opened 2016-01-29
check the permissions of Account Manager at the beginning of the broker flow

heidijinxujia:
1. Check the permissions of Account Manager at the beginning of the broker flow, so that the lib could detect early if the app has the necessary permissions. If the app does not have the permission, it will fail instead falling back to the local flow.

2.Adjusted the format several chunks of code

Issue #475, opened 2016-01-19
ADAL should fail if permissions to use broker are not available.


RandalliLama:
If the developer has not opted out of the broker then we should fail as soon as possible if it would not be possible to invoke it.  That means we should detect early that the app does not have the necessary permissions to the Account Manager and fail, rather than falling back to the local flow.

heidijinxujia:
In the BrokerProxy, we check the permission to Account Manager when we want to acquire token after validation. For this issue, do you mean we need to bring forward the permission checking before the validation?

weijjia:
we should fail upfront if developer clearly choose to use broker.

heidijinxujia:
Shipped for review last week. Thanks!

addev-ashish:
looks like this was fixed.

weijjia:
Why we are closing this? is the PR merged or this is a duplicate issue?

PR #613, opened 2016-05-04
check the permissions of Account Manager at the beginning of the broker flow

heidijinxujia:
Check the permissions of Account Manager at the beginning of the broker flow, so that the lib could detect early if the app has the necessary permissions. If the app does not have the permission, it will fail instead falling back to the local flow.

Ref PR #484

### DIFF
--- a/samples/hello/src/com/microsoft/aad/adal/hello/LoginFragment.java
+++ b/samples/hello/src/com/microsoft/aad/adal/hello/LoginFragment.java
@@ -20,6 +20,7 @@ import com.microsoft.aad.adal.AuthenticationContext;
 import com.microsoft.aad.adal.AuthenticationResult;
 import com.microsoft.aad.adal.IWindowComponent;
 import com.microsoft.aad.adal.PromptBehavior;
+import com.microsoft.aad.adal.UsageAuthenticationException;
 
 public class LoginFragment extends Fragment {
 
@@ -79,7 +80,12 @@ public class LoginFragment extends Fragment {
             CookieManager cookieManager = CookieManager.getInstance();
             cookieManager.removeAllCookie();
             CookieSyncManager.getInstance().sync();
-            mAuthContext.getCache().removeAll();
+            try {
+				mAuthContext.getCache().removeAll();
+			} catch (final UsageAuthenticationException e) {
+				// TODO Auto-generated catch block
+				e.printStackTrace();
+			}
         } else {
             // login
             mAuthContext.acquireToken(wrapFragment(LoginFragment.this), Constants.RESOURCE_ID,
--- a/samples/hello/src/com/microsoft/aad/adal/hello/MainActivity.java
+++ b/samples/hello/src/com/microsoft/aad/adal/hello/MainActivity.java
@@ -51,6 +51,7 @@ import com.microsoft.aad.adal.AuthenticationContext;
 import com.microsoft.aad.adal.AuthenticationResult;
 import com.microsoft.aad.adal.Logger;
 import com.microsoft.aad.adal.PromptBehavior;
+import com.microsoft.aad.adal.UsageAuthenticationException;
 
 public class MainActivity extends Activity {
 
@@ -281,12 +282,17 @@ public class MainActivity extends Activity {
     }
 
     public void onClickClearTokens(View view) {
-        if (mAuthContext != null && mAuthContext.getCache() != null) {
-            displayMessage("Clearing tokens");
-            mAuthContext.getCache().removeAll();
-        } else {
-            textView1.setText("Cache is null");
-        }
+        try {
+			if (mAuthContext != null && mAuthContext.getCache() != null) {
+			    displayMessage("Clearing tokens");
+			    mAuthContext.getCache().removeAll();
+			} else {
+			    textView1.setText("Cache is null");
+			}
+		} catch (final UsageAuthenticationException e) {
+			// TODO Auto-generated catch block
+			e.printStackTrace();
+		}
     }
 
     public void onClickClearCookies(View view) {
--- a/src/src/com/microsoft/aad/adal/ApplicationReceiver.java
+++ b/src/src/com/microsoft/aad/adal/ApplicationReceiver.java
@@ -96,13 +96,17 @@ public class ApplicationReceiver extends BroadcastReceiver {
                     // 1) there is saved request in sharedPreference
                     // 2) app has the correct configuration to get token from broker
                     // 3) the saved request is not timeout
-                    if (!StringExtensions.IsNullOrBlank(request) && brokerProxy.canSwitchToBroker() 
-                            && isRequestTimestampValidForResume(dateTimeForSavedRequest)) {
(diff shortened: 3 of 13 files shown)


heidijinxujia:
Respond to the code reviews and update the corresponding unit test.

heidijinxujia:
@weijjia Please have a look. I made a new sub-branch from the current dev and coded on it. Because the previous PR is too old and after I merging the dev there are a lot of incompatible code in it.

samples/hello/src/com/microsoft/aad/adal/hello/LoginFragment.java
```diff
@@ -79,7 +80,12 @@ public void buttonClicked(View view) {
             CookieManager cookieManager = CookieManager.getInstance();
             cookieManager.removeAllCookie();
             CookieSyncManager.getInstance().sync();
-            mAuthContext.getCache().removeAll();
+            try {
+				mAuthContext.getCache().removeAll();
```
weijjia: indent, please don't use tab

weijjia: check across files.

samples/hello/src/com/microsoft/aad/adal/hello/LoginFragment.java
```diff
@@ -79,7 +80,12 @@ public void buttonClicked(View view) {
...
+			} catch (final UsageAuthenticationException e) {
+				// TODO Auto-generated catch block
```
weijjia: removed the TODO

src/src/com/microsoft/aad/adal/BrokerProxy.java
```diff
@@ -124,38 +127,83 @@ public boolean verifyUser(String username, String uniqueid) {
...
-                && PackageManager.PERMISSION_GRANTED == pm.checkPermission(
-                        "android.permission.MANAGE_ACCOUNTS", mContext.getPackageName())
-                && PackageManager.PERMISSION_GRANTED == pm.checkPermission(
-                        "android.permission.USE_CREDENTIALS", mContext.getPackageName());
-        if (!permission) {
-            Logger.w(TAG, "Broker related permissions are missing for GET_ACCOUNTS, MANAGE_ACCOUNTS, USE_CREDENTIALS",
+    private boolean verifyManifestPermissions() throws UsageAuthenticationException {
+        final String methodName = ":verifyManifestPermissions";
+        final PackageManager packageManager = mContext.getPackageManager();
+        List<String> permerssionMissing = new ArrayList<String>();
```
weijjia: like I mentioned in your closed PR, I don't think you need to create a list for this.

weijjia: also nit: final

src/src/com/microsoft/aad/adal/BrokerProxy.java
```diff
@@ -124,38 +127,83 @@ public boolean verifyUser(String username, String uniqueid) {
...
+        Logger.v(
+                TAG + methodName,
+                "Broker related permissions checking starts.");       
+               
+        if(PackageManager.PERMISSION_GRANTED != packageManager.checkPermission(
```
weijjia: nit: space after if

tests/Functional/src/com/microsoft/aad/adal/test/AuthenticationContextTest.java
```diff
@@ -2402,9 +2415,14 @@ private ITokenCacheStore addItemToCache(ITokenCacheStore cache, String token,
     }
 
     private void clearCache(AuthenticationContext context) {
```
weijjia: just throw here

addev-ashish:
#475

heidijinxujia:
Under broker:
o	If target version is lower than 23, calling app has to have the following permissions declared in manifest(http://developer.android.com/reference/android/accounts/AccountManager.html): 
	GET_ACCOUNTS
	USE_CREDENTIALS
	MANAGE_ACOUNTS

o	If target version is 23, USE_CREDENTIALS and MANAGE_ACCOUNTS are already deprecated. But GET_ACCOUNTS is under protection level "dangerous", calling app is responsible for requesting the run-time permisson. You can reference Runtime permission request for API 23.

src/src/com/microsoft/aad/adal/ApplicationReceiver.java
```diff
@@ -96,13 +96,17 @@ public void onReceive(Context context, Intent intent) {
...
+                        if (!StringExtensions.IsNullOrBlank(request) && brokerProxy.canSwitchToBroker() 
+                                && isRequestTimestampValidForResume(dateTimeForSavedRequest)) {
+                            Logger.v(TAG + methodName, receivedInstalledPackageName + " is installed, start sending request to broker.");
+                            resumeRequestInBroker(context, request);
+                        } else {
+                            Logger.v(TAG + methodName, "No request saved in sharedpreferences or request already timeout"
+                                    + ", cannot resume broker request.");
+                        }
+                    } catch (final UsageAuthenticationException e) {
+                        Logger.v(TAG + methodName, e.getMessage());
```
RandalliLama: @heidijinxujia Is there a reason we are throwing the exception away here?  Which method might throw this and under what circumstances?

samples/hello/src/com/microsoft/aad/adal/hello/MainActivity.java
```diff
@@ -281,12 +282,17 @@ public void onClickUseToken(View view) {
...
+        try {
+			if (mAuthContext != null && mAuthContext.getCache() != null) {
+			    displayMessage("Clearing tokens");
+			    mAuthContext.getCache().removeAll();
+			} else {
+			    textView1.setText("Cache is null");
+			}
+		} catch (final UsageAuthenticationException e) {
+			// TODO Auto-generated catch block
+			e.printStackTrace();
```
RandalliLama: @heidijinxujia Does this actually log?  Why are you throwing the exception away.  Exactly which method can throw this and under what circumstances?

RandalliLama:
@heidijinxujia Can we move this check to near the beginning of acquireToken somewhere.  It should happen around the same time that we do normal parameter validation.  Then we would be able to just let the exception go back to the caller.

heidijinxujia:
Thanks for your reviews, Rich. I am working on it.

src/src/com/microsoft/aad/adal/ApplicationReceiver.java
heidijinxujia: This exception is thrown by canSwitchToBroker() which calls the permission checking. And this method onReceive() is an override method which does not throw exceptions. This is the reason why we need to catch the exception here.

samples/hello/src/com/microsoft/aad/adal/hello/MainActivity.java
heidijinxujia: The UsageAuthenticationException is thrown by getCache() which calls canSwitchToBroker(). Because in getCache(), if the broker is used, it will return cache implementation related to broker so that app can clear tokens for related accounts. In addition, according to your comment here, I will change the printStackTrace into logger.

Issue #614, opened 2016-05-05
Runtime exceptions thrown by keystore


weijjia:
Following are the top crashes thrown by keystore currently reported by outlook:

- 1. 

```
java.lang.IllegalStateException: could not generate key in keystore
       at android.security.AndroidKeyPairGenerator.generateKeyPair(AndroidKeyPairGenerator.java:99)
       at java.security.KeyPairGenerator$KeyPairGeneratorImpl.generateKeyPair(KeyPairGenerator.java:276)
       at com.microsoft.aad.adal.StorageHelper.getKeyPairFromAndroidKeyStore(StorageHelper.java:488)
       at com.microsoft.aad.adal.StorageHelper.getSecretKeyFromAndroidKeyStore(StorageHelper.java:416)
       at com.microsoft.aad.adal.StorageHelper.getKeyForVersion(StorageHelper.java:190)
       at com.microsoft.aad.adal.StorageHelper.decrypt(StorageHelper.java:259)
       at com.microsoft.aad.adal.DefaultTokenCacheStore.decrypt(DefaultTokenCacheStore.java:127)
       at com.microsoft.aad.adal.DefaultTokenCacheStore.getItem(DefaultTokenCacheStore.java:145)
       at com.microsoft.aad.adal.AuthenticationContext.getItemFromCache(AuthenticationContext.java:1503)
       at com.microsoft.aad.adal.AuthenticationContext.localFlow(AuthenticationContext.java:1381)
       at com.microsoft.aad.adal.AuthenticationContext.acquireTokenAfterValidation(AuthenticationContext.java:1373)
       at com.microsoft.aad.adal.AuthenticationContext.acquireTokenLocalCall(AuthenticationContext.java:1200)
       at com.microsoft.aad.adal.AuthenticationContext.access$600(AuthenticationContext.java:62)
       at com.microsoft.aad.adal.AuthenticationContext$6.run(AuthenticationContext.java:1159)
       at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1112)
       at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:587)
       at java.lang.Thread.run(Thread.java:818)
```

- 2. 

```
java.lang.NullPointerException: Attempt to invoke interface method 'int android.security.IKeystoreService.exist(java.lang.String, int)' on a null object reference
[...]

PR #615, opened 2016-05-05
(HotFix)catch reported runtime exceptions thrown by keystore and re-throw as checked exceptions to prevent app from crashing

weijjia:
Outlook is reporting three top crashes thrown by keystore, recorded in issue #614 
With the current limitations with Keystore, which could be wiped with screen lock type changing. Also, if we try to access keystore when it's locked, behavior is also undefined. 

To avoid app from crashing because of those issues, we need to catch those individual runtime exceptions, and re-throw as checked exceptions.

### DIFF
--- a/src/src/com/microsoft/aad/adal/StorageHelper.java
+++ b/src/src/com/microsoft/aad/adal/StorageHelper.java
@@ -40,7 +40,6 @@ import java.security.KeyStore;
 import java.security.KeyStoreException;
 import java.security.MessageDigest;
 import java.security.NoSuchAlgorithmException;
-import java.security.PrivateKey;
 import java.security.SecureRandom;
 import java.security.UnrecoverableKeyException;
 import java.security.cert.CertificateException;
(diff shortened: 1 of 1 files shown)


src/src/com/microsoft/aad/adal/StorageHelper.java
```diff
@@ -485,23 +484,53 @@ private void deleteKeyFile() {
...
-            generator.initialize(getKeyPairGeneratorSpec(mContext, start.getTime(), end.getTime()));
-            generator.generateKeyPair();
-            Logger.v(TAG, "Key entry is generated");
-        } else {
-            Logger.v(TAG, "Key entry is available");
+        
+        try {
+            if (!keyStore.containsAlias(KEY_STORE_CERT_ALIAS)) {
+                Logger.v(TAG, "Key entry is not available");
+                Calendar start = Calendar.getInstance();
```
RandalliLama: Since NullPointerException is such a broad and dangerous exception I'd like to catch it only around the functions that we think might throw it due to a bug.  I think it might be better to call keyStore.containsAlias with in its own try catch, save the result and then use it in an 'if' statement.  The you can have a second try{}catch(NullPointerException) wrapped only around the keystore methods.  You would then be able to exclude the date and logging functions from the try catch.

RandalliLama:
:shipit:

src/src/com/microsoft/aad/adal/StorageHelper.java
```diff
@@ -485,7 +484,24 @@ private void deleteKeyFile() {
...
+        } catch (final NullPointerException exception) {
+            // There is an issue with Android Keystore when remote service attempts 
+            // to access Keystore. 
+            // Changeset found for google source to address the related issue with 
+            // remote service accessing keystore :
+            // https://android.googlesource.com/platform/external/sepolicy/+/0e30164b17af20f680635c7c6c522e670ecc3df3
+            // The thrown exception in this case is:
+            // java.lang.NullPointerException: Attempt to invoke interface method 
+            // 'int android.security.IKeystoreService.exist(java.lang.String, int)' on a null object reference
+            // To avoid app from crashing, re-throw as checked exception
```
RPangrle: 👍

RPangrle:
:shipit:

sdabbiru:
:shipit:

PR #616, opened 2016-05-05
Telemetry: add the reason why auth token wasn't obtained

imakhalova:
Let's add property why auth token wasn't obtained

### DIFF
--- a/src/src/com/microsoft/aad/adal/AuthenticationContext.java
+++ b/src/src/com/microsoft/aad/adal/AuthenticationContext.java
@@ -31,7 +31,6 @@ import java.net.URLEncoder;
 import java.security.NoSuchAlgorithmException;
 import java.util.Date;
 import java.util.Iterator;
-import java.util.Map;
 import java.util.UUID;
 import java.util.concurrent.Callable;
 import java.util.concurrent.CountDownLatch;
--- a/src/src/com/microsoft/aad/adal/AuthenticationContext.java
+++ b/src/src/com/microsoft/aad/adal/AuthenticationContext.java
@@ -31,6 +31,7 @@ import java.net.URLEncoder;
 import java.security.NoSuchAlgorithmException;
 import java.util.Date;
 import java.util.Iterator;
+import java.util.Map;
 import java.util.UUID;
 import java.util.concurrent.Callable;
 import java.util.concurrent.CountDownLatch;
@@ -1338,6 +1339,8 @@ public class AuthenticationContext {
                 } catch (AuthenticationException ex) {
                     // pass back to caller for known exceptions such as failure
                     // to encrypt
+                    ClientAnalytics.logEvent(new RefreshTokenEvent(
+                            new InstrumentationPropertiesBuilder(request, ex), InstrumentationIDs.EVENT_RESULT_FAIL, true));
                     callbackHandle.onError(ex);
                     return null;
                 }
@@ -1348,6 +1351,9 @@ public class AuthenticationContext {
             if (result != null && result.getAccessToken() != null
                     && !result.getAccessToken().isEmpty()) {
                 Logger.v(TAG, "Token is returned from background call ");
+                ClientAnalytics.logEvent(new RefreshTokenEvent(
+                        new InstrumentationPropertiesBuilder(request, result), InstrumentationIDs.EVENT_RESULT_SUCCESS, true));
+
                 callbackHandle.onSuccess(result);
                 return result;
             }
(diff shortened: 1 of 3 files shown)


src/src/com/microsoft/aad/adal/AuthenticationContext.java
```diff
@@ -1453,17 +1454,22 @@ private AuthenticationResult localFlow(CallbackHandler callbackHandle,
...
         // refresh token does not exist or refresh token request failed to give back the token. 
         // If it's non-silent request, will try to acquire token interactively. 
         // if it's silent request, will return the AUTH_REFRESH_FAILED_PROMPT_NOT_ALLOWED.
         if (refreshItem == null || authResult == null 
                 || (authResult != null && StringExtensions.IsNullOrBlank(authResult.getAccessToken()))) {
             Logger.v(TAG, "Refresh token is not available or refresh token request failed to return token.");
+            Map<String, String> eventProperties = new InstrumentationPropertiesBuilder(request, authResult)
+                    .add(InstrumentationIDs.ERROR_CLASS, refreshItem == null ?
+                            InstrumentationIDs.REFRESH_TOKEN_NOT_FOUND :
+                            authResult == null ? InstrumentationIDs.AUTH_RESULT_EMPTY : InstrumentationIDs.AUTH_RESULT_CONTAINS_ERROR)
```
sdabbiru: AUTH_RESULT_CONTAINS_ERROR cannot be assumed. In this case, authResult does not contain an access token.

imakhalova: ok, let's rename it to NO_ACCESS_TOKEN_IN_RESPONSE. how does it sound?

src/src/com/microsoft/aad/adal/AuthenticationContext.java
```diff
@@ -1396,6 +1397,14 @@ private AuthenticationResult acquireTokenAfterValidation(CallbackHandler callbac
                             ADALError.DEVELOPER_ACTIVITY_IS_NOT_RESOLVED));
                 }
             } else {
+                Map<String, String> eventProperties = new InstrumentationPropertiesBuilder(request, result)
+                        .add(InstrumentationIDs.IS_BROKER_APP, Boolean.TRUE.toString())
+                        .add(InstrumentationIDs.ERROR_CLASS,
+                                result == null ? InstrumentationIDs.AUTH_RESULT_EMPTY : InstrumentationIDs.NO_ACCESS_TOKEN_IN_RESPONSE)
+                        .build();
+                ClientAnalytics.logEvent(
+                        InstrumentationIDs.AUTH_TOKEN_NOT_RETURNED,
```
agamemnum: Irina that would be hard to use - let's use event name "TokenRequest" and two properties "Result":"Failed", "Reason":"AUTH_TOKEN_NOT_RETURNED". 

Overall do not create - event type for each condition. Instead design event types around scenario.

src/src/com/microsoft/aad/adal/InstrumentationIDs.java
```diff
@@ -34,6 +34,11 @@
     /* Event invoked when auth token was successfully obtained */
     static final String REFRESH_TOKEN_REQUEST_SUCCEEDED = "RefreshTokenRequestSucceeded";
 
+    /* Error message that notifies that auth token wasn't obtained because there was no refresh token found*/
+    static final String REFRESH_TOKEN_NOT_FOUND = "RefreshTokenNotFound";
+    static final String AUTH_RESULT_EMPTY = "EmptyResponseFromServer";
+    static final String NO_ACCESS_TOKEN_IN_RESPONSE = "HasResponseFromServer";
```
weijjia: hm, NO_ACCESS_TOKEN_IN_RESPONSE = hasResponseFromServer? :)

src/src/com/microsoft/aad/adal/AuthenticationContext.java
imakhalova: sounds reasonable, changing

src/src/com/microsoft/aad/adal/AuthenticationContext.java
```diff
@@ -31,6 +31,7 @@
 import java.security.NoSuchAlgorithmException;
 import java.util.Date;
 import java.util.Iterator;
+import java.util.Map;
 import java.util.UUID;
```
addev-ashish: this import seems to be redundant now

addev-ashish:
Is there an issue filed for this PR?

addev-ashish:
:clock1:

imakhalova:
The issue that users complain that they got prompt for password too frequently and we need a way to collect statistics why this happens.

src/src/com/microsoft/aad/adal/AuthenticationContext.java
imakhalova: right, removing

addev-ashish:
If there is no issue filed yet, i think its ok to leave this one as is. Going forward we want to keep a mapping between PR's and issues, so that we know the context in which the fixes were made in.

weijjia:
:shipit:

PR #635, opened 2016-05-20
Fix the IdToken related errors in unit tests

heidijinxujia:
The change of the IdToken constructor cause the failures in several tests which correlated to the IdToken and OAuth class. This PR is aiming at fix this bug.


RandalliLama:
Looks ok to me. Love the removal of reflection.

PR #640, opened 2016-05-24
b2c fixes for reading profile info and tokencache store

weijjia:
1. ProfileInfo should only contain version, sub, name, preferredname and tenant id
2. when storing token cache item into cache, only store one token item. Also update the hashcode calculation for scopes. 
3. typo on the id_token_expires_in

### DIFF
--- a/samples/hello/src/com/microsoft/aad/adal/hello/Constants.java
+++ b/samples/hello/src/com/microsoft/aad/adal/hello/Constants.java
@@ -32,16 +32,17 @@ public class Constants {
     
     // AAD PARAMETERS
     // https://login.windows.net/tenantInfo
-    static final String AUTHORITY_URL = "https://login.windows-ppe.net/common";
+    static final String AUTHORITY_URL = "https://login.microsoftonline.com/common";// "https://login.windows-ppe.net/common";
 
     // Clientid is given from AAD page when you register your Android app
-    static final String CLIENT_ID = "b67b2169-6966-4597-b3e4-f698dc5df864";
+    static final String CLIENT_ID = "a7d8cef0-4145-49b2-a91d-95c54051fa3f";// "b67b2169-6966-4597-b3e4-f698dc5df864";
 
     // RedirectUri
     static final String REDIRECT_URL = "urn:ietf:wg:oauth:2.0:oob";// "msauth://com.example.userapp2/WiEb9EPiRwdK5vgH%2FQPajYeR%2F%2Bs%3D";
 
     // URI for the resource. You need to setup this resource at AAD
-    static final String[] SCOPE = new String[]{"https://outlook.office.com/Mail.Read"};
+    static final String[] SCOPE = new String[] { "User.Read" };// new
+                                                               // String[]{"https://outlook.office.com/Mail.Read"};
 
     // Endpoint we are targeting for the deployed WebAPI service
     static final String SERVICE_URL = "https://android.azurewebsites.net/api/values";
--- a/samples/hello/src/com/microsoft/aad/adal/hello/Constants.java
+++ b/samples/hello/src/com/microsoft/aad/adal/hello/Constants.java
@@ -32,17 +32,16 @@ public class Constants {
     
     // AAD PARAMETERS
     // https://login.windows.net/tenantInfo
-    static final String AUTHORITY_URL = "https://login.microsoftonline.com/common";// "https://login.windows-ppe.net/common";
+    static final String AUTHORITY_URL = "app-authority";
 
     // Clientid is given from AAD page when you register your Android app
-    static final String CLIENT_ID = "a7d8cef0-4145-49b2-a91d-95c54051fa3f";// "b67b2169-6966-4597-b3e4-f698dc5df864";
+    static final String CLIENT_ID = "app-client-id";
 
     // RedirectUri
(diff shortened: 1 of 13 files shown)


samples/hello/src/com/microsoft/aad/adal/hello/Constants.java
```diff
@@ -32,16 +32,17 @@
     
     // AAD PARAMETERS
     // https://login.windows.net/tenantInfo
-    static final String AUTHORITY_URL = "https://login.windows-ppe.net/common";
+    static final String AUTHORITY_URL = "https://login.microsoftonline.com/common";// "https://login.windows-ppe.net/common";
 
     // Clientid is given from AAD page when you register your Android app
-    static final String CLIENT_ID = "b67b2169-6966-4597-b3e4-f698dc5df864";
+    static final String CLIENT_ID = "a7d8cef0-4145-49b2-a91d-95c54051fa3f";// "b67b2169-6966-4597-b3e4-f698dc5df864";
```
kpanwar: we should not have any app information in the library.

samples/hello/src/com/microsoft/aad/adal/hello/Constants.java
```diff
@@ -32,16 +32,17 @@
...
     // RedirectUri
     static final String REDIRECT_URL = "urn:ietf:wg:oauth:2.0:oob";// "msauth://com.example.userapp2/WiEb9EPiRwdK5vgH%2FQPajYeR%2F%2Bs%3D";
 
     // URI for the resource. You need to setup this resource at AAD
-    static final String[] SCOPE = new String[]{"https://outlook.office.com/Mail.Read"};
+    static final String[] SCOPE = new String[] { "User.Read" };// new
+                                                               // String[]{"https://outlook.office.com/Mail.Read"};
 
     // Endpoint we are targeting for the deployed WebAPI service
     static final String SERVICE_URL = "https://android.azurewebsites.net/api/values";
```
kpanwar: same comment from above.

src/src/com/microsoft/aad/adal/AuthenticationConstants.java
```diff
@@ -135,6 +137,16 @@
         static final String ID_TOKEN_PASSWORD_CHANGE_URL = "pwd_url";
 
         public static final Object PROFILE_INFO = "profile_info";
+        
+        static final String OPEN_ID = "openid";
+
+        static final class PROFILE_INFO_CLAIM {
```
kpanwar: each class should be in a separate file.

src/src/com/microsoft/aad/adal/TokenCacheKey.java
```diff
@@ -14,317 +16,309 @@
...
+
+    private String mPolicy = "";
+
+    private TokenCacheKey() {
+    }
+
+    public String toJsonString() throws JSONException {
+        JSONObject obj = new JSONObject();
+        obj.put("a", mAuthority);
+        obj.put("s", StringExtensions.createStringFromArray(mScope, " "));
```
kpanwar: would the cache lookup work if scope order is swapped?

samples/hello/src/com/microsoft/aad/adal/hello/Constants.java
weijjia: oops...my bad, should get them removed.

src/src/com/microsoft/aad/adal/TokenCacheKey.java
weijjia: why does it matter in toJsonString?
Check getHashCodeForScope, mScope is converted to the treeset as we discussed, and we get the hashcode using the treeset.

src/src/com/microsoft/aad/adal/AuthenticationContext.java
```diff
@@ -1354,14 +1354,17 @@ private void setItemToCache(final AuthenticationRequest request, AuthenticationR
...
-            
-            // Record idtoken with clientid as scope
-            if(request.isIdTokenRequest()){
-                cacheItem.setScope(new String[]{request.getClientId()});
-                key.setScope(new String[]{request.getClientId()});
-                mTokenCacheStore.setItem(key, cacheItem);
+
+            // If request contains open_id, convert it to client id.
+            final String[] scopesInRequest = request.getScope();
+            if (scopesInRequest != null && scopesInRequest.length == 1
```
kpanwar: why would this massive if ever matter? In createAuthenticationResult you are putting correct values for scope already.

src/src/com/microsoft/aad/adal/ProfileInfo.java
```diff
@@ -18,26 +18,61 @@
...
 package com.microsoft.aad.adal;
 
+import java.util.HashMap;
+
+import android.util.Base64;
+
 class ProfileInfo {
+    static final String TAG = ProfileInfo.class.getSimpleName();
+
+    String mVersion;
```
kpanwar: final?

kpanwar:
:shipit: with build issues resolved.

src/src/com/microsoft/aad/adal/ProfileInfo.java
weijjia: can we leave it for now? The class needs more refactor, but not related to the fix.

src/src/com/microsoft/aad/adal/AuthenticationContext.java
weijjia: this is for the request itself, make sure we have the correct scope in the key itself.

PR #647, opened 2016-06-01
Suppress javadoc lint warnings for test files 

RandalliLama:
and enable 'unchecked' lint warnings.

### DIFF
--- a/config/checkstyle/suppressions.xml
+++ b/config/checkstyle/suppressions.xml
@@ -7,5 +7,5 @@
     "http://www.puppycrawl.com/dtds/suppressions_1_1.dtd">
 
 <suppressions>
-    <suppress checks="JavadocStyleCheck" files=".*Test.java"/>
+    <suppress checks="Javadoc.*" files=".*Test.java"/>
 </suppressions>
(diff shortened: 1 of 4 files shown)


weijjia:
@RandalliLama Checkstyle task is failing for travis. We need to either disable the task for now or fix all the style issues.

config/checkstyle/suppressions.xml
```diff
@@ -1,32 +1,11 @@
...
-  ~ limitations under the License.
-  -->
+
 
 <!DOCTYPE suppressions PUBLIC
     "-//Puppy Crawl//DTD Suppressions 1.1//EN"
     "http://www.puppycrawl.com/dtds/suppressions_1_1.dtd">
 
 <suppressions>
-    <suppress checks="[a-zA-Z0-9]*" files="R.java" />
```
weijjia: why are we removing it? R file is auto-generated, we don't need to run checkstyle on auto-generated file.

RandalliLama: @weijjia I removed all of these suppressions because they appear to have been copy and pasted from someone else's file.  I will add them back in as necessary, but for now I am not seeing any warnings for R files.

weijjia: Got it~ I'm still in the preference of not running checkstyle on auto-generated files even it's not failing now. It should not fail actually. :)

weijjia:
:shipit:

PR #651, opened 2016-06-02
fix failing tests for default token cache store

weijjia:
Mockito is having problem for mocking package visible methods. For the failed three tests, all we want to mock is the decrypted item, we don't even need to mock the behavior of loadSecretKeyForEncryption. Also, the current existing test is having the mock behavior to return null, there is no need to do it.


addev-ashish:
:shipit:

PR #653, opened 2016-06-06
A batch of fixes for checkstyle issues, mostly in tests.

RandalliLama:
This is an incomplete set of fixes.  However, I am running out of time and need to get these in as a batch before they become obsolete.  I managed to reduce the total number of warnings from ~1600 to under 600.


adal/src/androidTest/java/com/microsoft/aad/adal/OauthTests.java
```diff
@@ -73,7 +77,7 @@ public void testParseIdTokenPositive() throws UnsupportedEncodingException, Auth
         assertEquals("emailField", actual.getEmail());
         assertEquals("idpProvider", actual.getIdentityProvider());
         assertEquals("53c6acf2-2742-4538-918d-e78257ec8516", actual.getObjectId());
-        assertTrue(1387227772 == actual.getPasswordExpiration());
+        assertTrue(Util.TEST_PASSWORD_EXPIRATION == actual.getPasswordExpiration());
```
addev-ashish: can we flip the l-value and the r-value here, everywhere else in this file constants are on the right side.

addev-ashish:
>     }

nit: this function is not used.

---
Refers to: adal/src/androidTest/java/com/microsoft/aad/adal/HashMapExtensionTests.java:115 in 29c212e. [](commit_id = 29c212e1f827932499ccd6b9f14e0e8b25d5931f, deletion_comment = False)

adal/src/androidTest/java/com/microsoft/aad/adal/DefaultTokenCacheStoreTests.java
```diff
@@ -122,12 +122,12 @@ protected StorageHelper getStorageHelper() {
...
 
-    
+
     public void testDateTimeFormatterLocaleChange() throws AuthenticationException {
-        DefaultTokenCacheStore store = (DefaultTokenCacheStore)setupItems();
+        DefaultTokenCacheStore store = (DefaultTokenCacheStore) setupItems();
         List<TokenCacheItem> tokens = store.getTokensForResource("resource");
         // Serializing without miliseconds
-        long precision = 1000;
+        final long precision = 1000;
```
addev-ashish: since you did this. All the DefaultTokenCacheStore objects in the rest of this file can be final too.

adal/src/androidTest/java/com/microsoft/aad/adal/BrokerProxyTests.java
```diff
@@ -90,26 +90,26 @@ protected void setUp() throws Exception {
...
         // until it finds the correct one for ADAL broker.
         for (Signature signature : info.signatures) {
-            testSignature = signature.toByteArray();
+            mTestSignature = signature.toByteArray();
             MessageDigest md = MessageDigest.getInstance("SHA");
-            md.update(testSignature);
-            testTag = Base64.encodeToString(md.digest(), Base64.NO_WRAP);
+            md.update(mTestSignature);
+            mTestTag = Base64.encodeToString(md.digest(), Base64.NO_WRAP);
             break;
```
addev-ashish: since you are updating this, you can remove the for loop as we break at the end.

addev-ashish:
>     private static final int HTTP_OK = 200;

nit: consider [https://developer.android.com/reference/java/net/HttpURLConnection.html#HTTP_OK](https://developer.android.com/reference/java/net/HttpURLConnection.html#HTTP_OK)

---
Refers to: adal/src/androidTest/java/com/microsoft/aad/adal/AuthenticationContextTest.java:93 in 29c212e. [](commit_id = 29c212e1f827932499ccd6b9f14e0e8b25d5931f, deletion_comment = False)

adal/src/androidTest/java/com/microsoft/aad/adal/AuthenticationActivityUnitTest.java
```diff
@@ -67,13 +67,19 @@
...
+    private static final int TEST_REQUEST_ID2 = 166;
+
     private static final long CONTEXT_REQUEST_TIME_OUT = 20000;
 
     private static final long DEVICE_RESPONSE_WAIT = 500;
 
-    private Intent intentToStartActivity;
+    private static final int TEST_CALLING_UID = 333;
+
+    private static final int HTTP_OK = 200;
```
addev-ashish: same as previous file. we could use HTTPUrlConnection.HTTP_OK

adal/src/androidTest/java/com/microsoft/aad/adal/AcquireTokenSilentHandlerTest.java
```diff
@@ -45,93 +45,94 @@
...
+
     static final String TEST_IDTOKEN_USERID = "4f859989-a2ff-411e-9048-c322247ac62c";
-    
+
     static final String TEST_IDTOKEN_UPN = "admin@aaltests.onmicrosoft.com";
-    
+
+    static final int HTTP_OK = 200;
+    static final int HTTP_BAD_REQUEST = 400;
+    static final int HTTP_INTERNAL_SERVER_ERROR = 500;
```
weijjia: we should use HttpUrlConnection.HTTP_OK, HTTP_BAD_REQUEST, HTTP_INTERNAL_ERROR instead.

weijjia:
>     private class TestCacheItem {

I don't think I understand the change. Why we're creating the TestCacheItem? 

---
Refers to: adal/src/androidTest/java/com/microsoft/aad/adal/AuthenticationContextTest.java:2347 in 29c212e. [](commit_id = 29c212e1f827932499ccd6b9f14e0e8b25d5931f, deletion_comment = False)

adal/src/androidTest/java/com/microsoft/aad/adal/BrokerProxyTests.java
RandalliLama: @addev-ashish I don't feel comfortable making that change in this PR.  The comment and the code do not align and I don't know which is correct.  I've opened issue #657 to track the need to fix this.

RandalliLama:
@weijjia The change to add TestCacheItem was to satisfy a checkstyle rule that there shouldn't be more than 10 parameters to a method, which seems like a reasonable limit to me.

adal/src/androidTest/java/com/microsoft/aad/adal/OauthTests.java
RandalliLama: @addev-ashish fixed.

adal/src/androidTest/java/com/microsoft/aad/adal/DefaultTokenCacheStoreTests.java
RandalliLama: @addev-ashish fixed

addev-ashish:
>     }

deleted

---
In reply to: [224102518](https://github.com/AzureAD/azure-activedirectory-library-for-android/pull/653#issuecomment-224102518) [](ancestors = 224102518)

---
Refers to: adal/src/androidTest/java/com/microsoft/aad/adal/HashMapExtensionTests.java:115 in 29c212e. [](commit_id = 29c212e1f827932499ccd6b9f14e0e8b25d5931f, deletion_comment = False)

weijjia:
:shipit:

addev-ashish:
:shipit:

PR #658, opened 2016-06-08
Permission checking for Account Manager when broker is used

heidijinxujia:
Check the permissions of Account Manager at the beginning of acquireToken() if when broker is used, so that the lib could detect early if the app has the necessary permissions. If the app does not have the permission, it will throw the UsageAuthenticationException to the callback.

Ref PR #613

Under broker:
o If target version is lower than 23, calling app has to have the following permissions declared in manifest (http://developer.android.com/reference/android/accounts/AccountManager.html): 
 GET_ACCOUNTS
 USE_CREDENTIALS
 MANAGE_ACOUNTS

Beginning in Android 6.0 (API level 23), users grant permissions to apps while the app is running, not when they install the app. 

o If target version is 23, USE_CREDENTIALS and MANAGE_ACCOUNTS are already deprecated. But GET_ACCOUNTS is under protection level "dangerous", calling app is responsible for requesting the run-time permisson. You can reference Runtime permission request for API 23.

### DIFF
--- a/adal/src/main/java/com/microsoft/aad/adal/AuthenticationContext.java
+++ b/adal/src/main/java/com/microsoft/aad/adal/AuthenticationContext.java
@@ -41,6 +41,8 @@ import java.util.concurrent.FutureTask;
 import java.util.concurrent.atomic.AtomicReference;
 import java.util.concurrent.locks.Lock;
 import java.util.concurrent.locks.ReentrantReadWriteLock;
+import java.util.List;
+import java.util.ArrayList;
 
 import javax.crypto.NoSuchPaddingException;
 
@@ -59,6 +61,7 @@ import android.content.pm.PackageManager;
 import android.content.pm.ResolveInfo;
 import android.net.ConnectivityManager;
 import android.net.NetworkInfo;
+import android.os.Build;
 import android.os.Bundle;
 import android.os.Handler;
 import android.os.Looper;
@@ -583,6 +586,15 @@ public class AuthenticationContext {
 
         if (StringExtensions.IsNullOrBlank(redirectUri)) {
             redirectUri = getRedirectFromPackage();
+        }
+
+        //check the permissions needed if broker is used
+        if(AuthenticationSettings.INSTANCE.getUseBroker() == true) {
+            try {
+                verifyManifestPermissions();
+            } catch (UsageAuthenticationException exception) {
+                callback.onError(exception);
+            }
         }
 
         return redirectUri;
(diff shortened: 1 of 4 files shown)


weijjia:
@heidijinxujia Are we going to close the other one? Also, as we talked yesterday, are you going to move back the reflection tests and did a refactor later when favoring local cache PR is merged?

adal/src/main/java/com/microsoft/aad/adal/AuthenticationContext.java
```diff
@@ -1780,6 +1792,65 @@ private void checkInternetPermission() {
...
+                        "", ADALError.DEVELOPER_BROKER_PERMISSIONS_MISSING);
+            }
+
+            Logger.v(
+                    TAG + methodName,
+                    "Broker related permissions are verified");
+
+            if(permissionMissing.length() == 0) {
+                return true;
+            } else {
```
addev-ashish: not needed

adal/src/main/java/com/microsoft/aad/adal/AuthenticationContext.java
```diff
@@ -1780,6 +1792,65 @@ private void checkInternetPermission() {
...
+                throw new UsageAuthenticationException(
+                        ADALError.DEVELOPER_BROKER_PERMISSIONS_MISSING,
+                        "Broker related permissions are missing for " +
+                                permissionMissing.toString());
+            }
+        }
+        //this line will not be reached.
+        return false;
```
addev-ashish: i don't understand this?

adal/src/main/java/com/microsoft/aad/adal/AuthenticationContext.java
```diff
@@ -1780,6 +1792,65 @@ private void checkInternetPermission() {
     }
 
     /**
+     * To verify if App gives permissions to AccountManager to use broker.
+     *
+     * @throws UsageAuthenticationException
+     * @return boolean If the GET_ACCOUNTS, MANAGE_ACCOUNTS, USE_CREDENTIALS permissions are granted in the Manifest.xml
+     */
+    private boolean verifyManifestPermissions() throws UsageAuthenticationException {
+        if(Build.VERSION.SDK_INT <= 22) {
```
addev-ashish: don't like this, why 22? at a minimum add detailed comments.

addev-ashish:
Please merge your fix with Wei's and resubmit this PR.

adal/src/main/java/com/microsoft/aad/adal/AuthenticationContext.java
weijjia: on API 23 and above, if user don't request GET_ACCOUNTS permission, then are we allowing it?

adal/src/main/java/com/microsoft/aad/adal/AuthenticationContext.java
heidijinxujia: Do you mean the if-else is not needed? I think we should throw the exception if permission is missing, shouldn't we?

weijjia: there is no need to have a return here, just do a verify, if it fails, just throw.

PR #661, opened 2016-06-10
B2c fix cache item

weijjia:
For convergence, we store single entry. If we have RT in the cache, we'll try to use the RT. The token cache key used to find the RT either contains the displayableId or unique userId. However, when we initially get the token, the key we stored into cache contains both(If server returns to us, or null).


kpanwar:
:shipit:

Issue #570, opened 2016-04-01
Upgrade the SDK library to version 23 


heidijinxujia:
We are currently on v21, and will update the sdk to 23 in the future release.

PR #665, opened 2016-06-15
ADAL builds with marshmallow SDK

addev-ashish:
Changing the build.gradle files to use android sdk version 23
Apache library is not available by default, the testapp needs it. Adding a flag in testapp's build.gradle to use apache library. Issue #570

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,6 +7,8 @@ jdk:
 
 android:
   components:
+    - platform-tools
+    - tools
     - build-tools-23.0.3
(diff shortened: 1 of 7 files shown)


weijjia:
:shipit:

weijjia:
@addev-ashish onReceivedError in webvewclient is deprecated in API 23, since we already update to target 23, we should not use deprecated method in our code. Do a separate PR or revert this one and put in that fix. :)

PR #667, opened 2016-06-20
Update the signature validation for broker app

weijjia:
Update the logic for validating broker app signature. The two broker apps, Company Portal and Azure Authenticator are all only singed by Microsoft certs, and each of them only has one signature. The sdk will change to only accept one signature for broker app, and it has to be the one that the sdk trust.


weijjia:
@RandalliLama is fine with the change, @addev-ashish and @heidijinxujia I'm going to cherrypick the commit into another servicing branch branched off master and do a release.

Issue #677, opened 2016-07-11
Cleanup the thrown exceptions in the tests


weijjia:
In the tests, we should remove the exceptions from the throw part in the method signatures itself. We remove a couple of the reflection usage in the tests, but some of them are still left in the tests signature .

PR #678, opened 2016-07-11
cleanup up the exceptions in the tests files

weijjia:
Issue #677 
Cleanup the thrown exceptions in the tests which are no longer thrown with the removal of reflections.


addev-ashish:
:shipit:

PR #680, opened 2016-07-11
update suppressed test:testCreateFromResourceUrlPositive as external…

weijjia:
#issue 679
testCreateFromResourceUrlPositive was suppressed due to the external service is down, Enable the test again by mocking server response.


addev-ashish:
:shipit:

Issue #695, opened 2016-08-11
Outlook crash from NullPointerException during silent flow


weijjia:
Stacktrace from outlook. During silent flow, if the token item we store doesn't contain the refresh token, null pointer will be thrown when building the post request message with grant type as refresh_token. 

java.lang.NullPointerException: Attempt to invoke virtual method 'int java.lang.String.length()' on a null object reference
      at libcore.net.UriCodec.encode(UriCodec.java:132)
      at java.net.URLEncoder.encode(URLEncoder.java:57)
      at com.microsoft.aad.adal.StringExtensions.URLFormEncode(StringExtensions.java:83)
      at com.microsoft.aad.adal.Oauth2.buildRefreshTokenRequestMessage(Oauth2.java:177)
      at com.microsoft.aad.adal.Oauth2.refreshToken(Oauth2.java:302)
      at com.microsoft.aad.adal.AcquireTokenSilentHandler.acquireTokenWithRefreshToken(AcquireTokenSilentHandler.java:126)
      at com.microsoft.aad.adal.AcquireTokenSilentHandler.acquireTokenWithCachedItem(AcquireTokenSilentHandler.java:269)
      at com.microsoft.aad.adal.AcquireTokenSilentHandler.useMRRT(AcquireTokenSilentHandler.java:261)
      at com.microsoft.aad.adal.AcquireTokenSilentHandler.tryMRRT(AcquireTokenSilentHandler.java:203)
      at com.microsoft.aad.adal.AcquireTokenSilentHandler.tryRT(AcquireTokenSilentHandler.java:161)
      at com.microsoft.aad.adal.AcquireTokenSilentHandler.getAccessToken(AcquireTokenSilentHandler.java:104)
      at com.microsoft.aad.adal.AuthenticationContext.localFlow(AuthenticationContext.java:1461)
      at com.microsoft.aad.adal.AuthenticationContext.acquireTokenAfterValidation(AuthenticationContext.java:1448)
      at com.microsoft.aad.adal.AuthenticationContext.acquireTokenLocalCall(AuthenticationContext.java:1261)
      at com.microsoft.aad.adal.AuthenticationContext.access$900(AuthenticationContext.java:73)
      at com.microsoft.aad.adal.AuthenticationContext$7.run(AuthenticationContext.java:1220)
      at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1113)
[...]

PR #696, opened 2016-08-11
don't continue with rt request if the token cache item doesn't contain RT

weijjia:
#695 Outlook is seeing from the Nullpointer exception during silent request. 
To prevent app from crashing, we don't continue the silent request if the token cache item doesn't contain RT. Also, when reading the JSON response from server, we mark the flag isMRRT as true only when server returns resource as well as RT in the JSON response. 

Add unit tests to verify that if no rt exists in the cache, we return the null result.

### DIFF
--- a/adal/src/main/java/com/microsoft/aad/adal/Oauth2.java
+++ b/adal/src/main/java/com/microsoft/aad/adal/Oauth2.java
@@ -255,7 +255,9 @@ class Oauth2 {
                     expiresIn == null || expiresIn.isEmpty() ? AuthenticationConstants.DEFAULT_EXPIRATION_TIME_SEC
                             : Integer.parseInt(expiresIn));
 
-            if (response.containsKey(AuthenticationConstants.AAD.RESOURCE)) {
+            final String refreshToken = response.get(AuthenticationConstants.OAuth2.REFRESH_TOKEN);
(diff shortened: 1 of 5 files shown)


adal/src/main/java/com/microsoft/aad/adal/Oauth2.java
```diff
@@ -255,7 +255,9 @@ public static AuthenticationResult processUIResponseParams(Map<String, String> r
                     expiresIn == null || expiresIn.isEmpty() ? AuthenticationConstants.DEFAULT_EXPIRATION_TIME_SEC
                             : Integer.parseInt(expiresIn));
 
-            if (response.containsKey(AuthenticationConstants.AAD.RESOURCE)) {
+            final String refreshToken = response.get(AuthenticationConstants.OAuth2.REFRESH_TOKEN);
```
addev-ashish: >refreshToken [](start = 25, length = 12)

you can use this refreshToken in line 288

addev-ashish:
:shipit:

adal/src/main/java/com/microsoft/aad/adal/Oauth2.java
weijjia: done.

kpanwar:
:shipit: :rocket: :metal:

PR #698, opened 2016-08-11
Permission checking for Account Manager when broker is used

heidijinxujia:
Issue #475 
Check the permissions of Account Manager at the beginning of acquireToken() if when broker is used, so that the lib could detect early if the app has the necessary permissions. If the app does not have the permission, it will throw the UsageAuthenticationException to the callback.

Under broker:
o If target version is lower than 23, calling app has to have the following permissions declared in manifest (http://developer.android.com/reference/android/accounts/AccountManager.html): 
 GET_ACCOUNTS
 USE_CREDENTIALS
 MANAGE_ACOUNTS

Beginning in Android 6.0 (API level 23), users grant permissions to apps while the app is running, not when they install the app.
o If target version is 23, USE_CREDENTIALS and MANAGE_ACCOUNTS are already deprecated. But GET_ACCOUNTS is under protection level "dangerous", calling app is responsible for requesting the run-time permisson. You can reference Runtime permission request for API 23.

*Permission Checking Implementation Spec*

For Android version below M
For Android below 6.0 (API level 23), users grant permissions when they install the app (in the Manifest). So the function "CheckForPermissionAPI22orBelow" should do the following.
- check the GET_ACCOUNTS, USE_CREDENTIALS and MANAGE_ACOUNTS permissions upfront in acquireToken
- throw the UsageAuthenticationException to tell the developer that the permission need to be set in the manifest
- pass the exception into callBack

For Android.version >= M
Beginning in Android 6.0 (API level 23), users grant permissions to apps while the app is running, not when they install the app. So the function "CheckForPermissionAPI23orAbove" should do the following.
- check the permissions only when it is necessary, so we only check the GET_ACCOUNTS permission when broker is called
- check the permission at the end of canSwitchToBroker
- throw the UsageAuthenticationException and pass it to callback

### DIFF
--- a/adal/src/androidTest/java/com/microsoft/aad/adal/AcquireTokenRequestTest.java
+++ b/adal/src/androidTest/java/com/microsoft/aad/adal/AcquireTokenRequestTest.java
@@ -36,6 +36,7 @@ import android.content.pm.PackageInfo;
 import android.content.pm.PackageManager;
 import android.content.pm.ResolveInfo;
 import android.content.pm.Signature;
+import android.os.Build;
 import android.os.Bundle;
 import android.os.Handler;
 import android.test.AndroidTestCase;
@@ -780,18 +781,79 @@ public final class AcquireTokenRequestTest extends AndroidTestCase {
             cacheStore.removeAll();
         }
     }
-    
+
+    @SmallTest
+    public void testVerifyManifestPermissionMissingGetAccountsPermission() throws InterruptedException, PackageManager.NameNotFoundException {
+        final FileMockContext mockContext = createMockContext();
+        when(mockContext.getPackageManager().checkPermission(Mockito.refEq("android.permission.GET_ACCOUNTS"),
+                Mockito.anyString())).thenReturn(PackageManager.PERMISSION_DENIED);
+        AuthenticationSettings.INSTANCE.setUseBroker(true);
+
+        final AuthenticationContext authContext = new AuthenticationContext(mockContext,
+                VALID_AUTHORITY, false);
+        
+        final TestAuthCallback callback = new TestAuthCallback();
+        authContext.acquireToken(Mockito.mock(Activity.class), "resource", "clientid", authContext.getRedirectUriForBroker(),
+                "loginHint", callback);
+        final CountDownLatch signal = new CountDownLatch(1);
+        signal.await(ACTIVITY_TIME_OUT, TimeUnit.MILLISECONDS);
+
+        assertNotNull(callback.getCallbackException());
+        assertTrue(callback.getCallbackException() instanceof UsageAuthenticationException);
+        final UsageAuthenticationException usageAuthenticationException
+                = (UsageAuthenticationException) callback.getCallbackException();
+        assertTrue(usageAuthenticationException.getMessage().contains("GET_ACCOUNTS"));
+        assertEquals(ADALError.DEVELOPER_BROKER_PERMISSIONS_MISSING, usageAuthenticationException.getCode());
+    }
+
+    @SmallTest
+    public void testVerifyManifestPermissionMissingMultiPermissions() throws InterruptedException, PackageManager.NameNotFoundException {
+        if (Build.VERSION.SDK_INT < Build.VERSION_CODES.M) {
+            final FileMockContext mockContext = createMockContext();
+            when(mockContext.getPackageManager().checkPermission(Mockito.refEq("android.permission.GET_ACCOUNTS"),
+                    Mockito.anyString())).thenReturn(PackageManager.PERMISSION_GRANTED);
+            when(mockContext.getPackageManager().checkPermission(Mockito.refEq("android.permission.MANAGE_ACCOUNTS"),
+                    Mockito.anyString())).thenReturn(PackageManager.PERMISSION_DENIED);
+            when(mockContext.getPackageManager().checkPermission(Mockito.refEq("android.permission.USE_CREDENTIALS"),
+                    Mockito.anyString())).thenReturn(PackageManager.PERMISSION_DENIED);
+
+            AuthenticationSettings.INSTANCE.setUseBroker(true);
+            final AuthenticationContext authContext = new AuthenticationContext(mockContext,
+                    VALID_AUTHORITY, false);
+            final TestAuthCallback callback = new TestAuthCallback();
+            authContext.acquireToken(Mockito.mock(Activity.class), "resource", "clientid", authContext.getRedirectUriForBroker(),
+                    "loginHint", callback);
+            final CountDownLatch signal = new CountDownLatch(1);
+            signal.await(ACTIVITY_TIME_OUT, TimeUnit.MILLISECONDS);
+
+            assertNotNull(callback.getCallbackException());
+            assertTrue(callback.getCallbackException() instanceof UsageAuthenticationException);
+            final UsageAuthenticationException usageAuthenticationException
+                    = (UsageAuthenticationException) callback.getCallbackException();
+            assertTrue(usageAuthenticationException.getMessage().contains("MANAGE_ACCOUNTS"));
+            assertTrue(usageAuthenticationException.getMessage().contains("USE_CREDENTIALS"));
+            assertEquals(ADALError.DEVELOPER_BROKER_PERMISSIONS_MISSING, usageAuthenticationException.getCode());
+        }
+    }
+
     private FileMockContext createMockContext()
             throws PackageManager.NameNotFoundException {
 
         final FileMockContext mockContext = new FileMockContext(getContext());
 
+        mockContext.addPermission("android.permission.GET_ACCOUNTS");
+
+        if (Build.VERSION.SDK_INT < Build.VERSION_CODES.M) {
+            mockContext.addPermission("android.permission.MANAGE_ACCOUNTS");
+            mockContext.addPermission("android.permission.USE_CREDENTIALS");
+        }
+
         final AccountManager mockedAccountManager = getMockedAccountManager();
         mockContext.setMockedAccountManager(mockedAccountManager);
 
         final PackageManager mockedPackageManager = getMockedPackageManager();
-        mockContext.setMockedPackageManager(mockedPackageManager);
 
+        mockContext.setMockedPackageManager(mockedPackageManager);
         return mockContext;
     }
 
@@ -880,19 +942,20 @@ public final class AcquireTokenRequestTest extends AndroidTestCase {
 
         final PackageManager mockedPackageManager = Mockito.mock(PackageManager.class);
         when(mockedPackageManager.getPackageInfo(Mockito.anyString(), Mockito.anyInt())).thenReturn(mockedPackageInfo);
-        //mock permission
-        when(mockedPackageManager.checkPermission(Mockito.refEq("android.permission.GET_ACCOUNTS"),
-                Mockito.anyString())).thenReturn(PackageManager.PERMISSION_GRANTED);
-        when(mockedPackageManager.checkPermission(Mockito.refEq("android.permission.MANAGE_ACCOUNTS"),
-                Mockito.anyString())).thenReturn(PackageManager.PERMISSION_GRANTED);
-        when(mockedPackageManager.checkPermission(Mockito.refEq("android.permission.USE_CREDENTIALS"),
-                Mockito.anyString())).thenReturn(PackageManager.PERMISSION_GRANTED);
 
         // Mock intent query
         final List<ResolveInfo> activities = new ArrayList<>(1);
         activities.add(Mockito.mock(ResolveInfo.class));
         when(mockedPackageManager.queryIntentActivities(Mockito.any(Intent.class), Mockito.anyInt()))
                 .thenReturn(activities);
+
+        // Mock permissions
+        when(mockedPackageManager.checkPermission(Mockito.refEq("android.permission.GET_ACCOUNTS"),
+                Mockito.anyString())).thenReturn(PackageManager.PERMISSION_GRANTED);
+        when(mockedPackageManager.checkPermission(Mockito.refEq("android.permission.MANAGE_ACCOUNTS"),
+                Mockito.anyString())).thenReturn(PackageManager.PERMISSION_GRANTED);
+        when(mockedPackageManager.checkPermission(Mockito.refEq("android.permission.USE_CREDENTIALS"),
+                Mockito.anyString())).thenReturn(PackageManager.PERMISSION_GRANTED);
 
         return mockedPackageManager;
     }
--- a/adal/src/androidTest/java/com/microsoft/aad/adal/BrokerProxyTests.java
+++ b/adal/src/androidTest/java/com/microsoft/aad/adal/BrokerProxyTests.java
@@ -24,6 +24,7 @@
 package com.microsoft.aad.adal;
 
 import static org.mockito.Matchers.any;
+import static org.mockito.Matchers.anyInt;
 import static org.mockito.Matchers.anyString;
 import static org.mockito.Matchers.eq;
 import static org.mockito.Mockito.mock;
@@ -35,11 +36,13 @@ import java.lang.reflect.InvocationTargetException;
 import java.lang.reflect.Method;
 import java.security.MessageDigest;
 import java.security.NoSuchAlgorithmException;
+import java.util.ArrayList;
 import java.util.Calendar;
 import java.util.Date;
 import java.util.GregorianCalendar;
 import java.util.Locale;
 import java.util.UUID;
+import java.util.List;
 
 import org.mockito.Matchers;
 import org.mockito.Mockito;
@@ -61,6 +64,8 @@ import android.content.pm.PackageInfo;
 import android.content.pm.PackageManager;
 import android.content.pm.PackageManager.NameNotFoundException;
 import android.content.pm.Signature;
+import android.content.pm.ResolveInfo;
+import android.os.Build;
 import android.os.Bundle;
 import android.os.Handler;
 import android.test.AndroidTestCase;
@@ -112,14 +117,14 @@ public class BrokerProxyTests extends AndroidTestCase {
         String authenticatorType = AuthenticationConstants.Broker.BROKER_ACCOUNT_TYPE;
         String brokerPackage = "wrong";
         Signature signature = new Signature(mTestSignature);
-        prepareProxyForTest(brokerProxy, authenticatorType, brokerPackage, "test", signature, true, null);
+        prepareProxyForTest(brokerProxy, authenticatorType, brokerPackage, "test", signature, true, null, true);
 
         // action
         Method m = ReflectionUtils.getTestMethod(brokerProxy, "canSwitchToBroker");
-        boolean result = (Boolean) m.invoke(brokerProxy);
+        BrokerProxy.SwitchToBroker result = (BrokerProxy.SwitchToBroker) m.invoke(brokerProxy);
 
         // assert
-        assertFalse("verify should return false", result);
+        assertEquals(BrokerProxy.SwitchToBroker.CANNOT_SWITCH_TO_BROKER, result);
     }
 
     public void testCanSwitchToBrokerInvalidAuthenticatorType()
@@ -130,14 +135,14 @@ public class BrokerProxyTests extends AndroidTestCase {
         String authenticatorType = "invalid";
         String brokerPackage = AuthenticationConstants.Broker.COMPANY_PORTAL_APP_PACKAGE_NAME;
         Signature signature = new Signature(mTestSignature);
-        prepareProxyForTest(brokerProxy, authenticatorType, brokerPackage, "test", signature, true, null);
+        prepareProxyForTest(brokerProxy, authenticatorType, brokerPackage, "test", signature, true, null, true);
 
         // action
         Method m = ReflectionUtils.getTestMethod(brokerProxy, "canSwitchToBroker");
-        boolean result = (Boolean) m.invoke(brokerProxy);
+        BrokerProxy.SwitchToBroker result = (BrokerProxy.SwitchToBroker) m.invoke(brokerProxy);
 
         // assert
-        assertFalse("verify should return false", result);
+        assertEquals(BrokerProxy.SwitchToBroker.CANNOT_SWITCH_TO_BROKER, result);
     }
 
     public void testCanSwitchToBrokerInvalidSignature()
@@ -148,14 +153,14 @@ public class BrokerProxyTests extends AndroidTestCase {
         String authenticatorType = AuthenticationConstants.Broker.BROKER_ACCOUNT_TYPE;
         String brokerPackage = AuthenticationConstants.Broker.COMPANY_PORTAL_APP_PACKAGE_NAME;
         Signature signature = new Signature("74657374696e67");
-        prepareProxyForTest(brokerProxy, authenticatorType, brokerPackage, "test", signature, true, null);
+        prepareProxyForTest(brokerProxy, authenticatorType, brokerPackage, "test", signature, true, null, true);
 
         // action
         Method m = ReflectionUtils.getTestMethod(brokerProxy, "canSwitchToBroker");
-        boolean result = (Boolean) m.invoke(brokerProxy);
+        BrokerProxy.SwitchToBroker result = (BrokerProxy.SwitchToBroker) m.invoke(brokerProxy);
 
         // assert
-        assertFalse("verify should return false", result);
+        assertEquals(BrokerProxy.SwitchToBroker.CANNOT_SWITCH_TO_BROKER, result);
     }
 
     public void testCanSwitchToBrokerValid()
@@ -170,14 +175,14 @@ public class BrokerProxyTests extends AndroidTestCase {
         AuthenticationSettings.INSTANCE.setBrokerSignature(mTestTag);
         AuthenticationSettings.INSTANCE.setUseBroker(true);
         Account[] accts = getAccountList("valid", authenticatorType);
-        prepareProxyForTest(brokerProxy, authenticatorType, brokerPackage, contextPackage, signature, true, accts);
+        prepareProxyForTest(brokerProxy, authenticatorType, brokerPackage, contextPackage, signature, true, accts, true);
 
         // action
         Method m = ReflectionUtils.getTestMethod(brokerProxy, "canSwitchToBroker");
-        boolean result = (Boolean) m.invoke(brokerProxy);
+        BrokerProxy.SwitchToBroker result = (BrokerProxy.SwitchToBroker) m.invoke(brokerProxy);
 
         // assert
-        assertTrue("verify should return true", result);
+        assertEquals(BrokerProxy.SwitchToBroker.CAN_SWITCH_TO_BROKER, result);
     }
 
     public void testCanSwitchToBrokerValidSkip()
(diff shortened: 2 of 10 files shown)


adal/src/androidTest/java/com/microsoft/aad/adal/FileMockContext.java
```diff
@@ -67,8 +67,8 @@
     public FileMockContext(Context context) {
         mContext = context;
         // default
-        mRequestedPermissionName = "android.permission.INTERNET";
-        mResponsePermissionFlag = PackageManager.PERMISSION_GRANTED;
+        permissionMap.put("android.permission.INTERNET", PackageManager.PERMISSION_GRANTED);
+        //permissionMap.put("android.permission.GET_ACCOUNTS", PackageManager.PERMISSION_GRANTED);
```
addev-ashish: remove the commented code

adal/src/main/java/com/microsoft/aad/adal/AuthenticationContext.java
```diff
@@ -29,9 +29,11 @@
 import android.content.Context;
 import android.content.Intent;
 import android.content.pm.PackageManager;
+import android.os.Build;
 import android.os.Bundle;
 import android.os.Looper;
 import android.os.NetworkOnMainThreadException;
+import android.support.v4.content.ContextCompat;
```
addev-ashish: only used in commented code, remove this and all commented code

adal/src/main/java/com/microsoft/aad/adal/AuthenticationContext.java
```diff
@@ -849,9 +851,84 @@ private String checkInputParameters(final String resource, final String clientId
...
+        final String methodName = ":checkBrokerPermissions";
+        final PackageManager pm = mContext.getPackageManager();
+        final StringBuilder permissionMissing = new StringBuilder();
+
+        if (Build.VERSION.SDK_INT < Build.VERSION_CODES.M) {
+            Logger.v(
+                    TAG + methodName,
+                    "Broker related permissions checking starts.");
+
+            if (PackageManager.PERMISSION_GRANTED != pm.checkPermission("android.permission.GET_ACCOUNTS", mContext.getPackageName())) {
```
addev-ashish: >PackageManager.PERMISSION_GRANTED != pm.checkPermission("android.permission.GET_ACCOUNTS", mContext.getPackageName()) [](start = 16, length = 117)

normally we check function() != constant, lets follow that pattern

adal/src/main/java/com/microsoft/aad/adal/AuthenticationContext.java
```diff
@@ -849,9 +851,84 @@ private String checkInputParameters(final String resource, final String clientId
...
+            }
+        } else {
+            if (PackageManager.PERMISSION_GRANTED != pm.checkPermission("android.permission.GET_ACCOUNTS", mContext.getPackageName())) {
+                    //ContextCompat.checkSelfPermission(mContext, "android.permission.GET_ACCOUNTS")) {
+                permissionMissing.append("GET_ACCOUNTS");
+                Logger.w(
+                        TAG + methodName,
+                        "Broker related permissions are missing for GET_ACCOUNTS",
+                        "", ADALError.DEVELOPER_BROKER_PERMISSIONS_MISSING);
+            }
```
addev-ashish: this block repeats in both if and else, move it before the permissions check

adal/src/main/java/com/microsoft/aad/adal/AuthenticationContext.java
```diff
@@ -849,9 +851,84 @@ private String checkInputParameters(final String resource, final String clientId
...
+        }
+
+        if (permissionMissing.length() == 0) {
+            return true;
+        } else {
```
addev-ashish: >else [](start = 10, length = 4)

don't need else

adal/src/main/java/com/microsoft/aad/adal/BrokerProxy.java
```diff
@@ -159,15 +159,23 @@ public boolean canUseLocalCache() {
...
+        if (Build.VERSION.SDK_INT < Build.VERSION_CODES.M) {
+            permission = PackageManager.PERMISSION_GRANTED == pm.checkPermission(
                 "android.permission.GET_ACCOUNTS", mContext.getPackageName())
                 && PackageManager.PERMISSION_GRANTED == pm.checkPermission(
                         "android.permission.MANAGE_ACCOUNTS", mContext.getPackageName())
                 && PackageManager.PERMISSION_GRANTED == pm.checkPermission(
                         "android.permission.USE_CREDENTIALS", mContext.getPackageName());
+        } else {
+            permission = PackageManager.PERMISSION_GRANTED == pm.checkPermission(
+                "android.permission.GET_ACCOUNTS", mContext.getPackageName());
```
addev-ashish: you can move this to line 164 and if M, and with the other 2 checks

config/checkstyle/suppressions.xml
```diff
@@ -10,8 +10,8 @@
     <suppress checks="Javadoc.*" files=".*Test.java"/>
     <suppress checks="MethodName" files=".*Logger.java"/>
     <suppress checks="MethodName" files=".*LoggerTest.java"/>
-    <suppress checks="MemberName" files="AuthenticationContext.java" lines="813"/>
-    <suppress checks="VisibilityModifier" files="AuthenticationContext.java" lines="813"/>
+    <suppress checks="MemberName" files="AuthenticationContext.java" lines="815"/>
```
addev-ashish: >815 [](start = 76, length = 3)

this is not scalable, why not use the 
//CHECKSTYLE:OFF: checkstyle:membername

//CHECKSTYLE:ON: checkstyle:membername

adal/src/main/java/com/microsoft/aad/adal/AuthenticationContext.java
```diff
@@ -849,9 +851,84 @@ private String checkInputParameters(final String resource, final String clientId
...
+     * To verify if App gives permissions to AccountManager to use broker. 
+     * If target version is lower than 23, calling app has to have GET_ACCOUNTS, 
+     * MANAGE_ACCOUNTS, USE_CREDENTIALS permissions granted in the Manifest.xml.
+     * Beginning in Android 6.0, the run-time permission GET_ACCOUNTS is required
+     * which need to be requested in the runtime by the calling app.
+     * 
+     * @return true if all required permissions are granted, otherwise the exception will be thrown.
+     * @throws UsageAuthenticationException 
+     */
+    private boolean checkBrokerPermissions() throws UsageAuthenticationException {
```
weijjia: why we have the checkBrokerPermissions here and also do the same thing in verifyManifestPermissions? What is the difference between these two?

adal/src/androidTest/java/com/microsoft/aad/adal/AcquireTokenRequestTest.java
```diff
@@ -781,6 +782,98 @@ public void testResiliencyTokenReturnExtendedLifetimeOnwithoutRetry() throws Pac
...
+
+        assertNotNull(callback.getCallbackException());
+        assertTrue(callback.getCallbackException() instanceof UsageAuthenticationException);
+        final UsageAuthenticationException usageAuthenticationException
+                = (UsageAuthenticationException) callback.getCallbackException();
+        assertTrue(usageAuthenticationException.getMessage().contains("MANAGE_ACCOUNTS"));
+        assertEquals(ADALError.DEVELOPER_BROKER_PERMISSIONS_MISSING, usageAuthenticationException.getCode());
+    }
+
+    public void testVerifyManifestPermission_missingPermission_USE_CREDENTIALS() throws InterruptedException, PackageManager.NameNotFoundException {
```
weijjia: this is not the naming convention. We did the work to make sure that our code is clean from the two static analysis tools, even though tests files are excluded from the check, I think we should make sure we do our best to not introduce new code in test that is not compliant with the rule sets.

adal/src/main/java/com/microsoft/aad/adal/AuthenticationContext.java
heidijinxujia: In BrokerProxy, we only need a simple check (a boolean value) to indicator if all permission are granted. And here, we also need detailed info of the missing permission. I wrote this additional private function to 1) throw exception with detailed info of the missing permission 2) it makes the code clear instead of calling the method in BrokerProxy. How do you guys think? @addev-ashish @weijjia

adal/src/main/java/com/microsoft/aad/adal/AuthenticationContext.java
```diff
@@ -810,6 +811,8 @@ private IWindowComponent wrapActivity(final Activity activity) {
         }
 
         return new IWindowComponent() {
+            //CHECKSTYLE:OFF: checkstyle:MemberName
+            //CHECKSTYLE:OFF: checkstyle:VisibilityModifier
             Activity mRefActivity = activity;
```
addev-ashish: sorry about this, but how about just making this private? does it break anything?

adal/src/main/java/com/microsoft/aad/adal/AuthenticationContext.java
addev-ashish: i agree with Wei, these 2 functions should be merged.

---
In reply to: [75406588](https://github.com/AzureAD/azure-activedirectory-library-for-android/pull/698#discussion_r75406588) [](ancestors = 75406588)

adal/src/main/java/com/microsoft/aad/adal/AuthenticationContext.java
heidijinxujia: Sorry, do you mean the CheckStyle or the method?

addev-ashish: make mRefActivity private, checkstyle will stop complaining after that (so you do not need the checkstyle:off comments). I was asking if doing this would break anything?

adal/src/main/java/com/microsoft/aad/adal/AuthenticationContext.java
weijjia: I don't think that we should have another duplicate method to make a simple check. )

adal/src/main/java/com/microsoft/aad/adal/AuthenticationContext.java
```diff
@@ -849,9 +854,68 @@ private String checkInputParameters(final String resource, final String clientId
...
+     * @return true if all required permissions are granted, otherwise the exception will be thrown.
+     * @throws UsageAuthenticationException 
+     */
+    private boolean checkBrokerPermissions() throws UsageAuthenticationException {
+        final String methodName = ":checkBrokerPermissions";
+        final PackageManager pm = mContext.getPackageManager();
+        final StringBuilder permissionMissing = new StringBuilder(46);
+
+        if (pm.checkPermission("android.permission.GET_ACCOUNTS", mContext.getPackageName()) != PackageManager.PERMISSION_GRANTED) {
+            permissionMissing.append("GET_ACCOUNTS ");
```
weijjia: you can have a private method taking int he string builder, and permission name, wrap the append and logging there.

adal/src/main/java/com/microsoft/aad/adal/BrokerProxy.java
weijjia: This check should be combined as Ashish said. 
You can do in this way:
if (Build.Version.SDK_INT < M) {
    // check manage_account and use_credential
}

// check get_account

adal/src/androidTest/java/com/microsoft/aad/adal/AcquireTokenRequestTest.java
```diff
@@ -780,18 +781,86 @@ public void testResiliencyTokenReturnExtendedLifetimeOnwithoutRetry() throws Pac
...
+                VALID_AUTHORITY, false);
+        final TestAuthCallback callback = new TestAuthCallback();
+        authContext.acquireToken(Mockito.mock(Activity.class), "resource", "clientid", authContext.getRedirectUriForBroker(),
+                "loginHint", callback);
+
+        assertNull(callback.getCallbackException());
+    }
+
+    @SmallTest
+    public void testVerifyManifestPermissionMissingContactsPermission() throws InterruptedException, PackageManager.NameNotFoundException {
```
weijjia: Change the test name please, contact permission includes more.

adal/src/main/java/com/microsoft/aad/adal/AuthenticationContext.java
heidijinxujia: OK. I will merge these two function.

adal/src/main/java/com/microsoft/aad/adal/AuthenticationContext.java
heidijinxujia: All the test passed. I will try with the device manually if necessary.

adal/src/androidTest/java/com/microsoft/aad/adal/AcquireTokenRequestTest.java
heidijinxujia: OK.

adal/src/main/java/com/microsoft/aad/adal/AuthenticationContext.java
weijjia: @heidijinxujia it has nothing to do with test passed or not. What @addev-ashish means is that we can make the mRefActivity to be private then we can get rid of the comment to suppress checkstyle.

heidijinxujia: Yes, I understood. I addressed @addev-ashish 's commit and run the test. My previous reply is trying to answer @addev-ashish 's question on "if doing this would break anything". :)

adal/src/main/java/com/microsoft/aad/adal/AuthenticationContext.java
```diff
@@ -849,6 +850,15 @@ private String checkInputParameters(final String resource, final String clientId
             redirectUri = inputRedirectUri;
         }
 
+        //check the permissions required for the broker usage
+        if (AuthenticationSettings.INSTANCE.getUseBroker()) {
+            try {
+                mBrokerProxy.verifyManifestPermissions();
+            } catch (final UsageAuthenticationException exception) {
+                callback.onError(exception);
```
addev-ashish: >callback.onError(exception); [](start = 16, length = 28)

so you are going to call the error callback and then ruturn the redirect uri?

adal/src/main/java/com/microsoft/aad/adal/BrokerProxy.java
```diff
@@ -156,22 +162,45 @@ public boolean canUseLocalCache() {
...
-        boolean permission = PackageManager.PERMISSION_GRANTED == pm.checkPermission(
-                "android.permission.GET_ACCOUNTS", mContext.getPackageName())
-                && PackageManager.PERMISSION_GRANTED == pm.checkPermission(
-                        "android.permission.MANAGE_ACCOUNTS", mContext.getPackageName())
-                && PackageManager.PERMISSION_GRANTED == pm.checkPermission(
-                        "android.permission.USE_CREDENTIALS", mContext.getPackageName());
-        if (!permission) {
-            Logger.w(TAG, "Broker related permissions are missing for GET_ACCOUNTS, MANAGE_ACCOUNTS, USE_CREDENTIALS",
+    public boolean verifyManifestPermissions() throws UsageAuthenticationException {
+        StringBuilder permissionMissing = new StringBuilder(46);
```
addev-ashish: this should throw a magic number exception in one of the tools, did it not?

adal/src/main/java/com/microsoft/aad/adal/BrokerProxy.java
```diff
@@ -156,22 +162,45 @@ public boolean canUseLocalCache() {
...
+        if (permissionMissing.length() == 0) {
+            return true;
+        }
+
+        throw new UsageAuthenticationException(
+                ADALError.DEVELOPER_BROKER_PERMISSIONS_MISSING,
+                "Broker related permissions are missing for " + permissionMissing.toString());
+    }
+
+    private String checkPermission(String permissionName) {
```
addev-ashish: >String [](start = 35, length = 6)

final

adal/src/main/java/com/microsoft/aad/adal/BrokerProxy.java
addev-ashish: >StringBuilder [](start = 8, length = 13)

final

adal/src/main/java/com/microsoft/aad/adal/AuthenticationContext.java
heidijinxujia: Addressed

addev-ashish: final String redirectUri should also go below this block

---
In reply to: [75787853](https://github.com/AzureAD/azure-activedirectory-library-for-android/pull/698#discussion_r75787853) [](ancestors = 75787853)

adal/src/androidTest/java/com/microsoft/aad/adal/AcquireTokenRequestTest.java
```diff
@@ -780,18 +781,86 @@ public void testResiliencyTokenReturnExtendedLifetimeOnwithoutRetry() throws Pac
...
+    public void testVerifyManifestPermissionValid() throws InterruptedException, PackageManager.NameNotFoundException {
+        final FileMockContext mockContext = createMockContext();
+        AuthenticationSettings.INSTANCE.setUseBroker(true);
+        final AuthenticationContext authContext = new AuthenticationContext(mockContext,
+                VALID_AUTHORITY, false);
+        final TestAuthCallback callback = new TestAuthCallback();
+        authContext.acquireToken(Mockito.mock(Activity.class), "resource", "clientid", authContext.getRedirectUriForBroker(),
+                "loginHint", callback);
+
+        assertNull(callback.getCallbackException());
```
weijjia: Just out of curiosity, acquiretoken is async call, you know the test is not reliable to pass if we do it this way, right?

adal/src/main/java/com/microsoft/aad/adal/AuthenticationContext.java
weijjia: Why not throwing directly?

adal/src/androidTest/java/com/microsoft/aad/adal/AcquireTokenRequestTest.java
heidijinxujia: I will change it to sync call.

adal/src/main/java/com/microsoft/aad/adal/AuthenticationContext.java
heidijinxujia: UsageAuthenticationException is not a runtime exception. If we throw here directly, we have to add the throw statement in the checkInputParameters() function, which will affect all the functions calling it. That is we have to deal with the UsageAuthenticationException thrown by checkInputParameters in all the acquireToken function.

adal/src/main/java/com/microsoft/aad/adal/AuthenticationContext.java
```diff
@@ -843,6 +843,17 @@ private String checkInputParameters(final String resource, final String clientId
         }
 
         final String redirectUri;
+
+        //check the permissions required for the broker usage
+        if (AuthenticationSettings.INSTANCE.getUseBroker()) {
```
heidijinxujia: Move to method named "CheckForPermissionAPI22orLess". This should do the following:
1. if API < 23 && useBroker == true && permissionNotGranted, then return error. Here by return error, throw UsageAuthenticationException and pass it to callback
2. Wrap calls to checkInputParameters and checkForPermissionAPI22orLess into a wrapper function checkPreReqs and replace checkInputParameters

adal/src/main/java/com/microsoft/aad/adal/BrokerProxy.java
```diff
@@ -116,19 +116,25 @@ public BrokerProxy(final Context ctx) {
...
+        try {
+            // ADAL switches broker for following conditions:
+            // 1- app is not skipping the broker
+            // 2- permissions are set in the manifest,
+            // 3- if package is not broker itself for both company portal and azure
+            // authenticator
+            // 4- signature of the broker is valid
+            // 5- account exists
+            return AuthenticationSettings.INSTANCE.getUseBroker()
+                    && verifyManifestPermissions()
```
heidijinxujia: 1. rename to  "CheckForPermissionAPI23orAbove", if API > 22 && other four checks get passed && permissionNotGranted, then return error. Here by return error, throw UsageAuthenticationException and pass it to callback
2. this verification should be the last one to check

adal/src/main/java/com/microsoft/aad/adal/BrokerProxy.java
```diff
@@ -156,22 +169,59 @@ public boolean canUseLocalCache() {
...
+        if (permissionMissing.length() == 0) {
+            return true;
+        }
+
+        throw new UsageAuthenticationException(
+                ADALError.DEVELOPER_BROKER_PERMISSIONS_MISSING,
+                "Broker related permissions are missing for " + permissionMissing.toString());
+    }
+
+    public boolean verifyBrokerPermissionsAPI23AndHigher() throws UsageAuthenticationException {
```
heidijinxujia: annotation named @targetAPI

adal/src/androidTest/java/com/microsoft/aad/adal/AcquireTokenRequestTest.java
```diff
@@ -880,20 +955,21 @@ private PackageManager getMockedPackageManager() throws PackageManager.NameNotFo
...
 
+        // Mock permissions
+        when(mockedPackageManager.checkPermission(Mockito.refEq("android.permission.GET_ACCOUNTS"),
+                Mockito.anyString())).thenReturn(PackageManager.PERMISSION_GRANTED);
+        when(mockedPackageManager.checkPermission(Mockito.refEq("android.permission.MANAGE_ACCOUNTS"),
+                Mockito.anyString())).thenReturn(PackageManager.PERMISSION_GRANTED);
+        when(mockedPackageManager.checkPermission(Mockito.refEq("android.permission.USE_CREDENTIALS"),
+                Mockito.anyString())).thenReturn(PackageManager.PERMISSION_GRANTED);
+
         return mockedPackageManager;
```
heidijinxujia: revert

adal/src/main/java/com/microsoft/aad/adal/IBrokerProxy.java
```diff
@@ -37,12 +37,16 @@
...
      *         Account manager
      */
-    boolean canSwitchToBroker();
+    BrokerProxy.SwitchToBroker canSwitchToBroker();
     
     boolean verifyUser(String username, String uniqueid);
 
     boolean canUseLocalCache();
 
+    boolean verifyBrokerPermissionsAPI22AndLess() throws UsageAuthenticationException;
```
weijjia: Does these two need to be added in the interface?

heidijinxujia: They cannot be removed if we keep the declaration of mBrokerProxy in AuthenticationContext. Because verifyBrokerPermissionsAPI22AndLess() is called in AuthenticationContext, and the property mBrokerProxy is defined as below:
*private IBrokerProxy mBrokerProxy = null;*
We could not call these two functions if they are not defined in the interface IBrokerProxy.

One possible solution is to change to:
*private BrokerProxy mBrokerProxy = null;*

weijjia: ok, then keep what you have.

adal/src/androidTest/java/com/microsoft/aad/adal/AcquireTokenRequestTest.java
```diff
@@ -780,18 +781,79 @@ public void testResiliencyTokenReturnExtendedLifetimeOnwithoutRetry() throws Pac
...
         }
     }
-    
+
+    @SmallTest
+    public void testVerifyManifestPermissionMissingGetAccountsPermission() throws InterruptedException, PackageManager.NameNotFoundException {
+        final FileMockContext mockContext = createMockContext();
+        when(mockContext.getPackageManager().checkPermission(Mockito.refEq("android.permission.GET_ACCOUNTS"),
+                Mockito.anyString())).thenReturn(PackageManager.PERMISSION_DENIED);
+        AuthenticationSettings.INSTANCE.setUseBroker(true);
```
weijjia: In the tearDown, you should set it to the default value to clear the test state.

adal/src/androidTest/java/com/microsoft/aad/adal/BrokerProxyTests.java
```diff
@@ -825,6 +844,10 @@ private Context getMockContext(final Signature signature, final String brokerPac
         PackageManager mockPackageManager = getPackageManager(signature, brokerPackageName, permissionStatus);
         when(mockContext.getPackageManager()).thenReturn(mockPackageManager);
         when(mockContext.getPackageName()).thenReturn(contextPackageName);
+        List<ResolveInfo> mockList = new ArrayList<>();
+        mockList.add(new ResolveInfo());
+        int size = mockList.size();
```
weijjia: used anywhere?

adal/src/androidTest/java/com/microsoft/aad/adal/AcquireTokenRequestTest.java
heidijinxujia: This is set into the default value in my previous PR.

adal/src/androidTest/java/com/microsoft/aad/adal/BrokerProxyTests.java
heidijinxujia: Here in Line 848 below, mockList is used in mocking the returns of PackageManager.queryIntentActivities().

weijjia: sorry, I mean size.

adal/src/main/java/com/microsoft/aad/adal/BrokerProxy.java
```diff
@@ -107,28 +107,42 @@ public BrokerProxy(final Context ctx) {
...
+        boolean switchToBrokerFlag =
+                AuthenticationSettings.INSTANCE.getUseBroker()
                 && checkAccount(mAcctManager, "", "")
                 && !packageName.equalsIgnoreCase(AuthenticationSettings.INSTANCE.getBrokerPackageName())
                 && !packageName.equalsIgnoreCase(AuthenticationConstants.Broker.AZURE_AUTHENTICATOR_APP_PACKAGE_NAME)
                 && verifyAuthenticator(mAcctManager);
+
+        if (!switchToBrokerFlag) {
+            return SwitchToBroker.CANNOT_SWITCH_TO_BROKER;
+        } else {
```
addev-ashish: >else [](start = 10, length = 4)

dont need this else block

adal/src/androidTest/java/com/microsoft/aad/adal/AcquireTokenRequestTest.java
weijjia: oh ok, thanks.

addev-ashish:
:shipit:

PR #699, opened 2016-08-11
manual merge PR 696 from dev to service branch

weijjia:
manual merge PR 696 from dev to servicing branch. 
For issue 695: outlook is seeing empty refresh token stored in token cache item, it's only in outlook dogfood build, no user report yet, there isn't too much diagnostic info to tell the root cause. The fix is to prevent outlook from crashing in this case.

### DIFF
--- a/tests/Functional/src/com/microsoft/aad/adal/OauthTests.java
+++ b/tests/Functional/src/com/microsoft/aad/adal/OauthTests.java
@@ -636,11 +636,19 @@ public class OauthTests extends AndroidTestCase {
         assertEquals("Token is same", "token", result.getAccessToken());
         assertFalse("MultiResource token", result.getIsMultiResourceRefreshToken());
 
-        // multi resource token
+        // resource returned in JSON response, but RT is not returned.
         response.put(AuthenticationConstants.AAD.RESOURCE, "resource");
         result = (AuthenticationResult)m.invoke(null, response);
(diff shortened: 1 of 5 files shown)


weijjia:
@addev-ashish As we talked, here is the the PR manually merged.

tests/Functional/src/com/microsoft/aad/adal/OauthTests.java
```diff
@@ -636,11 +636,19 @@ public void testprocessUIResponseParams() throws IllegalArgumentException,
...
 
-        // multi resource token
+        // resource returned in JSON response, but RT is not returned.
         response.put(AuthenticationConstants.AAD.RESOURCE, "resource");
         result = (AuthenticationResult)m.invoke(null, response);
         assertEquals("Success status", AuthenticationStatus.Succeeded, result.getStatus());
         assertEquals("Token is same", "token", result.getAccessToken());
+        assertFalse("MultiResource token", result.getIsMultiResourceRefreshToken());
+		
+		// resource returned in JSON response and RT is also returned.
```
addev-ashish: >		 [](start = 0, length = 2)

tabs

weijjia: beyond compare is not good enough. :) updated.

addev-ashish:
:shipit:

PR #722, opened 2016-09-27
Fix potential Activity leak with IWindowComponents

christiandeange:
`AuthenticationContext` creates an anonymous inner `IWindowComponent` class that holds a reference to the Activity parameter. If the point of that anonymous inner class's `mRefActivity` is to ensure that the Activity isn't going to be leaked, then it needs to hold a WeakReference to it instead of just a regular one.

Creating a class to manage that is the easiest solution. For parity I've also created one to wrap a support Fragment, and made those classes public so they can be used by developers using this library.

### DIFF
--- a/adal/src/main/java/com/microsoft/aad/adal/AuthenticationContext.java
+++ b/adal/src/main/java/com/microsoft/aad/adal/AuthenticationContext.java
@@ -816,18 +816,7 @@ public class AuthenticationContext {
             throw new IllegalArgumentException("activity");
         }
 
-        return new IWindowComponent() {
-            private Activity mRefActivity = activity;
-
-            @Override
-            public void startActivityForResult(Intent intent, int requestCode) {
-                // if user closed an app or switched to another activity
-                // mRefActivity can die before this method got invoked
-                if (mRefActivity != null) {
-                    mRefActivity.startActivityForResult(intent, requestCode);
-                }
-            }
-        };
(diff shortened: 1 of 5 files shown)


adal/src/main/java/com/microsoft/aad/adal/AuthenticationContext.java
```diff
@@ -816,18 +816,7 @@ private IWindowComponent wrapActivity(final Activity activity) {
             throw new IllegalArgumentException("activity");
         }
 
-        return new IWindowComponent() {
-            private Activity mRefActivity = activity;
```
addev-ashish: Thank you for your contribution. I think we would achieve the same result by changing mRefActivity to be a WeakReference. do you agree?

cdeange: Not exactly. The `new IWindowComponent() { ... }` class holds a reference to the outer `AuthenticationContext` class as long as it is still in scope. Because of this, the `activity` parameter passed into `wrapActivity` will also still be in scope.

weijjia:
@cdeange Can you sign the CLA and resolve the merge conflicts? We would love to merge the PR once you do that :)

cdeange:
@weijjia I have signed the CLA, not sure why the PR status has not been updated yet to reflect that

christiandeange:
Closing due to inactivity.

PR #734, opened 2016-11-16
Remove the GET_ACCOUNTS permission

weijjia:
GET_ACCOUNTS permission is required today for ADAL to directly interact with account manager. For getBrokerUsers and getAuthToken call, ADAL needs to get all the users in the broker which results in a the call getAccountsByAccountType. 

To remove the requirement on the permission, adal needs to move away from directly calling account manager, all the call will be directed to the bound service running inside the active broker. adal will only talk to the service and the service will be responsible for interacting with account manager.


adal/src/main/java/com/microsoft/aad/adal/BrokerAccountServiceHandler.java
```diff
@@ -0,0 +1,363 @@
...
+    /**
+     * Private constructor to prevent class from being instantiate.
+     */
+    private BrokerAccountServiceHandler() { }
+
+    /**
+     * Get Broker users is a blocking call, cannot be executed on the main thread.
+     * @return An array of {@link UserInfo}s in the broker. If no user exists in the broker, empty array will be returned.
+     */
+    UserInfo[] getBrokerUsers(final Context context) throws IOException {
```
weijjia: Should add a util method to perform async call synchronously.

adal/src/androidTest/java/com/microsoft/aad/adal/BrokerAccountServiceTest.java
```diff
@@ -0,0 +1,309 @@
...
+import java.util.concurrent.Executors;
+
+/**
+ * Test cases for brokerAccountService and related operations in {@link BrokerProxy}.
+ */
+public final class BrokerAccountServiceTest extends ServiceTestCase<MockBrokerAccountService> {
+    private static ExecutorService sThreadExecutor = Executors.newSingleThreadExecutor();
+    private static final String VALID_AUTHORITY = "https://login.microsoftonline.com";
+
+    private IBinder mIBiner;
```
weijjia: type

userappwithbroker/src/main/res/layout/activity_main.xml
```diff
@@ -47,12 +47,20 @@
...
+        android:layout_below="@+id/AcquireTokenSilentForR2"
+        android:layout_centerHorizontal="true"
+        android:text="@string/GetBrokerUsers" />
+
     <EditText
         android:id="@+id/editTextUsername"
         android:layout_width="wrap_content"
         android:layout_height="wrap_content"
         android:layout_centerHorizontal="true"
-        android:layout_below="@+id/AcquireTokenSilentForR2"
```
addev-ashish: >AcquireTokenSilentForR2 [](start = 35, length = 23)

Not a UI expert, but does this look correct?

adal/src/main/java/com/microsoft/aad/adal/BrokerProxy.java
```diff
@@ -335,21 +344,60 @@ public AuthenticationResult getAuthTokenInBackground(final AuthenticationRequest
...
+    }
+
+    private boolean isServiceSupported(final Context context, final Intent intent) {
+        if (intent == null) {
+            return false;
+        }
+
+        final PackageManager packageManager = context.getPackageManager();
+        final List<ResolveInfo> infos = packageManager.queryIntentServices(intent, 0);
+        return infos != null && infos.size() > 0;
```
addev-ashish: >infos != null && infos.size() > 0 [](start = 15, length = 33)

you should also check for infos.serviceInfo to not be null.

adal/src/main/java/com/microsoft/aad/adal/BrokerProxy.java
```diff
@@ -583,9 +655,22 @@ private boolean isBrokerWithPRTSupport(final Intent intent) {
...
-        if (accountList.length > 0) {
-            return accountList[0].name;
+        if (isBrokerAccountServiceSupported()) {
+            verifyNotOnMainThread();
+
+            final UserInfo[] users;
+            try {
+                users = BrokerAccountServiceHandler.getInstance().getBrokerUsers(mContext);
+            } catch (final IOException e) {
+                return null;
```
addev-ashish: >null [](start = 23, length = 4)

Log?

adal/src/main/java/com/microsoft/aad/adal/BrokerProxy.java
```diff
@@ -807,17 +892,24 @@ private boolean verifyAuthenticator(final AccountManager am) {
...
-        final String methodName = ":getBrokerUsers";
         // Calling this on main thread will cause exception since this is
         // waiting on AccountManagerFuture
         if (Looper.myLooper() == Looper.getMainLooper()) {
             throw new IllegalArgumentException("Calling getBrokerUsers on main thread");
         }
 
+        if (isBrokerAccountServiceSupported()) {
+            return BrokerAccountServiceHandler.getInstance().getBrokerUsers(mContext);
+        } else {
```
addev-ashish: >else [](start = 10, length = 4)

else is not required

adal/src/main/java/com/microsoft/aad/adal/BrokerProxy.java
```diff
@@ -583,9 +655,22 @@ private boolean isBrokerWithPRTSupport(final Intent intent) {
      */
     public String getCurrentUser() {
         // authenticator is not used if there is not any user
-        Account[] accountList = mAcctManager.getAccountsByType(AuthenticationConstants.Broker.BROKER_ACCOUNT_TYPE);
-        if (accountList.length > 0) {
-            return accountList[0].name;
+        if (isBrokerAccountServiceSupported()) {
```
imakhalova: >isBrokerAccountServiceSupported [](start = 12, length = 31)

Can you make some public method isPermissionsRequieredForBrokerApp that will do the same as isBrokerAccountServiceSupported
So we can use it within app to detect whether we should prompt for permissions

userappwithbroker/src/main/res/layout/activity_main.xml
weijjia: This shows that the editText will be under the getBrokerUsers button.

adal/src/main/java/com/microsoft/aad/adal/BrokerProxy.java
weijjia: why? we're passing the intent, if the service is not supported, queryIntentServices won't return you back anything. 

---
In reply to: [95735693](https://github.com/AzureAD/azure-activedirectory-library-for-android/pull/734#discussion_r95735693) [](ancestors = 95735693)

adal/src/main/java/com/microsoft/aad/adal/BrokerProxy.java
weijjia: You are getting back a special error code(DEVELOPER_BROKER_PERMISSIONS_MISSING) when permisison is missing, you can request permisison based on the error code. No need to expose a new public API. 

---
In reply to: [101668304](https://github.com/AzureAD/azure-activedirectory-library-for-android/pull/734#discussion_r101668304) [](ancestors = 101668304)

imakhalova: Client doesn't setUseBroker to true if permissions are not set...And this actually was your design.

weijjia: I'm not sure i'm following. you should always opt-in to use broker and prompt user to grant the permission if it's missing (we'll send you an error back to let you know it's missing). 

---
In reply to: [102327617](https://github.com/AzureAD/azure-activedirectory-library-for-android/pull/734#discussion_r102327617) [](ancestors = 102327617)

weijjia: Ooops, codeflow doesn't get my reply posted here. Sorry about that. 
I'm not sure i'm following. you should always opt-in to use broker and prompt user to grant the permission if it's missing (we'll send you an error back to let you know it's missing).

Issue #741, opened 2016-12-03
Pkey auth failure on ADFS 2016


weijjia:
On adfs 2016, the submit url returns the full url with query string:
https://adfs.diyanetvakfi.org.tr:443/adfs/ls/?login_hint=some_username&wfresh=0&wauth=http%3a%2f%2fschemas.microsoft.com%2fws%2f2008%2f06%2fidentity%2fauthenticationmethod%2fpassword&client-request-id=928cc655-cb4a-4233-b83d-2b6f4e240a7c&username=some_username&wa=wsignin1.0&wtrealm=urn%3afederation%3aMicrosoftOnline&wctx=estsredirect%3d2%26estsrequest%3drQIIAaWRv4vTUADH7yXtoUW44yYHEcEKoiTNe0mTpnJg60l7te1xrb0z3d57fWlyTfJqfp0tDv4BDtpFdBSnAxdBOG4VFHTpKI6O5yK4uGludnT58OW7fL98v4VVKENFVjYAvCFmslqENqMVZlQkQpkmaRjbEiaQSGWtDClVKwbStXCjsD74-HTR-qLWjzdv8m6x8PgI3PIjnMROtVSi3Jd9l4Y84nYsxwz7Ucmm44ryMOUdaHXutYYuOagf7tIteyfg19StYwCWADwRwEK46mDP9WSfex7mYy-5PXJnOGBxiie2K_NwLMfhUgDfhLWdWhaHzsBDd86-_-OcCnmPj91gIa7xs2pIPiPlI_ZKJNZ9yEfN3iGd87SNHqVD5CVtH04ttZVSZPqjhpmQhpcM--UDgpTUQmZM9s2kFWybuNlTaLOjt2emQxuThKp1x0J783bQDdmDutf2uynZPRIvOXE8jbJF8NSVoymmLJKjyWzK5Gyid-JqRp8HJ2IRG6bOMCGSARVd0gxoSMRQscR0jJANMdJs-6sIljlwmrtwDqwXLq5cWbkuKOB3DrzOZ3d8-Pnnxa-3P1pv3r8Ezy8_-_Q5X9reH_S7Tm_ctZozHLH0LuyZJdUcdPU7k6C2V2d9rNfmfS1Ia5tGFZ6c_58D_wI1. 

When ADAL gets the challenge and responds to it, it tries add all the original query string sent in the original request, it's causing the extra "?" to be added.

PR #742, opened 2016-12-03
issue 741: remove the appended query string when responding to pkeyauth challenge

weijjia:
Issue #741: ADFS 2016 returns the full url for pkeyauth challenge. When ADAL receives the challenge and responds to it, it tries to add all the query string sent in the original authorize request back to the submit url, which adds an extra "?". Server explicitly sends back the submit uri, there is no need for us to append query parameter.


kpanwar:
🚢-it, but get buy in from @RandalliLama about sending x-client-ver to all the requests.

PR #743, opened 2016-12-05
Update the test app to be able to save ADAL logs

heidijinxujia:
Issue#738
- add the feature in the sample app to collect ADAL logs
- add the button in the UI to show the ADAL logs
- modify button for "invalidateRT"
- add button for "invalidateFRT"
- add button for "readLogs"
- fix the bug for invalidating the FRT

### DIFF
--- a/automationtestapp/src/main/java/com/microsoft/aad/automation/testapp/MainActivity.java
+++ b/automationtestapp/src/main/java/com/microsoft/aad/automation/testapp/MainActivity.java
@@ -74,15 +74,6 @@ public class MainActivity extends AppCompatActivity {
             }
         });
         
-        final Button getLogTokenButton = (Button) findViewById(R.id.getadallogs);
-        getLogTokenButton.setOnClickListener(new View.OnClickListener() {
-            @Override
-            public void onClick(View view) {
-                processShowLogs();
-            }
-        });
-        
-        
         // Button for acquireToken call
         final Button acquireTokenButton = (Button) findViewById(R.id.acquireToken);
         acquireTokenButton.setOnClickListener(new View.OnClickListener() {
@@ -163,6 +154,7 @@ public class MainActivity extends AppCompatActivity {
                     + e.getMessage());
         }
 
+        intent.putExtra(Constants.READ_LOGS, getADALLogs());
         launchResultActivity(intent);
     }
 
@@ -177,14 +169,13 @@ public class MainActivity extends AppCompatActivity {
                     + e.getMessage());
         }
 
+        intent.putExtra(Constants.READ_LOGS, getADALLogs());
         launchResultActivity(intent);
     }
     
     private void processShowLogs() {
         Intent intent = new Intent();
-        final ArrayList<String> allLogs = new ArrayList<>();
-        allLogs.add(getADALLogs());
-        intent.putStringArrayListExtra(Constants.READ_LOGS, allLogs);
+        intent.putExtra(Constants.READ_LOGS, getADALLogs());
         launchResultActivity(intent);
     }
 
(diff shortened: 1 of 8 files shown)


heidijinxujia:
Hi @weijjia @kpanwar I've addressed the comments. Please let me know if there is anything more you would like to add into the test app. Thanks a lot. :)

automationtestapp/src/main/java/com/microsoft/aad/automation/testapp/MainActivity.java
```diff
@@ -50,17 +53,28 @@
...
 
     @Override
     protected void onCreate(Bundle savedInstanceState) {
         super.onCreate(savedInstanceState);
         setContentView(R.layout.activity_main);
-
         mContext = getApplicationContext();
+        
+        mADALLogs = new StringBuffer();
+        Logger.getInstance().setExternalLogger(new Logger.ILogger() {
```
weijjia: I think you can set the external logger in the Application class, that way we won't need to have it in both MainActivity and SigninActivity.

automationtestapp/src/main/java/com/microsoft/aad/automation/testapp/ResultActivity.java
```diff
@@ -61,6 +66,10 @@ protected void onCreate(Bundle savedInstanceState) {
         }
         mTextView.setText(resultText);
 
+        mLogView = (TextView) findViewById(R.id.adalLogs);
+        mLogView.setMovementMethod(new ScrollingMovementMethod());
```
weijjia: do you need the log text box to be scrollable?

automationtestapp/src/main/java/com/microsoft/aad/automation/testapp/ResultActivity.java
```diff
@@ -45,6 +48,8 @@
 public class ResultActivity extends AppCompatActivity {
 
     private TextView mTextView;
+    private TextView mLogView;
+    private StringBuffer mADALLogs = new StringBuffer();
```
weijjia: why the mADALLogs cannot be declared locally?

automationtestapp/src/main/java/com/microsoft/aad/automation/testapp/ResultActivity.java
heidijinxujia: It depends on the requirement. We don't need to set the movement method for the text box if we don't need to show all the logs.

automationtestapp/src/main/java/com/microsoft/aad/automation/testapp/MainActivity.java
heidijinxujia: Got it, thanks @weijjia :)

automationtestapp/src/main/java/com/microsoft/aad/automation/testapp/ResultActivity.java
weijjia: My understanding is that the textbox is used to save the logs, not for us to actually show it. IMHO, there is no need to set the movement method. :)

automationtestapp/src/main/java/com/microsoft/aad/automation/testapp/ResultActivity.java
heidijinxujia: Addressed. Moved to the application class.

automationtestapp/src/main/java/com/microsoft/aad/automation/testapp/ResultActivity.java
heidijinxujia: Addressed. Thank you, @weijjia

kpanwar:
:shipit:

PR #744, opened 2016-12-06
Adding *.pom to .gitignore

iambmelt:
@weijjia I took the liberty of adding `*.pom` to `./gitignore` to support the general case instead of adding it to `./adal/.gitignore`. If you'd prefer to do this one way or another, feel free to let me know 😄


weijjia:
It looks good to me. :shipit:

PR #747, opened 2016-12-13
Restrict changes to additive, set tests to use old flow

iambmelt:
This PR restricts the changes of AD FS validation to be purely additive. It sets tests to use the original, not AD FS validation flow.


iambmelt:
Tests not expected to pass: PMD validation fails (empty conditional) + outstanding `// TODOs`

PR #756, opened 2017-01-09
Add back spinner log message to AuthenticationActivity.

wyber:
Add back a log message indicating the spinner state on
the AuthenticationActivity.


weijjia:
Change looks good. For side note, Company Portal today uses the log message to determine the time user stays on the sign in page for their SLA. We'll need this until we get telemetry in.

Issue #753, opened 2017-01-03
Crash:  invalid date string: Unparseable date: "ag`a`a`b`cbbGMT+00:00" (at offset 0)


khalidaw:
Reason: 
java.lang.IllegalArgumentException: invalid date string: Unparseable date: "ag`a`a`b`cbbGMT+00:00" (at offset 0)

Location: 
com.microsoft.aad.adal.StorageHelper.generateKeyPairFromAndroidKeyStore, line 414

Can be related to number in Arabic issue, but for sure this should be in catch an not cause app crashes.

e54999:
e549999@yahoo.com.tw

e54999:
0974294297@e549999.onmicrosoft.com

hjymandroid:
It crashes bad for us.

Exception Backtrace:
1	android.security.keystore.AndroidKeyStoreKeyPairGeneratorSpi.generateKeyPair	AndroidKeyStoreKeyPairGeneratorSpi.java, line 504
2	java.security.KeyPairGenerator$KeyPairGeneratorImpl.generateKeyPair	KeyPairGenerator.java, line 276
3	com.microsoft.aad.adal.StorageHelper.generateKeyPairFromAndroidKeyStore	StorageHelper.java, line 414
4	com.microsoft.aad.adal.StorageHelper.getKeyOrCreate	StorageHelper.java, line 362
5	com.microsoft.aad.adal.StorageHelper.loadSecretKeyForEncryption	StorageHelper.java, line 320
6	com.microsoft.aad.adal.StorageHelper.encrypt	StorageHelper.java, line 178
7	com.microsoft.aad.adal.DefaultTokenCacheStore.encrypt	DefaultTokenCacheStore.java, line 122
8	com.microsoft.aad.adal.DefaultTokenCacheStore.setItem	DefaultTokenCacheStore.java, line 190
9	com.microsoft.aad.adal.TokenCacheAccessor.setItemToCacheForUser	TokenCacheAccessor.java, line 227
10	com.microsoft.aad.adal.TokenCacheAccessor.updateTokenCache	TokenCacheAccessor.java, line 165
11	com.microsoft.aad.adal.AcquireTokenInteractiveRequest.acquireTokenWithAuthCode	AcquireTokenInteractiveRequest.java, line 121
12	com.microsoft.aad.adal.AcquireTokenRequest$4.run	AcquireTokenRequest.java, line 651
[...]

Issue #706, opened 2016-08-29
IllegalArgumentException when trying to create AuthenticationContext in Arabic locale


dbahat:
Seeing the following issue when trying to create AuthenticationContext using the DefaultTokenCacheStore, when the device locale is in Arabic or Farsi.

Possible root cause:  StorageHelper.getKeyPairFromAndroidKeyStore() calls Calendar.getInstance() without specifying a locale.

Getting the following exception:
java.lang.IllegalArgumentException: invalid date string: Unparseable date: "
af`hbiac`a`cGMT+00:00" (at offset 0)
at com.android.org.bouncycastle.asn1.DERUTCTime.<init>(DERUTCTime.java:98)
at com.android.org.bouncycastle.asn1.x509.Time.<init>(Time.java:62)
at com.android.org.bouncycastle.x509.X509V3CertificateGenerator.setNotBefore(X509V3CertificateGenerator.java:112)
at android.security.AndroidKeyPairGenerator.generateKeyPair(AndroidKeyPairGenerator.java:127)
at java.security.KeyPairGenerator$KeyPairGeneratorImpl.generateKeyPair(KeyPairGenerator.java:276)
at com.microsoft.aad.adal.StorageHelper.getKeyPairFromAndroidKeyStore(StorageHelper.java:517)
at com.microsoft.aad.adal.StorageHelper.getSecretKeyFromAndroidKeyStore(StorageHelper.java:426)
at com.microsoft.aad.adal.StorageHelper.loadSecretKeyForAPI(StorageHelper.java:168)
at com.microsoft.aad.adal.DefaultTokenCacheStore.<init>(DefaultTokenCacheStore.java:103)
at com.microsoft.aad.adal.AuthenticationContext.<init>(AuthenticationContext.java:152)

heidijinxujia:
@dbahat Hi, which branch are you using?

dbahat:
Seeing this in release 1.2.0

dbahat:
Seems to be due to an issue in Android. 
Opened issue https://code.google.com/p/android/issues/detail?id=228196

hjymandroid:
Any progress on this bug? It is blocker for us to use ADAL now.

weijjia:
Please check this wiki https://github.com/AzureAD/azure-activedirectory-library-for-android/wiki/Common-Issues-With-AndroidKeyStore

PR #757, opened 2017-01-10
Khalid/fix crash when arabic numbers

khalidaw:
https://github.com/AzureAD/azure-activedirectory-library-for-android/issues/753
https://github.com/AzureAD/azure-activedirectory-library-for-android/issues/706

Seems like its happening because of Android bug.
This PR prevent the crash of the application, but doesn't solve the issue it self.
commented in the code.


hjymandroid:
Instead of fixing it by catch it, could you please just use english when generating the date, like the one here?

https://gitlab.com/fdroid/fdroidclient/commit/04e318c9ca122868c205551fa0af617fe9b2c754

weijjia:
Please check this wiki. https://github.com/AzureAD/azure-activedirectory-library-for-android/wiki/Common-Issues-With-AndroidKeyStore

PR #768, opened 2017-02-07
Add telemetry name/value prefix and replace hyphens with underscores

msolivialuo:


weijjia:
Changes look good to me, but i would like @addev-ashish to take a look at it as it's telemetry related.

addev-ashish:
:shipit:

Issue #707, opened 2016-08-30
User gets prompted for password when no network and uses Broker flow


imakhalova:
Instead of DEVICE_CONNECTION_IS_NOT_AVAILABLE aquire token returns AUTH_REFRESH_FAILED_PROMPT_NOT_ALLOWED or shows UI.

And actually any error that occures in broker app, adal handles as non retryable error...

heidijinxujia:
Hi @imakhalova , could you paste details of the log here? Thanks.

imakhalova:
Repro steps:
acquire token for 1 resource (using broker flow), turn off network, try acquire token for another resource, you get AUTH_REFRESH_FAILED_PROMPT_NOT_ALLOWED

You can collect logs on your own as it has 100% repro.

heidijinxujia:
Hi @imakhalova I've tried the steps on the useappwithbroker.apk with adal-1.10.0 on Android N. I first tried to acquire token for a mam account from resource1 (using broker flow) and get the token back from broker (I install AA here as broker) successfully. And turned off network and tried to acquire token for the same account with resource2. It returns DEVICE_CONNECTION_IS_NOT_AVAILABLE instead of AUTH_REFRESH_FAILED_PROMP_NOT_ALLOWED. Could you give more details of your steps, like test app, adal version and testing account or paste the log here? Because based on the test I did so far, the error code returned is as expected. And the log did not show anything wrong. Thanks.

heidijinxujia:
Do you mean you tried acquireTokenSilent on resource 2 instead of acquireToken? I still didn't get the error you mentioned. Please send me your log. As to the redirectUri, you can see the expected redirectUri from the log if the one you entered is not correct.

imakhalova:
Here is logs:
09-20 15:58:26.693 26769-19525/com.microsoft.skydrive V/PRNGFixes: 2016-09-20 22:58:26-67a47dc2-70c7-45cd-8d67-6041ae2bb18c-No need to apply the fix ver:1.10.0
09-20 15:58:26.693 26769-19525/com.microsoft.skydrive V/[ADAL]PRNGFixes: 2016-09-20 22:58:26-67a47dc2-70c7-45cd-8d67-6041ae2bb18c-No need to apply the fix ver:1.10.0
[...]

PR #782, opened 2017-02-23
Return DEVICE_CONNECTION_IS_NOT_AVAILABLE  error when no network and uses Broker flow

heidijinxujia:
Ref #707 
The aim of this PR is to return the ADALError.DEVICE_CONNECTION_IS_NOT_AVAILABLE instead of AUTH_REFRESH_FAILED_PROMPT_NOT_ALLOWED  when a silentAuth with PRT in broker is called while no network connection available.

To-do
- add unit test
- add version checking

### DIFF
--- a/adal/src/main/java/com/microsoft/aad/adal/BrokerProxy.java
+++ b/adal/src/main/java/com/microsoft/aad/adal/BrokerProxy.java
@@ -22,27 +22,6 @@
 // THE SOFTWARE.
 package com.microsoft.aad.adal;
 
-import java.io.ByteArrayInputStream;
-import java.io.IOException;
-import java.io.InputStream;
-import java.security.GeneralSecurityException;
-import java.security.MessageDigest;
-import java.security.NoSuchAlgorithmException;
-import java.security.cert.CertPath;
-import java.security.cert.CertPathValidator;
-import java.security.cert.CertificateEncodingException;
-import java.security.cert.CertificateException;
-import java.security.cert.CertificateFactory;
-import java.security.cert.PKIXParameters;
-import java.security.cert.TrustAnchor;
-import java.security.cert.X509Certificate;
-import java.util.ArrayList;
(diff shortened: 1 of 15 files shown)


adal/src/main/java/com/microsoft/aad/adal/BrokerProxy.java
```diff
@@ -338,6 +338,9 @@ private Bundle getAuthTokenFromAccountManager(final AuthenticationRequest reques
                 Logger.e(TAG, AUTHENTICATOR_CANCELS_REQUEST, "", ADALError.AUTH_FAILED_CANCELLED, e);
             } catch (AuthenticatorException e) {
                 Logger.e(TAG, AUTHENTICATOR_CANCELS_REQUEST, "", ADALError.BROKER_AUTHENTICATOR_NOT_RESPONDING);
+                if (!StringExtensions.isNullOrBlank(e.getMessage()) && e.getMessage().contains(ADALError.DEVICE_CONNECTION_IS_NOT_AVAILABLE.toString())) {
```
weijjia: What if I make an acquireToken call and no token in the broker cache? or acquireToken call with PromptBehavior as Force_prompt?

weijjia: Also this won't work with the new bound service. Why can't we check network connection right before we talk to broker? in the case of new service, we can check connection when service is on bound.

heidijinxujia: Yes, I will add the network checking at the beginning of talking to broker in the following commit, which will cover both the bound service situation and the old way. 

For Force_prompt.Force_prompt, shouldTrySilentFlow(authenticationRequest) will return false, the auto process will go to the acquireTokenInteractiveFlow() where the network connection is checked at the beginning.

PR #783, opened 2017-02-24
Return DEVICE_CONNECTION_IS_NOT_AVAILABLE error when no network and uses Broker flow

heidijinxujia:
Ref #707 
The aim of this PR is to return the ADALError.DEVICE_CONNECTION_IS_NOT_AVAILABLE instead of AUTH_REFRESH_FAILED_PROMPT_NOT_ALLOWED when a silentAuth with PRT in broker is called while no network connection available.
- get the exception error message from broker and throw the AuthenticationException if the AuthenticatorExceptions is caused by ADALError.DEVICE_CONNECTION_IS_NOT_AVAILABLE
- add corresponding unit test
- add the testGetAuthTokenVerifyNoNetwork into BrokerAccountServiceTest

### DIFF
--- a/adal/src/androidTest/java/com/microsoft/aad/adal/BrokerAccountServiceTest.java
+++ b/adal/src/androidTest/java/com/microsoft/aad/adal/BrokerAccountServiceTest.java
@@ -159,6 +159,29 @@ public final class BrokerAccountServiceTest extends ServiceTestCase<MockBrokerAc
         latch.await();
     }
 
+    public void testGetAuthTokenVerifyNoNetwork() throws InterruptedException {
+        final CountDownLatch latch = new CountDownLatch(1);
+        sThreadExecutor.execute(new Runnable() {
+            @Override
+            public void run() {
+                final Context mockContext = getMockContext();
+                final FileMockContext context = new FileMockContext(mockContext);
+                context.setConnectionAvaliable(false);
+
+                try {
+                    final Bundle bundle = BrokerAccountServiceHandler.getInstance().getAuthToken(context, new Bundle());
+                    fail();
+                } catch (final AuthenticationException e) {
+                    Assert.assertTrue(e.getCode() == ADALError.DEVICE_CONNECTION_IS_NOT_AVAILABLE);
+                } finally {
+                    latch.countDown();
+                }
+            }
+        });
+        latch.await();
+
+    }
+
     @SmallTest
     public void testBrokerProxyGetAuthToken() throws InterruptedException {
         final CountDownLatch latch = new CountDownLatch(1);
--- a/adal/src/androidTest/java/com/microsoft/aad/adal/BrokerProxyTests.java
+++ b/adal/src/androidTest/java/com/microsoft/aad/adal/BrokerProxyTests.java
@@ -673,6 +673,30 @@ public class BrokerProxyTests extends AndroidTestCase {
         }
     }
 
+    public void testGetAuthTokenInBackgroundVerifyErrorNoNetworkConnection() throws NameNotFoundException {
+        final String acctName = "testAcct123";
+        final AuthenticationRequest authRequest = createAuthenticationRequest("https://login.windows.net/omercantest", "resource", "client",
+                "redirect", acctName.toLowerCase(Locale.US), PromptBehavior.Auto, "", UUID.randomUUID(), false);
+        final FileMockContext context = new FileMockContext(getContext());
+        context.setConnectionAvaliable(false);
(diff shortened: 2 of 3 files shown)


adal/src/main/java/com/microsoft/aad/adal/BrokerAccountServiceHandler.java
```diff
@@ -128,9 +128,10 @@ public Bundle getAuthToken(final Context context, final Bundle requestBundle) th
             public void onSuccess(BrokerAccountServiceConnection result) {
                 final IBrokerAccountService brokerAccountService = result.getBrokerAccountServiceProvider();
                 try {
+                    HttpWebRequest.throwIfNetworkNotAvaliable(context);
```
weijjia: Does it work? Aren't we still skipping the access token lookup even it's still valid?

heidijinxujia: Hi @weijjia I've checked in a new commit.

adal/src/main/java/com/microsoft/aad/adal/AcquireTokenRequest.java
```diff
@@ -275,7 +275,17 @@ private AuthenticationResult tryAcquireTokenSilent(final AuthenticationRequest a
 
         if (shouldTrySilentFlow(authenticationRequest)) {
             Logger.v(TAG, "Try to acquire token silently, return valid AT or use RT in the cache.");
-            authenticationResult = acquireTokenSilentFlow(authenticationRequest);
+            try {
```
weijjia: sorry i'm a little bit confused, what are we doing here?

heidijinxujia: When BrokerAccountService is binded and we call acquireToken for an account which doesn't exist in the broker. Adal will try going to look up the account in broker before trying interactive auth. So BrokerAccountService.acquireTokenSilently will be called to get the token in broker. createErrorBundle(ADALError.AUTH_REFRESH_FAILED_PROMPT_NOT_ALLOWED) will be returned to BrokerAccountServiceHandler and invoke the onError() and AuthenticationException(ADALError.AUTH_REFRESH_FAILED_PROMPT_NOT_ALLOWED) will be thrown to the calling method BrokerProxy.getAuthTokenInBackground(). And the exception will popped up all way to tryAcquireTokenSilent() and this is where caused a problem. If we don't catch the exception here, it will throw and reach to the callback.onError in the app before we trying the interavtive auth. We only want to throw the ADALError.AUTH_REFRESH_FAILED_PROMPT_NOT_ALLOWED error when it is a silent call.

heidijinxujia: The point of this fix is that when using the binding service, we did not check if it is a silent call before throwing AuthenticationException(ADALError.AUTH_REFRESH_FAILED_PROMPT_NOT_ALLOWED).

weijjia: You're putting it in the wrong place. I mentioned to you on slack and said there is already a fix checked in for this, but it went to release branch directly. 

Here is the commit for it https://github.com/AzureAD/azure-activedirectory-library-for-android/commit/e6f34a208260a3af6f1fa13027e8148153998cb1

heidijinxujia: Is it OK to commit this branch into the release branch? Or merge the release branch here and keep committing into dev? @weijjia

heidijinxujia: Agree on the commit. I was also thinking to add the request parameter into getResultFromBrokerResponse. But I was worried to change the function parameters. It's cool that you already have the fix.

weijjia: I think i explained it in the slack channel. I was going to release the contacts permission change and found bugs when testing it. For all the bugs found during the testing should go to release branch and once all the validations are done, we merge it back. 

For your change, it's something came later and it's not on top of the release branch. That's why I'm asking you to send it to dev and we squash it, we'll cherry pick the change from dev to release branch. 

Does it make sense?

heidijinxujia: Cool. I can remove the try-catch here and squash the commits.

heidijinxujia:
Squashed all the commits into one.

weijjia:
This one looks good to me. :shipit:

PR #786, opened 2017-03-07
update broker signature in the test to mock the broker app

weijjia:
update broker signature in the test to mock broker call.


addev-ashish:
:shipit:

PR #789, opened 2017-03-09
Fix the unit test testGetAuthTokenVerifyNoNetwork

heidijinxujia:


iambmelt:
I say :shipit: -- currently sleuthing `testKeyPair()` oddness with @heidijinxujia but that issue appears unrelated

Issue #793, opened 2017-03-14
Javadoc build is broken after migration to Java8


addev-ashish:
The javadoc generated is 1KB in size.

PR #794, opened 2017-03-14
Working around java8 enforcements around javadoc #793

addev-ashish:
Java8 has a lot stricter checks of javadoc, adding the xdoclint to none.
Also seems the exclude syntax has been changed.

### DIFF
--- a/adal/build.gradle
+++ b/adal/build.gradle
@@ -140,7 +140,13 @@ task sourcesJar(type: Jar) {
 task javadoc(type: Javadoc) {
     source = android.sourceSets.main.java.srcDirs
     exclude '**/*.aidl'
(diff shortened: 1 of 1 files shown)


adal/build.gradle
```diff
@@ -139,7 +139,8 @@ task sourcesJar(type: Jar) {
 
 task javadoc(type: Javadoc) {
     source = android.sourceSets.main.java.srcDirs
-    exclude {'*.aidl'}
+    exclude '**/*.aidl'
+    options.addStringOption('Xdoclint:none', '-quiet')
```
weijjia: you should add it only when the current java version is 8.

weijjia:
🚢 it :shipit:

Issue #712, opened 2016-09-04
Telemetry: When you press back on MFA failure, you get a webview cancel with err_cache_miss error and apps dont know how to handle it


khalidaw:
We see this error a lot in telemetry.

What is the reason for such error and crash (beside changing orientation that we know about)?

heidijinxujia:
Hi, where did you see this error? I didn't find it in the Telemetry PR. Thanks.

weijjia:
@heidijinxujia I don't think they're talking about the telemetry PR, they observed the error occurrences in their telemetry system. 

@khalidaw err_cache_miss could happen when device running on low memory and the system kills the AuthenticationActivity when it's going background or . If this happens, and activity is re-initialized, it will be restored from the saved state as well as the webview. @heidijinxujia Can you try to repro it if turning on "Don't not keep activities" in the developer options?

sdabbiru:
CP telemetry indicates that over 2.5K hit for this error - perhaps a lot of users do MFA. We should address this soon.

imakhalova:
Our telemetry shows this error as well.
This message within server response Error Code:-1 net::ERR_CACHE_MISS
sample CorrelationId f6daed14-5831-43ce-9863-d19cf87901f0

hjymandroid:
We also see a ton of this 
`com.microsoft.aad.adal.AuthenticationContext.waitingRequestOnError(AuthenticationContext.java:1026)
\tat com.microsoft.aad.adal.AuthenticationContext.onActivityResult(AuthenticationContext.java:900)
\tat com.yammer.droid.ui.login.LoginFragment.onMAMActivityResult(LoginFragment.java:124)
\tat com.microsoft.intune.mam.client.app.offline.OfflineFragmentBehavior.onActivityResult(OfflineFragmentBehavior.java:134)
\tat com.microsoft.intune.mam.client.support.v4.app.MAMFragment.onActivityResult(MAMFragment.java:85)
\tat android.support.v4.app.FragmentActivity.onActivityResult(FragmentActivity.java:167)
\tat com.microsoft.intune.mam.client.support.v7.app.MAMAppCompatActivityBase.onActivityResultReal(MAMAppCompatActivityBase.java:337)
\tat com.microsoft.intune.mam.client.app.offline.OfflineActivityBehavior.onMAMActivityResult(OfflineActivityBehavior.java:382)
[...]

PR #796, opened 2017-03-15
Fixing err_cache_miss error #712

addev-ashish:
If the auth flow in the webview navigates the user to a MFA page, and the user is unable to answer the MFA prompt, subsequently the user hits the back button, this will lead to an err_cache_miss as the page is attempted to be loaded from the cache. Setting LOAD_CACHE_ELSE_NETWORK let the page to be fetched again from the server.


weijjia:
change looks good. But can we run the set of tests we talked today before merging it?

PR #797, opened 2017-03-17
Add new trusted authority host

heidijinxujia:
Ref #787 
- add new trusted authority host login.microsoftonline.us


addev-ashish:
:shipit:

PR #801, opened 2017-03-24
Fix for handling the permission missing scenario when silent auth is called directly

heidijinxujia:
Fix for handling the permission missing scenario when performing an acquireTokenSilent call.
1. there is a bug that we never check if permissions are granted to use broker when calling acquireTokenSilent directly. Because we assume it will alway do the interactive first where the permission checking is performed. However, this will broken the token share scenario where the acquireTokenSilent is called directly.
2. for the silent call, the flow is to look into the local cache first then go to the broker. So we should check the broker permission only when adal doesn't get any luck in local cache and tries to acquire token silently with broker.

Follow-up question,
1. For both Android <=22 and >=23, should we keep it same checking behavior for the silent call?
2. If BrokerAccountService is supported, then GET_ACCOUNTS permission is no longer needed for both  <=22 and >=23 or not?

### DIFF
--- a/adal/src/androidTest/java/com/microsoft/aad/adal/AcquireTokenRequestTest.java
+++ b/adal/src/androidTest/java/com/microsoft/aad/adal/AcquireTokenRequestTest.java
@@ -789,8 +789,8 @@ public final class AcquireTokenRequestTest extends AndroidTestCase {
         final ITokenCacheStore cacheStore = getTokenCache(getExpireDate(-MINUS_MINUITE), true, true, getExpireDate(EXTEND_MINUS_MINUTE));
         cacheStore.removeItem(CacheKey.createCacheKeyForRTEntry(VALID_AUTHORITY, "resource", "clientId", TEST_USERID));
         cacheStore.removeItem(CacheKey.createCacheKeyForRTEntry(VALID_AUTHORITY, "resource", "clientId", TEST_UPN));
-        cacheStore.getItem(CacheKey.createCacheKeyForMRRT(VALID_AUTHORITY,"clientId", TEST_USERID)).setFamilyClientId(AuthenticationConstants.MS_FAMILY_ID);
-        cacheStore.getItem(CacheKey.createCacheKeyForMRRT(VALID_AUTHORITY,"clientId", TEST_UPN)).setFamilyClientId(AuthenticationConstants.MS_FAMILY_ID);
+        cacheStore.getItem(CacheKey.createCacheKeyForMRRT(VALID_AUTHORITY, "clientId", TEST_USERID)).setFamilyClientId(AuthenticationConstants.MS_FAMILY_ID);
+        cacheStore.getItem(CacheKey.createCacheKeyForMRRT(VALID_AUTHORITY, "clientId", TEST_UPN)).setFamilyClientId(AuthenticationConstants.MS_FAMILY_ID);
         
         final FileMockContext mockContext = createMockContext();
         final AuthenticationContext authContext = new AuthenticationContext(mockContext,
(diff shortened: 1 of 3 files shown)


adal/src/main/java/com/microsoft/aad/adal/AcquireTokenRequest.java
```diff
@@ -338,9 +340,29 @@ private AuthenticationResult acquireTokenSilentFlow(final AuthenticationRequest
...
                 || !mBrokerProxy.verifyUser(authenticationRequest.getLoginHint(), authenticationRequest.getUserId())) {
             return authResult;
+        } else if (switchToBrokerFlag == BrokerProxy.SwitchToBroker.NEED_PERMISSIONS_TO_SWITCH_TO_BROKER) {
+            //For android M and above
+            throw new UsageAuthenticationException(
+                        ADALError.DEVELOPER_BROKER_PERMISSIONS_MISSING,
+                        "Broker related permissions are missing for GET_ACCOUNTS");
+        }
+
+        //check if App gives permissions of broker usage
```
weijjia: why we're checking again? Didn't the else above already throw the permission missing error?

adal/src/main/java/com/microsoft/aad/adal/AcquireTokenRequest.java
```diff
@@ -737,6 +759,19 @@ private void waitingRequestOnError(final CallbackHandler handler, final Authenti
         }
     }
 
+    private String checkPermission(final String permissionName) {
```
weijjia: we don't have any existing permission checking method?

adal/src/main/java/com/microsoft/aad/adal/AcquireTokenRequest.java
heidijinxujia: BrokerProxy.canSwitchToBroker() only checks the GET_ACCOUNTS permission for Android 23+ which need to be granted when app is running. But we also need to check permissions for Android 22 and below, which are set in the manifest. We did not check the permissions required for both Android 23+ ans Android 22-. My current concern is if we should check permissions for Android 22- at the very beginning of the acquireTokenSilent call, because these permissions need to be set by the developer in the manifest if using broker.

adal/src/main/java/com/microsoft/aad/adal/AcquireTokenRequest.java
heidijinxujia: We do have this checkPermission() function in BrokerProxy. It is a private method belong to BrokerProxy class. However,I have a concern to change it to package visible. 
AcquireTokenRequest.mBrokerProxy is the object of Interface IBrokerProxy. If we change checkPermission to package visiable, to call this function, we could do
    BrokerProxy object = (BrokerProxy) mBrokerProxy;
and call brokerproxy.checkPermission().

nazukj:
can you please ensure a test case is added for this scenario? thank you :). Also is the issue tracked with a github/VSTS issue?

adal/src/androidTest/java/com/microsoft/aad/adal/AcquireTokenRequestTest.java
```diff
@@ -873,6 +873,54 @@ public void testVerifyManifestPermissionMissingMultiPermissions() throws Interru
...
+        when(mockContext.getPackageManager().checkPermission(Mockito.refEq("android.permission.GET_ACCOUNTS"),
+                Mockito.anyString())).thenReturn(PackageManager.PERMISSION_DENIED);
+        AuthenticationSettings.INSTANCE.setUseBroker(true);
+        final AuthenticationContext authContext = new AuthenticationContext(mockContext,
+                VALID_AUTHORITY, false);
+        try {
+            authContext.acquireTokenSilentSync("resource", "clientid", "userid");
+            fail("Expect an exception");
+        } catch (final Exception exception) {
+            assertTrue(exception instanceof AuthenticationException);
```
addev-ashish: >assertTrue(exception instanceof AuthenticationException); [](start = 12, length = 57)

you don't need to do this, catch AuthenticationException, let the others fall through

adal/src/androidTest/java/com/microsoft/aad/adal/BrokerAccountServiceTest.java
```diff
@@ -159,14 +159,14 @@ public void run() {
...
 
-    public void testGetAuthTokenVerifyNoNetwork() throws InterruptedException, AuthenticatorException,OperationCanceledException, IOException {
+    public void testGetAuthTokenVerifyNoNetwork() throws InterruptedException, AuthenticatorException, OperationCanceledException, IOException {
         final CountDownLatch latch = new CountDownLatch(1);
         sThreadExecutor.execute(new Runnable() {
             @Override
             public void run() {
                 final Context mockContext = getMockContext();
                 Bundle requestBundle = new Bundle();
-                requestBundle.putString("isConnectionAvaliable","false");
```
addev-ashish: not a part of this change, please revert this file.

adal/src/androidTest/java/com/microsoft/aad/adal/MockBrokerAccountService.java
```diff
@@ -76,7 +76,7 @@ public synchronized Bundle getBrokerUsers() throws RemoteException {
         @Override
         public synchronized Bundle acquireTokenSilently(Map requestParameters) throws RemoteException {
             final Bundle bundle = new Bundle();
-            if(requestParameters.containsKey("isConnectionAvaliable")) {
+            if (requestParameters.containsKey("isConnectionAvaliable")) {
```
addev-ashish: please revert.

heidijinxujia: OK

adal/src/androidTest/java/com/microsoft/aad/adal/BrokerAccountServiceTest.java
heidijinxujia: OK

adal/src/androidTest/java/com/microsoft/aad/adal/AcquireTokenRequestTest.java
```diff
@@ -873,6 +873,54 @@ public void testVerifyManifestPermissionMissingMultiPermissions() throws Interru
         }
     }
 
+    public void testSilentRequestMissingPermissionHandlingForAndroid22andBelow() throws InterruptedException, PackageManager.NameNotFoundException {
+        if (Build.VERSION.SDK_INT < Build.VERSION_CODES.M) {
```
heidijinxujia: I've tried. But I don't think we could do a mock on the either PackageManager.PackageInfo.versionCode or Build.VERSION.SDK_INT. Because they are both int variables not methods, and they don't have getters for us to pass the mock.

addev-ashish:
:shipit:

PR #802, opened 2017-03-28
Expand Resiliency Error Codes to all 5xx Status Codes

heidijinxujia:
ISSUE #772
- expand resiliency error codes to all 5xx status codes.
- correpsonding change on the unit tests in AcquireTokenRequestTest.java

### DIFF
--- a/adal/src/androidTest/java/com/microsoft/aad/adal/AcquireTokenRequestTest.java
+++ b/adal/src/androidTest/java/com/microsoft/aad/adal/AcquireTokenRequestTest.java
@@ -613,7 +613,7 @@ public final class AcquireTokenRequestTest extends AndroidTestCase {
         Mockito.when(mockedConnection.getOutputStream()).thenReturn(Mockito.mock(OutputStream.class));
         Mockito.when(mockedConnection.getInputStream()).thenReturn(Util.createInputStream(Util.getErrorResponseBody("HTTP_GATEWAY_TIMEOUT")),
                 Util.createInputStream(Util.getErrorResponseBody("HTTP_BAD_GATEWAY")));
-        Mockito.when(mockedConnection.getResponseCode()).thenReturn(HttpURLConnection.HTTP_GATEWAY_TIMEOUT, HttpURLConnection.HTTP_BAD_GATEWAY);
+        Mockito.when(mockedConnection.getResponseCode()).thenReturn(HttpURLConnection.HTTP_GATEWAY_TIMEOUT, HttpURLConnection.HTTP_NOT_FOUND);
 
         try {
             authContext.acquireTokenSilentSync("resource", "clientid", TEST_USERID);
@@ -765,8 +765,8 @@ public final class AcquireTokenRequestTest extends AndroidTestCase {
         HttpUrlConnectionFactory.setMockedHttpUrlConnection(mockedConnection);
         Util.prepareMockedUrlConnection(mockedConnection);
         Mockito.when(mockedConnection.getOutputStream()).thenReturn(Mockito.mock(OutputStream.class));
-        Mockito.when(mockedConnection.getInputStream()).thenReturn(Util.createInputStream(Util.getErrorResponseBody("HTTP_BAD_GATEWAY")));
-        Mockito.when(mockedConnection.getResponseCode()).thenReturn(HttpURLConnection.HTTP_BAD_GATEWAY);
+        Mockito.when(mockedConnection.getInputStream()).thenReturn(Util.createInputStream(Util.getErrorResponseBody("HTTP_CONFLICT")));
+        Mockito.when(mockedConnection.getResponseCode()).thenReturn(HttpURLConnection.HTTP_NOT_FOUND);
 
         try {
             authContext.acquireTokenSilentSync("resource", "clientid", TEST_USERID);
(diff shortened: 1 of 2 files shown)


adal/src/main/java/com/microsoft/aad/adal/Oauth2.java
```diff
@@ -629,25 +631,22 @@ private AuthenticationResult processTokenResponse(HttpWebResponse webResponse, f
...
-            case HttpURLConnection.HTTP_INTERNAL_ERROR:
-            case HttpURLConnection.HTTP_GATEWAY_TIMEOUT:
-            case HttpURLConnection.HTTP_UNAVAILABLE:
-                throw new ServerRespondingWithRetryableException("Unexpected server response " + webResponse.getBody());
-            default:
-                throw new AuthenticationException(ADALError.SERVER_ERROR, "Unexpected server response " + webResponse.getBody());
+            } catch (final JSONException jsonException) {
+                throw new AuthenticationException(ADALError.SERVER_INVALID_JSON_RESPONSE, "Can't parse server response " + webResponse.getBody(), jsonException);
+            }
+        } else if (statusCode >= HttpURLConnection.HTTP_INTERNAL_ERROR && statusCode < MAX_RESILIENCY_ERROR_CODE) {
```
addev-ashish: Add a comment on why we are doing this.

addev-ashish:
:shipit:

adal/src/androidTest/java/com/microsoft/aad/adal/AcquireTokenRequestTest.java
```diff
@@ -765,8 +765,8 @@ public void testResiliencyTokenReturnExtendedLifetimeOnwithoutRetry() throws Pac
         HttpUrlConnectionFactory.setMockedHttpUrlConnection(mockedConnection);
         Util.prepareMockedUrlConnection(mockedConnection);
         Mockito.when(mockedConnection.getOutputStream()).thenReturn(Mockito.mock(OutputStream.class));
-        Mockito.when(mockedConnection.getInputStream()).thenReturn(Util.createInputStream(Util.getErrorResponseBody("HTTP_BAD_GATEWAY")));
-        Mockito.when(mockedConnection.getResponseCode()).thenReturn(HttpURLConnection.HTTP_BAD_GATEWAY);
+        Mockito.when(mockedConnection.getInputStream()).thenReturn(Util.createInputStream(Util.getErrorResponseBody("HTTP_CONFLICT")));
```
addev-ashish: >HTTP_CONFLICT [](start = 117, length = 13)

just curious, what is this?

adal/src/androidTest/java/com/microsoft/aad/adal/AcquireTokenRequestTest.java
```diff
@@ -613,7 +613,7 @@ public void testResiliencyTokenReturnExtendedLifetimeOnwithRetryFail() throws Pa
         Mockito.when(mockedConnection.getOutputStream()).thenReturn(Mockito.mock(OutputStream.class));
         Mockito.when(mockedConnection.getInputStream()).thenReturn(Util.createInputStream(Util.getErrorResponseBody("HTTP_GATEWAY_TIMEOUT")),
                 Util.createInputStream(Util.getErrorResponseBody("HTTP_BAD_GATEWAY")));
-        Mockito.when(mockedConnection.getResponseCode()).thenReturn(HttpURLConnection.HTTP_GATEWAY_TIMEOUT, HttpURLConnection.HTTP_BAD_GATEWAY);
+        Mockito.when(mockedConnection.getResponseCode()).thenReturn(HttpURLConnection.HTTP_GATEWAY_TIMEOUT, HttpURLConnection.HTTP_NOT_FOUND);
```
addev-ashish: >HTTP_NOT_FOUND [](start = 126, length = 14)

why this change?

adal/src/androidTest/java/com/microsoft/aad/adal/AcquireTokenRequestTest.java
heidijinxujia: This is a non 5xx status code mock. To check the AuthenticationException instead  of  ServerRespondingWithRetryableException is thrown when adal get a non 5xx response.

adal/src/androidTest/java/com/microsoft/aad/adal/AcquireTokenRequestTest.java
heidijinxujia: HTTP_BAD_GATEWAY is 502 which was not treated as retryable status code before. As we have expend the code to all 5xx, HTTP_NOT_FOUND 404 is used here to mock a non-retryable server error code here.

adal/src/main/java/com/microsoft/aad/adal/Oauth2.java
heidijinxujia: Hi @kgidewall is there any specific reason that why we change the resiliency logic to retry all 5xx? Thanks.

kgidewall: We had a discussion across teams and platforms (ADAL, WAM, server, etc.) and decided that it made more sense to cover all network errors, and not just a few specific ones. All platforms are now changing to this.

kgidewall: Is the change here just to replace a 'case' statement with an 'if' statement?  Why are we doing that?

kgidewall: Just a nit here, but I would prefer that you set MAX_RESILIENCY_ERROR_CODE = 599 and make the if statement be "statusCode <= MAX_RESILIENCY_ERROR_CODE". I think it makes the definition of MAX_RESILIENCY_ERROR_CODE more correct.

Issue #809, opened 2017-03-30
loadSecretKeyForDecryption uses the cached key without checking the blob


addev-ashish:
If there are multiple blobs and one is encrypted with user defined key and the other is encrypted with keystore key, the attempt to load the user defined key will always fail.

addev-ashish:
https://github.com/AzureAD/azure-activedirectory-library-for-android/pull/810

PR #810, opened 2017-03-30
moving the check for the class variable that holds secretkey to later #809

addev-ashish:
We should only read the AndroidKeystore key if we know that the blob was encrypted using that. If the key was user defined we should use getSecretKey to get the key.


weijjia:
I think the change is safe. One thing you can try for repro is, instead of setting the key during onCreate, try to change it per acquire token request in the test app itself.

Issue #816, opened 2017-04-11
Resiliency error code range should be [500, 599]


weijjia:
The current code is including 600 as the max range of resiliency error code, while the max should be 599. Also please add unit tests for both the min and max error codes.

addev-ashish:
fix merged into Dev.

PR #817, opened 2017-04-11
Fix the resiliency maximum server error range to 599

heidijinxujia:
issue #816 
- fix the resiliency server error range maximum to 599.
- add the unit tests for the min status code 500 and max status code 599


heidijinxujia:
Thank you, @kgidewall I've addressed your comment in the latest commit.

PR #825, opened 2017-05-02
Update Azure Authenticator app signature.

HSAR:


adal/src/main/java/com/microsoft/aad/adal/AuthenticationConstants.java
```diff
@@ -451,7 +451,7 @@ private AuthenticationConstants() { }
          * Signature info for Azure authenticator app that installs authenticator
          * component.
          */
-        public static final String AZURE_AUTHENTICATOR_APP_SIGNATURE = "ho040S3ffZkmxqtQrSwpTVOn9r0=";
```
mjcheetham: Why has this changed?

HSAR: In order to automate webviews in an Android APK, a setting to allow remote webview debugging has to be turned on. My assumption is that building with that setting turned on changed the signature.

heidijinxujia: The authenticator signature is hard-coded which should be the same with the sign of prod build of Authenticator. The change you are trying to push will effect all other dev who are using our library. So it could not change here.

heidijinxujia: And this change is only for automation right? @HSAR

HSAR:
@heidijinxujia Ah, you're absolutely right - I shouldn't change this.

adal/src/main/java/com/microsoft/aad/adal/AuthenticationConstants.java
HSAR: @heidijinxujia Right. Although it should be noted that msauth automation won't work without this change.

HSAR: I advise taking this fix into a private branch, building it, and using that APK.

Issue #580, opened 2016-04-08
StorageHelperTests:testKeyPair test can't work.


RandalliLama:
This test is currently failing for me, and I don't see how it could have ever worked.  There doesn't appear to be anything in the test file that would place any keypairs in the KeyStore.  The setup method passes in an explicit key which causes StorageHelper to bypass creating and storing a key in the KeyStore.

The only way that I can see that this might have worked is if it is run on a device where a keypair was previously created by a test app.

I am going to disable it for now.

heidijinxujia:
Merged into dev branch.

PR #830, opened 2017-05-03
Fix for failing unit tests

heidijinxujia:
There are two failing tests.
- testKeyPair()
    Filed on issue #580. The failing is caused by the secretKey setting in the setup which caused the conflict when loadSecretKeyForEncryption(). It is expected to return null from AuthenticationSettings.INSTANCE.getSecretKeyData() and created the key pair by calling generateKeyPairFromAndroidKeyStore().
- testCreateFromResourceUrlInvalidFormat()
    The fail is caused by the network connection, the test passed when the test device is connected to WiFi. And get failed otherwise. The fix is to add the HTTP response mocking.

### DIFF
--- a/adal/src/androidTest/java/com/microsoft/aad/adal/AuthenticationParamsTests.java
+++ b/adal/src/androidTest/java/com/microsoft/aad/adal/AuthenticationParamsTests.java
@@ -24,6 +24,7 @@
 package com.microsoft.aad.adal;
 
 import java.io.IOException;
+import java.io.OutputStream;
 import java.lang.reflect.InvocationTargetException;
 import java.lang.reflect.Method;
 import java.net.HttpURLConnection;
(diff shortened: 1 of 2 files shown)


addev-ashish:
please add a better description to the PR.

weijjia:
ah, i think we should suppress the tests.

addev-ashish:
fix the branch warning above then :shipit:

heidijinxujia:
@HSAR @addev-ashish added you guys into the reviewers, please approve the PR. Then I can merge it.

HSAR:
CI now failing.

HSAR:
testKeyPair is now failing again.

weijjia:
you need to add the suppress annotation.

weijjia:
@heidijinxujia suppress is not deleting the test. Can you add @Suppress annotation?

HSAR:
If you are deciding to delete or ignore a test, be sure to make a note in the relevant work item or test case explaining why.

heidijinxujia:
Moved to another PR. Because the commits cannot be squashed.

PR #836, opened 2017-05-18
Weij/multiple tokens user not provided

weijjia:
if silent request doesn't specify user and there are multiple tokens existed for the given app, and resource, we should fail.


adal/src/androidTest/java/com/microsoft/aad/adal/AcquireTokenRequestTest.java
```diff
@@ -398,6 +399,85 @@ public void testLocalSilentFailedBrokerSilentReturnErrorCannotTryWithInteractive
...
+                assertTrue(authenticationException.getCode().equals(ADALError.AUTH_FAILED_USER_MISMATCH));
+                latch.countDown();
+            }
+        });
+        latch.await();
+    }
+
+    @Test
+    public void testMutipleMRRTExistForTheSameApp() throws PackageManager.NameNotFoundException, InterruptedException {
+        final ITokenCacheStore cacheStore = getTokenCache(getExpireDate(MINUS_MINUITE), true, false, null);
```
jaredsburrows: 👍 final

adal/src/main/java/com/microsoft/aad/adal/TokenCacheAccessor.java
```diff
@@ -258,6 +262,34 @@ void removeTokenCacheItem(final TokenCacheItem tokenCacheItem, final String reso
         Telemetry.getInstance().stopEvent(mTelemetryRequestId, cacheEvent,
                 EventStrings.TOKEN_CACHE_DELETE);
     }
+
+    boolean isMultipleRTsMatchingGivenAppAndResource(final String clientId, final String resource) {
```
addev-ashish: I think we should look at merging these 2 functions together, pass null as the resource when you don't need resource.

adal/src/main/java/com/microsoft/aad/adal/TokenCacheAccessor.java
```diff
@@ -348,6 +380,13 @@ private boolean isUserMisMatch(final String user, final TokenCacheItem tokenCach
         return !user.equalsIgnoreCase(tokenCacheItem.getUserInfo().getDisplayableId()) 
                 && !user.equalsIgnoreCase(tokenCacheItem.getUserInfo().getUserId());
     }
+
+    private void throwIfMultipleATExisted(final String clientId, final String resource, final String user) throws AuthenticationException {
```
addev-ashish: >throwIfMultipleATExisted [](start = 17, length = 24)

you should make this package scope and call it from AcquireTokenSilentHandler

adal/src/main/java/com/microsoft/aad/adal/TokenCacheAccessor.java
```diff
@@ -258,6 +262,34 @@ void removeTokenCacheItem(final TokenCacheItem tokenCacheItem, final String reso
...
+            if (tokenCacheItem.getAuthority().equalsIgnoreCase(mAuthority) && tokenCacheItem.getClientId().equalsIgnoreCase(clientId)
+                    && resource.equalsIgnoreCase(tokenCacheItem.getResource()) && !tokenCacheItem.getIsMultiResourceRefreshToken()) {
+                regularRTsMatchingRequest.add(tokenCacheItem);
+            }
+        }
+
+        return regularRTsMatchingRequest.size() > 1;
+    }
+
+    boolean isMultipleMRRTsMatchingGivenApp(final String clientId) {
```
addev-ashish: >isMultipleMRRTsMatchingGivenApp [](start = 12, length = 31)

following making throwif* function pacakage scope you should make these 2 private.

adal/src/main/java/com/microsoft/aad/adal/TokenCacheAccessor.java
weijjia: nah, that will make the lookup logic kinda messy.

adal/src/main/java/com/microsoft/aad/adal/TokenCacheAccessor.java
weijjia: why?

adal/src/main/java/com/microsoft/aad/adal/TokenCacheAccessor.java
weijjia: um, i cannot. They're used directly by AcquireTokenHandler. Let's talk tomorrow

PR #837, opened 2017-05-19
Update the manifest for the Testapp

heidijinxujia:
All PR suddenly show "VSTS: ADAL Android CI - Waiting for status to be reported" which prevent them from merging. This PR is used to triage and fix the VSTS CI issue.

### DIFF
--- a/testapp/src/main/AndroidManifest.xml
+++ b/testapp/src/main/AndroidManifest.xml
@@ -5,8 +5,8 @@
     android:versionName="1.0" >
 
     <uses-sdk
-        android:minSdkVersion="9"
-        android:targetSdkVersion="19" />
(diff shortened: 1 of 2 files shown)


testapp/src/main/AndroidManifest.xml
```diff
@@ -5,8 +5,8 @@
     android:versionName="1.0" >
 
     <uses-sdk
-        android:minSdkVersion="9"
-        android:targetSdkVersion="19" />
+        android:minSdkVersion="14"
+        android:targetSdkVersion="22"/>
```
jaredsburrows: Please always set the `targetSdkVersion` to the current stable release of Android, API 25.

PR #841, opened 2017-05-25
Update Gson to 2.8.0

jaredsburrows:
@addev-ashish 
Closes #838.

PR #847, opened 2017-06-02
VSTS CI check for #841

heidijinxujia:
Trying a solution to trigger the VSTS CI for forked pull request.


heidijinxujia:
fyi. VSTS: ADAL Android CI (Development) failed because VSTS CI failed to installed the testapp. I'm investigating in it now.

PR #851, opened 2017-06-05
update travis to enable testing running with emulator 

weijjia:
1. add emulator wake up for travis and enable unit tests running with emulator 
2. enable gradle cache. 
3. remove dependency for the test app

### DIFF
--- a/adal/src/androidTest/java/com/microsoft/aad/adal/AndroidTestHelper.java
+++ b/adal/src/androidTest/java/com/microsoft/aad/adal/AndroidTestHelper.java
@@ -24,6 +24,7 @@
 package com.microsoft.aad.adal;
 
 import android.annotation.SuppressLint;
+import android.content.Context;
 import android.content.pm.PackageInfo;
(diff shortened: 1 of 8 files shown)


adal/src/androidTest/java/com/microsoft/aad/adal/AndroidTestHelper.java
```diff
@@ -56,8 +57,9 @@ protected void setUp() throws Exception {
         System.setProperty("dexmaker.dexcache", getInstrumentation().getTargetContext().getCacheDir().getPath());
 
         // ADAL is set to this signature for now
```
iambmelt: Should this comment be updated / removed no that the testapp isn't being used?

weijjia: We're still setting signatures. Instead of using the hardcoded test app name, we're using the auto-generated app.

jaredsburrows:
Great work!

PR #859, opened 2017-06-08
Clean up the Checkstyle and Lint

heidijinxujia:
Clean up the errors and warnings reported from Checkstyle and Lint.

### DIFF
--- a/adal/build.gradle
+++ b/adal/build.gradle
@@ -67,6 +67,7 @@ android {
         disable 'GradleDependency'
         disable 'GradleOverrides'
         disable 'OldTargetApi'
+        disable 'PackageManagerGetSignatures'
     }
 
     libraryVariants.all { variant ->
--- a/adal/src/main/java/com/microsoft/aad/adal/JWSBuilder.java
+++ b/adal/src/main/java/com/microsoft/aad/adal/JWSBuilder.java
@@ -59,7 +59,7 @@ class JWSBuilder implements IJWSBuilder {
     /**
      * Payload for JWS.
      */
-    class Claims {
+    final class Claims {
         @SerializedName("aud")
         private String mAudience;
 
@@ -70,7 +70,7 @@ class JWSBuilder implements IJWSBuilder {
         private String mNonce;
 
         /**
-         * No args constructor for use in serialization for Gson to prevent usage of sun.misc.Unsafe
+         * No args constructor for use in serialization for Gson to prevent usage of sun.misc.Unsafe.
          */
         @SuppressWarnings("unused")
         private Claims() {
@@ -80,7 +80,7 @@ class JWSBuilder implements IJWSBuilder {
     /**
      * Header that includes algorithm, type, thumbprint, keys, and keyid.
      */
-    class JwsHeader {
+    final class JwsHeader {
         @SerializedName("alg")
         private String mAlgorithm;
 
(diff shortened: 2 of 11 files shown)


adal/build.gradle
```diff
@@ -67,6 +67,7 @@ android {
         disable 'GradleDependency'
         disable 'GradleOverrides'
         disable 'OldTargetApi'
+        disable 'PackageManagerGetSignatures'
```
iambmelt: This addition is unclear to me. @heidijinxujia What's the context for this one? Or the warning it triggers?

iambmelt:
Just one question about the disabled check in the gradle file. The rest looks good to me 👍

adal/build.gradle
heidijinxujia: PackageManagerGetSignatures detects the Potential Multiple Certificate Exploit. This is reported in BrokerAccountServiceTest.java Line 280 which reading app signatures from getPackageInfo. 

Here is some description. 
"Improper validation of app signatures could lead to issues where a malicious app submits itself to the Play Store with both its real certificate and a fake certificate and gains access to functionality or information it shouldn't have due to another application only checking for the fake certificate and ignoring the rest. Please make sure to validate all signatures returned by this method. "

```
    private SignatureData getSignature(final Context context, final String packageName)
            throws PackageManager.NameNotFoundException, NoSuchAlgorithmException {
        PackageInfo info = context.getPackageManager().getPackageInfo(packageName,
                PackageManager.GET_SIGNATURES);

        // Broker App can be signed with multiple certificates. It will look
        // all of them
        // until it finds the correct one for ADAL broker.
        byte[] signatureByte;
        String signatureTag;
        for (final Signature signature : info.signatures) {
            signatureByte = signature.toByteArray();
            MessageDigest md = MessageDigest.getInstance("SHA");
            md.update(signatureByte);
            signatureTag = Base64.encodeToString(md.digest(), Base64.NO_WRAP);
            return new SignatureData(new Signature(signatureByte), signatureTag);
        }

        return null;
    }
```
I suppressed this warning because we need to pass the signature here for testing. The signature is already validated in the following for-loop. And I am not sure why it is still reporting this warning. @iambmelt Please let me know if you have another solution for this. :)

nazukj: i suspect this could be happening because the method returns SignatureData and returns Null (2 different return statements) instead of one consolidates one. On a whim, can you try refactoring (locally) to combine both returns into one... (i.e. declare a variable with null to start with and then set to the correct value in for loop with just one return outside the loop) and then running the tool again to see if exception still occurs?

---
In reply to: [121008721](https://github.com/AzureAD/azure-activedirectory-library-for-android/pull/859#discussion_r121008721) [](ancestors = 121008721)

heidijinxujia: @iambmelt ^

iambmelt: I don't actually see _any_ validation happening here. Is it just me, or does this method simply return whatever info.signatures[0] contains?

nazukj:
:clock1:

adal/src/main/java/com/microsoft/aad/adal/JWSBuilder.java
```diff
@@ -59,7 +59,7 @@
     /**
      * Payload for JWS.
      */
-    class Claims {
+    final class Claims {
```
addev-ashish: why the signature change as a part of this change?

Issue #863, opened 2017-06-09
Telemetry: acquireTokenSilentSync creates a dummy api event


addev-ashish:
one extra api event is created if this api is used.

addev-ashish:
#864

Issue #853, opened 2017-06-06
DNS Assistance Causes Unit Test Failure


iambmelt:
Some ISPs provide [DNS Assistance ](https://www.verizon.com/support/consumer/internet/opt-out-of-dns-assist)when hostname lookups fail. In the case of an ISP or network DNS providing such assistance, [`WebRequestHandlerTest#testNonExistentUrl()` ](https://github.com/AzureAD/azure-activedirectory-library-for-android/blob/dev/adal/src/androidTest/java/com/microsoft/aad/adal/WebRequestHandlerTests.java#L179-L190)can fail (by actually resolving to _something_) because of DNS-side redirection. While this may not be an issue for CI - some users/contributors may see this test fail on their network.

```java
junit.framework.AssertionFailedError: Unreachable host, should throw IOException
at junit.framework.Assert.fail(Assert.java:50)
at com.microsoft.aad.adal.WebRequestHandlerTests.testNonExistentUrl(WebRequestHandlerTests.java:185)
at java.lang.reflect.Method.invoke(Native Method)
at android.test.InstrumentationTestCase.runMethod(InstrumentationTestCase.java:220)
at android.test.InstrumentationTestCase.runTest(InstrumentationTestCase.java:205)
at junit.framework.TestCase.runBare(TestCase.java:134)
at junit.framework.TestResult$1.protect(TestResult.java:115)
at android.support.test.internal.runner.junit3.AndroidTestResult.runProtected(AndroidTestResult.java:77)
at junit.framework.TestResult.run(TestResult.java:118)
at android.support.test.internal.runner.junit3.AndroidTestResult.run(AndroidTestResult.java:55)
at junit.framework.TestCase.run(TestCase.java:124)
at android.support.test.internal.runner.junit3.NonLeakyTestSuite$NonLeakyTest.run(NonLeakyTestSuite.java:63)
at junit.framework.TestSuite.runTest(TestSuite.java:243)
at junit.framework.TestSuite.run(TestSuite.java:238)
at android.support.test.internal.runner.junit3.DelegatingTestSuite.run(DelegatingTestSuite.java:103)
at android.support.test.internal.runner.junit3.AndroidTestSuite.run(AndroidTestSuite.java:69)
at android.support.test.internal.runner.junit3.JUnit38ClassRunner.run(JUnit38ClassRunner.java:103)
at org.junit.runners.Suite.runChild(Suite.java:128)
[...]

PR #864, opened 2017-06-09
Removing duplicate event entries for acquireTokenSilentSync #863

addev-ashish:
API event is already finished in an earlier function, the call to stop event causes a dummy apievent getting inserted into the aggregated map


iambmelt:
Changes look good to me. Tests are passing locally (except #853 - unrelated issue) -- CI is failing on Travis, but not on VSTS. I'll approve this change and open a separate issue for troubleshooting the CI.

For due diligence, we may want to get CI passing in all environments prior to merging -- I don't intend to take a hard-line on this point, though; I'd prefer if we weren't totally blocked by CI unreliability.

heidijinxujia:
Travis failed because of the unit test
com.microsoft.aad.adal.AuthenticationContextTest > testFamilyClientIdCorrectlyStoredInCache

PR #877, opened 2017-06-27
Corrects Javadoc of AuthenticationActivity#cancelAuthenticationRequest(int)

### LINKED ISSUE #876: AuthenticationContext#cancelAuthenticationActivity(int) does not behave as documented
From the [Javadoc](https://github.com/AzureAD/azure-activedirectory-library-for-android/blob/dev/adal/src/main/java/com/microsoft/aad/adal/AuthenticationContext.java#L868-L879)
>@return true: if there is a valid waiting request and cancel message send successfully. false: Request does not exist or cancel message not send

However, looking at the implementation we see that inverse is true. [`true` is returned if no callback is awaiting cancellation.](https://github.com/AzureAD/azure-activedirectory-library-for-android/blob/dev/adal/src/main/java/com/microsoft/aad/adal/AuthenticationContext.java#L883-L887)

I recommend the documentation be updated to match the functionality as-coded.

iambmelt:
This closes #876 -- this issue was discovered in testing.


nazukj:
:shipit:

PR #880, opened 2017-06-28
add claims challenge support 

weijjia:
Claims challenge could be returned from middle tier service if it's doing downstream obo. Adal needs to update to 1) add new overloads to take in the claims challenge 2) skip cache lookup. 3) send passed in claims challenge to authorize endpoint


adal/src/main/java/com/microsoft/aad/adal/AuthenticationContext.java
```diff
@@ -426,7 +426,51 @@ public void acquireToken(Activity activity, String resource, String clientId,
...
+    public void acquireToken(final Activity activity, final String resource, final String clientId, @Nullable String redirectUri,
+                             @Nullable final String loginHint, @Nullable final PromptBehavior prompt, @Nullable String extraQueryParameters,
+                             @Nullable final String claims, final AuthenticationCallback<AuthenticationResult> callback) {
+        throwIfClaimsInBothExtraQpAndClaimsParameter(claims, extraQueryParameters);
+
+        if (checkPreRequirements(resource, clientId, callback)
+                && checkADFSValidationRequirements(loginHint, callback)) {
+            redirectUri = getRedirectUri(redirectUri);
+            final String requestId = Telemetry.registerNewRequest();
+            final APIEvent apiEvent = createApiEvent(mContext, clientId, requestId, EventStrings.ACQUIRE_TOKEN_8);
```
addev-ashish: nit: kindly update this number in the telemetry doc if not already done

adal/src/main/java/com/microsoft/aad/adal/BrokerProxy.java
```diff
@@ -653,6 +653,15 @@ private Bundle getBrokerOptions(final AuthenticationRequest request) {
         if (request.getPrompt() != null) {
             brokerOptions.putString(AuthenticationConstants.Broker.ACCOUNT_PROMPT, request.getPrompt().name());
         }
+
+        if (!StringExtensions.isNullOrBlank(request.getClaimsChallenge())) {
```
addev-ashish: These 2 checks do exactly the same thing, why not merge them into a single if block?

adal/src/main/java/com/microsoft/aad/adal/AuthenticationContext.java
weijjia: kk, will do.

PR #892, opened 2017-07-10
Create issue_template.md

nazukj:
Adding a template for issues to ensure we get relevant information when new ones are logged.


iambmelt:
This looks like a good template for us. We could consider advising consumers of this library what _not_ to post as well: PII, OII, credentials, secrets, tokens.

PR #896, opened 2017-07-11
Cleans up imports across all modules

### LINKED ISSUE #895: Clean up project imports
Discovered while troubleshooting VSTS-CI builds: there are many unused and unsorted imports in the codebase. Use Android Studio's import optimizer to tidy these up.

iambmelt:
This change cleans and sorts imports across all ADAL project modules. It closes #895


adal/src/androidTest/java/com/microsoft/aad/adal/AuthenticationActivityUnitTest.java
```diff
@@ -23,13 +23,20 @@
...
+import android.content.Context;
+import android.content.Intent;
+import android.os.AsyncTask;
+import android.os.Bundle;
+import android.test.ActivityUnitTestCase;
+import android.test.RenamingDelegatingContext;
+import android.test.UiThreadTest;
+import android.test.suitebuilder.annotation.SmallTest;
+import android.webkit.WebView;
+import android.webkit.WebViewClient;
```
heidijinxujia: Are they the same? It looks like they just swapped.

heidijinxujia:
I think I am kind of confused here, I saw a lot of imports are just re-ordered. Is it necessary?

adal/src/androidTest/java/com/microsoft/aad/adal/AuthenticationActivityUnitTest.java
iambmelt: Should be the same, but alphabetized -- the output of running Android Studio's import optimizer

heidijinxujia:
OK. Got it.

heidijinxujia:
As to Android Studio's import optimizer, is there a way we could run it automatically per build in Android Studio? Or this is an one-time task,

iambmelt:
@heidijinxujia After a little sleuthing, it turns out the answer is yes! And it doesn't even mean setting up a `git` pre-commit hook 😎 

Check out `File | Settings | Editor | General | Auto Import | Optimize imports on the fly`
imports

PR #900, opened 2017-07-13
Use versions.gradle to aggregate versions across subproject

jaredsburrows:
@nazukj @addev-ashish @weijjia :)
Similar to projects like ButterKnife, aggregate common variables in a single `.gradle` file. See example: https://github.com/JakeWharton/butterknife/blob/master/build.gradle#L15.

Once this is in, I will do the same for MSAL :).


iambmelt:
Thank you for your PR @jaredsburrows! -- I'll review this PR today and see that a secondary reviewer from the team gets eyes on it too 👍

jaredsburrows:
@iambmelt Thanks. Once this is approved and merged. I will do the same for MSAL. This helps organize and keeps track of dependencies.

gradle/versions.gradle
```diff
@@ -0,0 +1,39 @@
...
+}
+
+// Plugins
+ext.plugs = [
+    "gradle"                  : "com.android.tools.build:gradle:2.2.3",
+    "androidMavenGradlePlugin": "com.github.dcendents:android-maven-gradle-plugin:1.4.1"
+]
+
+// Dependencies
+ext.deps = [
```
weijjia: Love the version change, it helps us to have a single place for all the versions. Thanks!! 
But the dependency part, it just introduce one extra level, do we really need it?

jaredsburrows: One is for plugins only and the other is for dependencies only.

weijjia: I get that :) Question is that does the dependencies part really help?

jaredsburrows: What are you asking?

`plugs` is for Gradle plugins only - `buildscript` block
`deps` is for Gradle dependencies only - `dependencies` block

It is a clear and definite separation. Do you have a suggestion here?

jaredsburrows: Are you saying you want it all in `deps`?

weijjia: Dependencies is clearly declared in a dependency block, so I'm sorry that I don't quite understand how does your change help organizing the dependencies. Also, you are renaming the dependency, each dependency is uniquely declared by groupid + artifactid + version. To me, this part is only introducing one extra level.

jaredsburrows: @weijjia This project has multi `build.gradle` files. Aggregating all of the GAVs(groupid:artifactid:version) together, we can better keep track of the versions we are using.

weijjia:
@iambmelt If you're fine with the change, can you merge it when it passes on vsts?

jaredsburrows:
@weijjia Thanks. Like I said in the previous comment, I can combine everything into "deps" if you would like?

weijjia:
Like I said, I like the way in this PR to organize versions. But i don't quite agree on the deps part. When I look at an sdk dependency, i'll check the sdk build.gradle and checck the dependency section on build.gradle. With this change, i'll see something maybe called "runtime", what is runtime? I'll have to go to deps and figure out what is the real dependency. We're not maintaining a massive dependency list, so don't think it's needed. I'll leave this to the team though :)

iambmelt:
OK - @weijjia and I chatted on our side about what a good middle-ground here might look like and we came up with the following.

Let's keep the dependencies declared (by their full-name, minus the version) in each `<module>/build.gradle` file - let's have the versions which are common to every module reference a centralized constant so we can tune all the `minSdkVersions`, `targetSdkVersions`, `compileSdkVersions`, and `buildToolsVersions` at once. For dependencies which get used in multiple projects, let's have them share a version constant: if it gets used once, let's just keep it inline.

jaredsburrows:
@iambmelt I am now confused by your last comment. You want each dependencies in each `modules/build.gradle`.

weijjia:
@jaredsburrows Yup. We want a shared version, but dependencies are declared in its own module.

jaredsburrows:
@weijjia Can you show me an example? How can we have a shared version but then declare them elsewhere?

weijjia:
@jaredsburrows Sorry, i don't think I understand your question. Essentially, @iambmelt and I are asking you only to aggregate versions.

iambmelt:
Sorry for the confusion here.

The idea is that the dependencies are still declared by the customary name in the customary place (`dependencies`) in the `build.gradle` of each module. The part we would like to see separated out into a constant is the version **number**

So there's a declaration like:

```groovy
ext {
    supportLibraryVersion = '25.3.1'
    // more stuff like this
}
```

and then each module's `build.gradle` would reference that, without using `deps`

```groovy
dependencies {
    compile "com.android.support:support-v4:$rootProject.ext.supportLibraryVersion"
}
```

jaredsburrows:
@iambmelt @weijjia All fixed. Anything else?

iambmelt:
Rebooted CI - pushed empty commit to reinit build for merge

jaredsburrows:
@iambmelt Travis had a successful build.

iambmelt:
@jaredsburrows Yep - changes all look great. Sadly, our VSTS-CI is not presently configured to report build statuses for forks and our use of protected branches locks-out merge until Travis _and_ VSTS pass. I've copied your PR over to #909 to work around this. My apologies for the runaround :-/

PR #923, opened 2017-07-26
improve code coverage of StorageHelper from 50% to 70%

piqiums:
I did a small refactor in the StorageHelper class, and added another testcase.

### DIFF
--- a/adal/src/main/java/com/microsoft/aad/adal/StorageHelper.java
+++ b/adal/src/main/java/com/microsoft/aad/adal/StorageHelper.java
@@ -304,22 +304,30 @@ public class StorageHelper {
      */
     synchronized SecretKey loadSecretKeyForEncryption() throws IOException,
             GeneralSecurityException {
+        final byte[] secretKeyData = AuthenticationSettings.INSTANCE.getSecretKeyData();
+        return loadSecretKeyForEncryption(secretKeyData == null ? VERSION_ANDROID_KEY_STORE : VERSION_USER_DEFINED);
+    }
+
+    /**
+     * Get Secret Key based on API level to use in encryption. Decryption key
+     * depends on version# since user can migrate to new Android.OS
+     *
+     * @param defaultBlobVersion the blobVersion to use by default
+     * @return SecretKey Get Secret Key based on API level to use in encryption.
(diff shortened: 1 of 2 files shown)


iambmelt:
@piqiums @weijjia I have a PR into this branch here #924 to test if this builds under `Precise`

adal/src/main/java/com/microsoft/aad/adal/StorageHelper.java
```diff
@@ -304,22 +304,30 @@ public String decrypt(final String encryptedBlob)
...
+    /**
+     * Get Secret Key based on API level to use in encryption. Decryption key
+     * depends on version# since user can migrate to new Android.OS
+     *
+     * @param defaultBlobVersion the blobVersion to use by default
+     * @return SecretKey Get Secret Key based on API level to use in encryption.
+     * @throws GeneralSecurityException
+     * @throws IOException
+     */
+    synchronized SecretKey loadSecretKeyForEncryption(String defaultBlobVersion) throws IOException,
```
iambmelt: is the visibility of this method appropriate?

iambmelt: Which is to say: I understand it cannot be both `private` and effectively unit tested. My concern is that it shares name with another method and if our intention is not to invoke this function from another file this may be making a hole that's easy to fall into.

piqiums: I think it's appropriate since it's the same with the existing "loadSecretKeyForEncryption()" , which takes no arguments. "loadSecretKeyForEncryption()" is very difficult to test, and thus raised issue like https://github.com/AzureAD/azure-activedirectory-library-for-android/issues/580  , it's suppressed now and the code coverage is lowered. Now I am adding another method with only one more argument in the method declaration, and it's the real implementation of "loadSecretKeyForEncryption()" . So now it's testable and code coverage is better.

iambmelt:
Updated ADAL Code Coverage Report

## Overall
adal_coverage_overall

## Class-by-Class (Top 10)
adal_coverage_top_10

/cc @nazukj

PR #938, opened 2017-08-09
UT migration, deprecating ActivityUnitTestCase

piqiums:
UT migration, deprecating ActivityUnitTestCase.
Part of the UT method migration https://identitydivision.visualstudio.com/DevEx/Android/_workitems?id=264686 .

### DIFF
--- a/adal/src/androidTest/java/com/microsoft/aad/adal/AuthenticationActivityUnitTest.java
+++ b/adal/src/androidTest/java/com/microsoft/aad/adal/AuthenticationActivityUnitTest.java
@@ -90,7 +90,7 @@ public class AuthenticationActivityUnitTest {
     private Intent mIntentToStartActivity;
 
     @Rule
-    public ActivityTestRule<AuthenticationActivity> mActivityRule = new ActivityTestRule(AuthenticationActivity.class, true, false);
+    public final ActivityTestRule<AuthenticationActivity> mActivityRule = new ActivityTestRule<>(AuthenticationActivity.class, true, false);
 
     @Before
     public void setUp() throws Exception {
--- a/adal/src/androidTest/java/com/microsoft/aad/adal/AuthenticationActivityUnitTest.java
+++ b/adal/src/androidTest/java/com/microsoft/aad/adal/AuthenticationActivityUnitTest.java
@@ -31,12 +31,16 @@ import android.content.Context;
 import android.content.Intent;
 import android.os.AsyncTask;
 import android.os.Bundle;
-import android.test.ActivityUnitTestCase;
+import android.support.test.rule.ActivityTestRule;
+import android.support.test.runner.AndroidJUnit4;
 import android.test.RenamingDelegatingContext;
-import android.test.UiThreadTest;
-import android.test.suitebuilder.annotation.SmallTest;
 import android.webkit.WebView;
 import android.webkit.WebViewClient;
+
+import org.junit.Before;
+import org.junit.Rule;
+import org.junit.Test;
+import org.junit.runner.RunWith;
 
 import java.io.Serializable;
 import java.lang.reflect.Constructor;
@@ -54,6 +58,11 @@ import javax.crypto.SecretKeyFactory;
 import javax.crypto.spec.PBEKeySpec;
 import javax.crypto.spec.SecretKeySpec;
 
+import static android.support.test.InstrumentationRegistry.getInstrumentation;
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertNotNull;
+import static org.junit.Assert.assertNotSame;
+import static org.junit.Assert.assertTrue;
 import static org.mockito.Matchers.any;
 import static org.mockito.Matchers.anyString;
 import static org.mockito.Matchers.eq;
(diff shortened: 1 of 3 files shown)


adal/src/androidTest/java/com/microsoft/aad/adal/AuthenticationActivityUnitTest.java
```diff
@@ -79,20 +89,13 @@
 
     private Intent mIntentToStartActivity;
 
-    private AuthenticationActivity mActivity;
-
-    public AuthenticationActivityUnitTest() {
-        super(AuthenticationActivity.class);
-    }
+    @Rule
+    public ActivityTestRule<AuthenticationActivity> mActivityRule = new ActivityTestRule(AuthenticationActivity.class, true, false);
```
iambmelt: It looks like this field can be made `final` -- I'm also seeing a warning in my IDE for using the raw-type constructor which can be suppressed by replacing this line with:
`    public final ActivityTestRule<AuthenticationActivity> mActivityRule = new ActivityTestRule<>(AuthenticationActivity.class, true, false);`

adal/src/androidTest/java/com/microsoft/aad/adal/AuthenticationActivityUnitTest.java
```diff
@@ -173,26 +174,32 @@ public void testReturnToCaller() throws IllegalArgumentException, NoSuchFieldExc
...
         MockDeviceCertProxy.reset();
         MockDeviceCertProxy.setIsValidIssuer(true);
         MockDeviceCertProxy.setPrivateKey(null);
-        String url = AuthenticationConstants.Broker.PKEYAUTH_REDIRECT
+        final String url = AuthenticationConstants.Broker.PKEYAUTH_REDIRECT
                 + "?Nonce=nonce1234&CertAuthorities=ABC&Version=1.0&SubmitUrl=submiturl&Context=serverContext";
-        WebViewClient client = getCustomWebViewClient();
-        WebView mockview = new WebView(getActivity().getApplicationContext());
-        ReflectionUtils.setFieldValue(mActivity, "mSpinner", null);
+        final WebViewClient client = getCustomWebViewClient();
```
iambmelt: There appears to be some copy/paste looking code here - can we reduce this in some way?

adal/src/androidTest/java/com/microsoft/aad/adal/AuthenticationActivityUnitTest.java
```diff
@@ -216,41 +223,54 @@ public void testWebviewAuthenticationException() throws IllegalArgumentException
...
          * case 1: url = "http://login.microsoftonline.com/"
          * case 2: url = "https://login.microsoftonline.com/"
          */
-        String url = "https://login.microsoftonline.com/";
-        WebViewClient client = getCustomWebViewClient();
-        WebView mockview = new WebView(getActivity().getApplicationContext());
-        ReflectionUtils.setFieldValue(mActivity, "mSpinner", null);
-        assertEquals(false, client.shouldOverrideUrlLoading(mockview, url));
+        final String url = "https://login.microsoftonline.com/";
+        final WebViewClient client = getCustomWebViewClient();
```
iambmelt: There appears to be some copy/paste looking code here - can we reduce this in some way?

adal/src/androidTest/java/com/microsoft/aad/adal/AuthenticationActivityUnitTest.java
```diff
@@ -216,41 +223,54 @@ public void testWebviewAuthenticationException() throws IllegalArgumentException
...
         /*
          * case 1: url = "about:blank"
          */
-        String url = "about:blank";
-        WebViewClient client = getCustomWebViewClient();
-        WebView mockview = new WebView(getActivity().getApplicationContext());
-        ReflectionUtils.setFieldValue(mActivity, "mSpinner", null);
-        assertEquals(true, client.shouldOverrideUrlLoading(mockview, url));
+        final String url = "about:blank";
+        final WebViewClient client = getCustomWebViewClient();
```
iambmelt: There appears to be some copy/paste looking code here - can we reduce this in some way?

adal/src/androidTest/java/com/microsoft/aad/adal/AuthenticationActivityUnitTest.java
piqiums: Since there're 2 intents starting the Activity among the test case class, I could not launch Activity in the 'setUp' (with @Before) method. What I could do is only launching Activity in separate test methods. Previously, there's the "@UiThreadTest" annotation on the method. But I have to remove it since it would crash on launching Activity. Then, there're some methods must be called on UI thread, then I have to use 'mActivityRule.runOnUiThread'. Yes, there's some copy/paste on code that must run on UI thread. thanks

adal/src/androidTest/java/com/microsoft/aad/adal/AuthenticationActivityUnitTest.java
piqiums: done

adal/src/androidTest/java/com/microsoft/aad/adal/AuthenticationActivityUnitTest.java
iambmelt: k I can live with it 👍

Issue #901, opened 2017-07-14
Outlook: ADAL should return different error message when communcation with broker fails


addev-ashish:
When broker auth is involved, if the broker returns null or an error happens communicating with broker ADAL falls back to returning AUTH_REFRESH_FAILED_PROMPT_NOT_ALLOWED. 
This makes the apps think that auth failed, and they take the users to the UI breaking the SSO experience.

piqiums:
looking into this.
Ashish, do you know the steps to repro it? thanks

heidijinxujia:
#973

nazukj:
@heidijinxujia if this is fixed and merged, can you please close this issue?

PR #945, opened 2017-08-16
return a different error message other than 'AUTH_REFRESH_FAILED_PROMPT_NOT_ALLOWED' when communcation with broker fails.

piqiums:
return a different error message other than 'AUTH_REFRESH_FAILED_PROMPT_NOT_ALLOWED' when communcation with broker fails.

fix https://github.com/AzureAD/azure-activedirectory-library-for-android/issues/901#issue-243075944

### DIFF
--- a/adal/src/main/java/com/microsoft/aad/adal/BrokerProxy.java
+++ b/adal/src/main/java/com/microsoft/aad/adal/BrokerProxy.java
@@ -315,15 +315,12 @@ class BrokerProxy implements IBrokerProxy {
         final Bundle bundleResult;
         if (isBrokerAccountServiceSupported()) {
             bundleResult = BrokerAccountServiceHandler.getInstance().getAuthToken(mContext, requestBundle, brokerEvent);
-            if (bundleResult == null) {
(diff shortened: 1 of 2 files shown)


adal/src/main/java/com/microsoft/aad/adal/BrokerProxy.java
```diff
@@ -315,12 +315,15 @@ public AuthenticationResult getAuthTokenInBackground(final AuthenticationRequest
         final Bundle bundleResult;
         if (isBrokerAccountServiceSupported()) {
             bundleResult = BrokerAccountServiceHandler.getInstance().getAuthToken(mContext, requestBundle, brokerEvent);
+            if (bundleResult == null) {
```
weijjia: i'm wondering that there will be cases when bundle result is null, it's actually indicating that interactive flow is needed.

piqiums: I tried the 2 cases:
1) no token existed, the 'bundleResult' is not NULL.
2) error returned when you redeem refresh token. the 'bundleResult' is NULL.

addev-ashish:
This did not fix the issue completely, lets start over.

Issue #913, opened 2017-07-24
login.windows.net authority migration to login.microsoftonline.com


weijjia:
<ul>
  <li>There currently are a number of DNS aliases and redirects used for EvoSTS, all of which currently redirect to login.microsoftonline.com, we're looking to migrate all of these islands into a single source of tokens to eliminate this source of extra authentication prompts. </li>
  <li>The same tokens are issued for all the aliased authority URLs, however client SDKs currently don't have visibility into this aliasing behavior and use the original authority URL provided by the client application for caching tokens, client applications can receive these authorities from a multitude of sources creating islands of tokens when different sources provide different authorities, even though they are equivalent tokens. This results in multiple auth prompts, possibly within the same app, and "islands of SSO"
 effects when multiple apps are involved.</li>
</ul>

iambmelt:
@weijjia Now that #956 #977 and #974 are merged, are we clear to close this issue?

weijjia:
PRs are merged, closed the issue.

PR #956, opened 2017-09-01
Authority migration: authority validation threading update and extend validation cache to store metadata

weijjia:

<ul>
  <li>https://github.com/AzureAD/azure-activedirectory-library-for-android/issues/913</li>
  <li>1) extend authority validation cache to store metada 2) update threading handling for authority validation </li>
</ul>

### DIFF
--- a/adal/src/androidTest/java/com/microsoft/aad/adal/AuthorityValidationMetadataCacheTest.java
+++ b/adal/src/androidTest/java/com/microsoft/aad/adal/AuthorityValidationMetadataCacheTest.java
@@ -1,0 +1,120 @@
+//  Copyright (c) Microsoft Corporation.
+//  All rights reserved.
+//
+//  This code is licensed under the MIT License.
+//
+//  Permission is hereby granted, free of charge, to any person obtaining a copy
+//  of this software and associated documentation files(the "Software"), to deal
+//  in the Software without restriction, including without limitation the rights
+//  to use, copy, modify, merge, publish, distribute, sublicense, and / or sell
+//  copies of the Software, and to permit persons to whom the Software is
+//  furnished to do so, subject to the following conditions :
+//
+//  The above copyright notice and this permission notice shall be included in
+//  all copies or substantial portions of the Software.
+//
+//  THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+//  IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+//  FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+//  AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+//  LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+//  OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+//  THE SOFTWARE.
+
+package com.microsoft.aad.adal;
+
+import android.support.test.runner.AndroidJUnit4;
+
+import junit.framework.Assert;
+
+import org.json.JSONException;
+import org.junit.After;
+import org.junit.Before;
+import org.junit.Test;
+import org.junit.runner.RunWith;
+
+import java.net.MalformedURLException;
+import java.net.URL;
+import java.util.HashMap;
+import java.util.Map;
+
+/**
+ * Tests for {@link AuthorityValidationMetadataCache}.
+ */
+@RunWith(AndroidJUnit4.class)
+public class AuthorityValidationMetadataCacheTest extends AndroidTestHelper {
+
+    @Before
+    public void setUp() throws Exception {
+        super.setUp();
+    }
+
+    @After
+    public void tearDown() throws Exception {
+        super.tearDown();
+        AuthorityValidationMetadataCache.clearAuthorityValidationCache();
+    }
+
+    @Test
+    public void testProcessInstanceDiscoveryMetadata() throws MalformedURLException, JSONException {
+        final URL authorityURL = new URL("https://login.windows.net/common");
+        AuthorityValidationMetadataCache.processInstanceDiscoveryMetadata(authorityURL, getDiscoveryResponse());
+
+        Map<String, InstanceDiscoveryMetadata> authorityValidationMap = AuthorityValidationMetadataCache.getAuthorityValidationMetadataCache();
+        Assert.assertTrue(authorityValidationMap.size() == 5);
+        Assert.assertTrue(authorityValidationMap.containsKey("login.windows.net"));
+        Assert.assertTrue(authorityValidationMap.containsKey("login.microsoftonline.com"));
+        Assert.assertTrue(authorityValidationMap.containsKey("sts.microsoft.com"));
+        Assert.assertTrue(authorityValidationMap.containsKey("login.microsoft.com"));
+        Assert.assertTrue(authorityValidationMap.containsKey("login.microsoftonline.de"));
+    }
+
+    @Test
+    public void testMetadataNotReturned() throws MalformedURLException, JSONException {
+        final URL authorityUrl = new URL("https://login.windows.net/common");
+        AuthorityValidationMetadataCache.processInstanceDiscoveryMetadata(authorityUrl, getDiscoveryResponseWithNoMetadata());
+
+        Map<String, InstanceDiscoveryMetadata> authorityValidationMap = AuthorityValidationMetadataCache.getAuthorityValidationMetadataCache();
+        Assert.assertTrue(authorityValidationMap.size() == 1);
+        Assert.assertTrue(authorityValidationMap.containsKey("login.windows.net"));
+    }
+
+    private Map<String, String> getDiscoveryResponseWithNoMetadata() {
+        final Map<String, String> discoveryResponse = new HashMap<>();
+        discoveryResponse.put(AuthorityValidationMetadataCache.TENANT_DISCOVERY_ENDPOINT, "valid_endpoint");
+
+        return discoveryResponse;
+    }
+
+    static Map<String, String> getDiscoveryResponse() {
+        final Map<String, String> metadata = new HashMap<>();
+        metadata.put(AuthorityValidationMetadataCache.TENANT_DISCOVERY_ENDPOINT, "valid_endpoint");
+        metadata.put(AuthorityValidationMetadataCache.META_DATA, getMetadata());
+
+        return metadata;
+    }
+
+    static String getMetadata() {
+        final String metadata = "[" +
+                "{" +
+                "\"preferred_network\": \"login.microsoftonline.com\"," +
+                "\"preferred_cache\": \"login.windows.net\"," +
+                "\"aliases\": [" +
+                "\"login.microsoftonline.com\"," +
+                "\"login.windows.net\"," +
+                "\"login.microsoft.com\"," +
+                "\"sts.microsoft.com\"" +
+                "]" +
+                "}," +
+                "{" +
+                "\"preferred_network\": \"login.microsoftonline.de\"," +
+                "\"preferred_cache\": \"login.microsoftonline.de\"," +
+                "\"aliases\": [" +
+                "\"login.microsoftonline.de\"" +
+                "]" +
+                "}" +
+                "]";
+
+        return metadata;
+    }
+}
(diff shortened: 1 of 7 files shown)


adal/src/androidTest/java/com/microsoft/aad/adal/AuthorityValidationMetadataCacheTest.java
```diff
@@ -0,0 +1,120 @@
...
+import java.util.Map;
+
+/**
+ * Tests for {@link AuthorityValidationMetadataCache}.
+ */
+@RunWith(AndroidJUnit4.class)
+public class AuthorityValidationMetadataCacheTest extends AndroidTestHelper {
+
+    @Before
+    public void setUp() throws Exception {
```
iambmelt: Is this `@Override` purely for future convenience? Is it needed?

adal/src/main/java/com/microsoft/aad/adal/AuthorityValidationMetadataCache.java
```diff
@@ -0,0 +1,125 @@
...
+ */
+
+final class AuthorityValidationMetadataCache {
+    private static final String TAG = AuthorityValidationMetadataCache.class.getSimpleName();
+
+    static final String TENANT_DISCOVERY_ENDPOINT = "tenant_discovery_endpoint";
+
+    static final String META_DATA = "metadata";
+
+    static final String PREFERRED_NETWORK = "preferred_network";
```
iambmelt: minor-nit: visibility for this `PREFERRED_NETWORK`, `PREFERRED_CACHE`, and `ALIASES` could `private`

adal/src/main/java/com/microsoft/aad/adal/Discovery.java
```diff
@@ -54,21 +55,21 @@
...
+    private static final String API_VERSION_VALUE = "1.1";
 
     private static final String AUTHORIZATION_ENDPOINT_KEY = "authorization_endpoint";
 
     private static final String INSTANCE_DISCOVERY_SUFFIX = "common/discovery/instance";
 
     private static final String AUTHORIZATION_COMMON_ENDPOINT = "/common/oauth2/authorize";
 
-    private static final String TENANT_DISCOVERY_ENDPOINT = "tenant_discovery_endpoint";
+    private static ReentrantLock sLock = null;
```
iambmelt: Does this variable need to be declared `volatile`? It looks as though it is intended to be read by multiple threads, possibly concurrently

adal/src/main/java/com/microsoft/aad/adal/Discovery.java
```diff
@@ -321,7 +327,22 @@ private URL buildQueryString(final String instance, final String authorizationEn
         return new URL(builder.build().toString());
     }
 
+    /**
+     * @return {@link ReentrantLock} for locking the network request queue.
+     */
+    private ReentrantLock getLock() {
```
iambmelt: This check feels a little weird for a few reasons. It behaves like a factory (in that it returns new instances) but `sLock` is not assigned here.

It also initializes a `static` field from an instance method and that feels strange.

I suspect that if the lock is not being acquired in one method and released in another, and there aren't any `waits` involved a more primitive synchronization technique may be sufficient (such as a synchronzied method or block, without the use of a `ReentrantLock`) -- I'll leave this up to your judgement but this looks dodgy

adal/src/main/java/com/microsoft/aad/adal/InstanceDiscoveryMetadata.java
```diff
@@ -0,0 +1,70 @@
...
+//  LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+//  OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+//  THE SOFTWARE.
+
+package com.microsoft.aad.adal;
+
+import java.util.ArrayList;
+import java.util.List;
+
+class InstanceDiscoveryMetadata {
```
iambmelt: Could/should be `final`?

adal/src/androidTest/java/com/microsoft/aad/adal/AuthorityValidationMetadataCacheTest.java
weijjia: you mean @Override or @Before?

iambmelt: I mean `@Override` (the annotation is missing) - but is it necessary if it just calls `super`?

adal/src/main/java/com/microsoft/aad/adal/AuthorityValidationMetadataCache.java
weijjia: done.

adal/src/main/java/com/microsoft/aad/adal/Discovery.java
weijjia: it's a good point. updated.

adal/src/main/java/com/microsoft/aad/adal/Discovery.java
weijjia: synchronization method is tied to an object. If there are multiple discovery objects are created, it's not doing any synchronization actually.

adal/src/main/java/com/microsoft/aad/adal/InstanceDiscoveryMetadata.java
weijjia: yup, done.

adal/src/main/java/com/microsoft/aad/adal/Discovery.java
```diff
@@ -101,18 +102,29 @@ void validateAuthorityADFS(final URL authorizationEndpoint, final String domain)
...
+        if (AAD_WHITELIST_HOSTS.contains(authorizationEndpoint.getHost().toLowerCase(Locale.US))) {
+            trustedHost = authorityHost;
+        } else {
+            trustedHost = TRUSTED_QUERY_INSTANCE;
+        }
+
+        try {
+            sLock = getLock();
+            sLock.lock();
+            performInstanceDiscovery(authorizationEndpoint, trustedHost);
```
addev-ashish: this is looking dangerous. You took a lock and then made a network request, say the request blocks up for 10 minutes (like we have seen previously). It will block every other validateAuthority request. What is the downside of making multiple discovery requests?

adal/src/main/java/com/microsoft/aad/adal/Discovery.java
```diff
@@ -241,13 +260,14 @@ private boolean sendRequest(final URL queryUrl) throws IOException, JSONExceptio
...
 
-            return discoveryResponse.containsKey(TENANT_DISCOVERY_ENDPOINT);
+            return discoveryResponse;
         } finally {
             ClientMetrics.INSTANCE.endClientMetricsRecord(
                     ClientMetricsEndpointType.INSTANCE_DISCOVERY, mCorrelationId);
         }
     }
 
+
```
addev-ashish: nit: extra line

adal/src/main/java/com/microsoft/aad/adal/Discovery.java
```diff
@@ -321,7 +327,22 @@ private URL buildQueryString(final String instance, final String authorizationEn
...
+                }
+            }
+        }
+
+        return sLock;
+    }
+
     Set<String> getValidHosts() {
-        return VALID_HOSTS;
+        return AAD_WHITELIST_HOSTS;
```
addev-ashish: nit: should this be AAD_WHITELISTED_HOSTS? just curious

adal/src/main/java/com/microsoft/aad/adal/AuthorityValidationMetadataCache.java
```diff
@@ -0,0 +1,125 @@
...
+    }
+
+    private static InstanceDiscoveryMetadata processSingleJsonArray(final JSONObject jsonObject) throws JSONException {
+        final String preferredNetwork = jsonObject.getString(PREFERRED_NETWORK);
+        final String preferredCache = jsonObject.getString(PREFERRED_CACHE);
+        final JSONArray aliasArray = jsonObject.getJSONArray(ALIASES);
+
+        final List<String> aliases = new ArrayList<>();
+        for (int i = 0; i < aliasArray.length(); i++) {
+            aliases.add((String) aliasArray.get(i));
```
addev-ashish: this can be changed to aliasArray.getString

adal/src/main/java/com/microsoft/aad/adal/Discovery.java
weijjia: Thanks for bringing this up. I was actually planning to talk to Ryan about adding timeout.

adal/src/main/java/com/microsoft/aad/adal/Discovery.java
weijjia: done.

adal/src/main/java/com/microsoft/aad/adal/Discovery.java
weijjia: updated

adal/src/androidTest/java/com/microsoft/aad/adal/AuthorityValidationMetadataCacheTest.java
weijjia: nah, i think it's fine.

adal/src/main/java/com/microsoft/aad/adal/Discovery.java
weijjia: One network request at a time for authority validation is by design. The issue we have seen before is about write timeout, for which we cannot set a default value. Instance discovery is a Get request, we should be fine for this case.

RPangrle: Conceptually, the lock should be protecting solely the data structure, and not the entire authority validation request. A serialized task queue should be used to ensure that only a singular authority validation request is made at any given moment.

What are the side effects of calling a blocking operation on the main thread on Android? The documentation for this method says:

```
If the lock is held by another thread then the current thread becomes disabled for thread scheduling purposes and lies dormant until the lock has been acquired, at which time the lock hold count is set to one.
```

Which my interpretation of would mean if you called it on the main UI thread it would sleep the thread until the lock was released... which would probably not be what you want.

addev-ashish: @RPangrle calling network on UI thread will crash it. My concern was if the lock was taken and we hit the network and that call gets delayed or stuck, all of that apps threads doing auth will get stuck behind this one lock. Which will not be a deadlock, but still a slowdown.
@weijjia I understand that you mentioned that GET has a timeout, but Ryan mentions here that the intention of the lock was just to protect the data structure. Can we change the code to reflect that?

weijjia: @addev-ashish so first there is no network request made on main thread.  Second, what would be delayed? read timeout and connection timeout is 30 seconds. Also, data structure is already protected by concurrent map, how we are going to make sure one request goes out at a time? The lock is solely for making sure network call for only instance discovery request. You have to provide me concrete examples showing your concern. Only worry won't help here. If it helps for us to go back to the design and figure out what the design is, I'm happy to do that.

adal/src/main/java/com/microsoft/aad/adal/AcquireTokenRequest.java
```diff
@@ -182,26 +208,43 @@ private void validateAcquireTokenRequest(final AuthenticationRequest authenticat
...
+    private void updatePreferredNetworkLocation(final URL authorityUrl, final AuthenticationRequest request, final InstanceDiscoveryMetadata metadata)
+            throws AuthenticationException {
+        if (metadata == null || !metadata.isValidated()) {
+            return;
+        }
+
+        if (authorityUrl.getHost().equalsIgnoreCase(metadata.getPreferredNetwork())) {
+            try {
+                final URL replacedAuthority = new URL(new Uri.Builder().authority(metadata.getPreferredNetwork()).appendPath(authorityUrl.getPath()).build().toString());
+                request.setAuthority(replacedAuthority.toString());
```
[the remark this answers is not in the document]
weijjia: I should have removed this part. updated.

adal/src/main/java/com/microsoft/aad/adal/Discovery.java
```diff
@@ -54,21 +55,21 @@
...
+    private static final String API_VERSION_VALUE = "1.1";
 
     private static final String AUTHORIZATION_ENDPOINT_KEY = "authorization_endpoint";
 
     private static final String INSTANCE_DISCOVERY_SUFFIX = "common/discovery/instance";
 
     private static final String AUTHORIZATION_COMMON_ENDPOINT = "/common/oauth2/authorize";
 
-    private static final String TENANT_DISCOVERY_ENDPOINT = "tenant_discovery_endpoint";
+    private volatile static ReentrantLock sLock;
```
RPangrle: Can you add a comment explaining exactly what this lock is for, or better yet a more descriptive variable name?

weijjia: Sure. This one is already merged, I'll update in another PR.

Issue #899, opened 2017-07-11
OnePlus: If bindservice returns false or the service connection call back is not called, the callback for AT or ATS is not called


addev-ashish:
Please ensure the following information is provided when logging a new issue:

<ul>
  <li>Expected behavior: Everytime AT or ATS is called, the success or error callback should be called</li>
  <li>Actual behavior: if https://github.com/AzureAD/azure-activedirectory-library-for-android/blob/35c3eb19c257b82ccd29fb256dc7d1f8e337be67/adal/src/main/java/com/microsoft/aad/adal/BrokerAccountServiceHandler.java#L318 returns false, then the callback is not called</li>
  <li>Steps to reproduce the problem: Its repro'ing in OnePlus3 with Android 7.1.1</li>
  <li>Specifications like the version of the library, operating system etc: ADAL 1.12.0, OnePlus3 7.1.1</li>
  <li>Relevant logs and traces: Waiting on logs from customer</li>
</ul>

wuqi713:
I have this question also.
1. company portal  5.0.3743.0
2. ADAL 1.12.0
3. Android 6.0

nazukj:
Blocked on device logs.

trevorwang:
```
08-11 15:46:25.904 30920-30920/com.example.abc I/hwaps: JNI_OnLoad
08-11 15:46:25.984 30920-30920/com.example.abc V/AudioManager: playSoundEffect   effectType: 0
08-11 15:46:25.984 30920-30920/com.example.abc V/AudioManager: querySoundEffectsEnabled...
08-11 15:46:26.016 30920-30920/com.example.abc V/BrokerProxy: 2017-08-11 07:46:26-null-Misused of the checking function. This is function is only used to checking the broker permissions of API 22 or less. ver:1.13.0
08-11 15:46:26.026 30920-30920/com.example.abc V/AcquireTokenRequest: 2017-08-11 07:46:26-987c7e3f-8344-4e16-8abd-eae77ff00ddf-Sending async task from thread:30920 ver:1.13.0
08-11 15:46:26.029 30920-31016/com.example.abc V/AcquireTokenRequest: 2017-08-11 07:46:26-987c7e3f-8344-4e16-8abd-eae77ff00ddf-Running task in thread:31016 ver:1.13.0
[...]

PR #972, opened 2017-09-20
fix issue #899, OnePlus bindService issue.

piqiums:
resolve https://github.com/AzureAD/azure-activedirectory-library-for-android/issues/899

Previously, we didn't check the return value of 'bindService'. It seems 'bindService' returned false quite often on OnePlus and some other devices, then we need to call 'unbindService' in this case to avoid any leak of ServiceConnection.

### DIFF
--- a/adal/src/main/java/com/microsoft/aad/adal/BrokerAccountServiceHandler.java
+++ b/adal/src/main/java/com/microsoft/aad/adal/BrokerAccountServiceHandler.java
@@ -54,6 +54,8 @@ final class BrokerAccountServiceHandler {
     private ConcurrentMap<BrokerAccountServiceConnection, CallbackExecutor<BrokerAccountServiceConnection>> mPendingConnections = new ConcurrentHashMap<>();
     private static ExecutorService sThreadExecutor = Executors.newCachedThreadPool();
 
+    private static final int BIND_SERVICE_TIMEOUT_THRESHOLD = 30000;
+
     private static final class InstanceHolder {
         static final BrokerAccountServiceHandler INSTANCE = new BrokerAccountServiceHandler();
     }
@@ -327,6 +329,18 @@ final class BrokerAccountServiceHandler {
         if (!serviceBound) {
             context.unbindService(connection);
             callback.onError(new AuthenticationException(ADALError.BROKER_BIND_SERVICE_FAILED));
+        } else {
+            // make sure it's bound to the service with success in a period not longer than the TIMEOUT
+            final Handler handler = new Handler(Looper.getMainLooper());
+            handler.postDelayed(new Runnable(){
+                @Override
+                public void run() {
+                    if (!connection.isBound()) {
+                        context.unbindService(connection);
+                        callback.onError(new AuthenticationException(ADALError.BROKER_BIND_SERVICE_FAILED));
+                    }
+                }
+            }, BIND_SERVICE_TIMEOUT_THRESHOLD);
         }
     }
 
@@ -388,6 +402,10 @@ final class BrokerAccountServiceHandler {
 
         public void setTelemetryEvent(final BrokerEvent event) {
             mEvent = event;
+        }
+
+        public boolean isBound() {
+            return mBound;
         }
     }
 }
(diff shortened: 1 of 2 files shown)


adal/src/main/java/com/microsoft/aad/adal/BrokerAccountServiceHandler.java
```diff
@@ -313,6 +316,9 @@ private void bindToBrokerAccountService(final Context context, final Callback<Br
         final Intent brokerAccountServiceToBind = getIntentForBrokerAccountService(context);
 
         final BrokerAccountServiceConnection connection = new BrokerAccountServiceConnection();
+        BindServiceTimeoutRunnable timeoutRunnable = new BindServiceTimeoutRunnable(context, connection, callback);
```
iambmelt: Could be declared `final`

adal/src/main/java/com/microsoft/aad/adal/BrokerAccountServiceHandler.java
```diff
@@ -324,13 +330,41 @@ private void bindToBrokerAccountService(final Context context, final Callback<Br
...
+
+        public BindServiceTimeoutRunnable(Context context, BrokerAccountServiceConnection connection, final Callback<BrokerAccountServiceConnection> callback) {
+            mContext = context;
+            mServiceConnection = connection;
+            mCallback = callback;
+        }
+
+        @Override
+        public void run() {
+            if (!mServiceConnection.isBound()) {
```
iambmelt: It's unclear to me what's going on here. This appears to check that the service *isn't* bound before calling `unbindService(Context)`.

piqiums: the context is : there's some cases that the 'bindService' take too long, the callbacks of bindService ('onServiceConnected' ) were not called. So we added a timeout here. After 'bindService' was called, we set a timer of 30 seconds for now, if the service is bound with success in the 30 seconds, then cancel the timer in 'onServiceConnected'. If 'onServiceConnected' was not called in 30 seconds, then we unbind it in case it will block other requests since it's a single thread executor.

adal/src/main/java/com/microsoft/aad/adal/BrokerAccountServiceHandler.java
piqiums: will do

adal/src/main/java/com/microsoft/aad/adal/BrokerAccountServiceHandler.java
```diff
@@ -324,13 +330,41 @@ private void bindToBrokerAccountService(final Context context, final Callback<Br
         if (brokerEvent != null) {
             brokerEvent.setBrokerAccountServiceBindingSucceed(serviceBound);
         }
+        if (!serviceBound) {
+            connection.unBindService(context);
+            callback.onError(new AuthenticationException(ADALError.BROKER_BIND_SERVICE_FAILED));
+        } else {
+            // make sure it's bound to the service with success in a period not longer than the TIMEOUT
+            mUiHandler.postDelayed(timeoutRunnable, BIND_SERVICE_TIMEOUT_THRESHOLD);
```
heidijinxujia: What kind of exception will be thrown if the service bounding reaches the time threshold? Will it return a different errorCode compared to ADALError.BROKER_BIND_SERVICE_FAILED?

piqiums: it will throw an AuthenticationException of "ADALError.BROKER_BIND_SERVICE_FAILED". "ADALError.BROKER_BIND_SERVICE_FAILED" is a newly added error. please check line 357. thanks

heidijinxujia: Thanks for the reply. Is it possible to add the cause details in to the AuthenticationException(ADALError code, String details)? Currently, there is two cases here, Line 335 and Line 357. And it is difficult to tell the difference by ADALError only. Please correct me if there is anything I misunderstood. Thanks.

piqiums: good point, I've made the change. thanks

weijjia:
I think we should add unit tests to simulate the timeout case and make sure correct error code  is sent back. I understand that a lot of the code you're adding are all private methods, but we should still have coverage for those changes.

addev-ashish:
We made a more targeted fix, and we can revisit if required

PR #974, opened 2017-09-22
make sure authority validation cache's key is not case sensitive

weijjia:

<ul>
  <li>#913</li>
  <li>Make sure authority metadata cache key is all lower case</li>
</ul>

### DIFF
--- a/adal/src/main/java/com/microsoft/aad/adal/AuthorityValidationMetadataCache.java
+++ b/adal/src/main/java/com/microsoft/aad/adal/AuthorityValidationMetadataCache.java
@@ -88,7 +88,7 @@ final class AuthorityValidationMetadataCache {
     }
 
     static void updateInstanceDiscoveryMap(final String host, final InstanceDiscoveryMetadata metadata) {
-        sAadAuthorityHostMetadata.put(host, metadata);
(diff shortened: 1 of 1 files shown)


adal/src/main/java/com/microsoft/aad/adal/AuthorityValidationMetadataCache.java
```diff
@@ -88,7 +88,7 @@ static void processInstanceDiscoveryMetadata(final URL authorityUrl, final Map<S
     }
 
     static void updateInstanceDiscoveryMap(final String host, final InstanceDiscoveryMetadata metadata) {
-        sAadAuthorityHostMetadata.put(host, metadata);
+        sAadAuthorityHostMetadata.put(host.toLowerCase(Locale.US), metadata);
```
rayluo: Interesting. I was originally doing the same thing in dotnet, but then a couple [pre-existing test cases there](https://github.com/AzureAD/azure-activedirectory-library-for-dotnet/blob/Release-3.17.0/tests/Test.ADAL.Common/TokenCacheTests.cs#L177) contains mixed case already, and similar normalization caused them to break. Currently I (am forced to) preserve the case in my PR. But perhaps I should not, so that we will have functional parity across our libraries. :-)

PR #976, opened 2017-09-25
fix the case that 'bindService' returned false for issue #899

piqiums:
This is a part of fix to issue https://github.com/AzureAD/azure-activedirectory-library-for-android/issues/899 .
 It will fix the case that "bindService" return false. In this case, this change will unbind the service according to Android doc to avoid memory leak on ServiceConnection.

### DIFF
--- a/adal/src/main/java/com/microsoft/aad/adal/BrokerAccountServiceHandler.java
+++ b/adal/src/main/java/com/microsoft/aad/adal/BrokerAccountServiceHandler.java
@@ -327,6 +327,7 @@ final class BrokerAccountServiceHandler {
         if (!serviceBound) {
             connection.unBindService(context);
             callback.onError(new AuthenticationException(ADALError.BROKER_BIND_SERVICE_FAILED));
+            Logger.e(TAG, "Failed to bind service to broker app", "'bindService' returned false", ADALError.BROKER_BIND_SERVICE_FAILED);
         }
     }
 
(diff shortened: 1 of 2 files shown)


heidijinxujia:
Hi Ping, will this change be duplicate partially with #972 ?

adal/src/main/java/com/microsoft/aad/adal/BrokerAccountServiceHandler.java
```diff
@@ -324,6 +324,10 @@ private void bindToBrokerAccountService(final Context context, final Callback<Br
         if (brokerEvent != null) {
             brokerEvent.setBrokerAccountServiceBindingSucceed(serviceBound);
         }
+        if (!serviceBound) {
+            connection.unBindService(context);
+            callback.onError(new AuthenticationException(ADALError.BROKER_BIND_SERVICE_FAILED));
```
addev-ashish: Can we add an error log here.

piqiums: done. please check the updated PR. thanks

heidijinxujia:
Thank you for the reply. In addition, is it possible to add the unit test for the case "bindService' returned false"?

piqiums:
@heidijinxujia it's a simplified solution of the PR 972. I am going to check in the code for the case that 'bindService' returns false only. Ashish and I decided to move the timeout implementation to next release. When it comes to unit test, it's always better to have the change unit tested. But this change is in a private method, and furthermore, there's no way to check the status after the 'unbindService' is called on ServiceConnection. So I couldn't add the unit test in this case.  If you want to know more about the reason why this change is made in this way, please check https://developer.android.com/guide/components/bound-services.html#Binding , the NOTE in item 2. thanks

PR #977, opened 2017-09-26
foci serialize API update for authority migration update

weijjia:

<ul>
  <li>#913</li>
  <li>update serialize api, use tokenCacheAccessor which contains the cache fallback logic</li>
</ul>

### DIFF
--- a/adal/src/androidTest/java/com/microsoft/aad/adal/Util.java
+++ b/adal/src/androidTest/java/com/microsoft/aad/adal/Util.java
@@ -25,6 +25,8 @@ package com.microsoft.aad.adal;
 import android.util.Base64;
 import android.util.Log;
 
+import org.json.JSONException;
+import org.json.JSONObject;
 import org.mockito.Mockito;
 
 import java.io.ByteArrayInputStream;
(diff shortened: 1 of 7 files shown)


adal/src/androidTest/java/com/microsoft/aad/adal/Util.java
```diff
@@ -91,7 +91,19 @@ static TokenCacheItem getTokenCacheItem(final String authority, final String res
         
         return tokenCacheItem;
     }
-    
+
+    static String getSuccessTokenResponseWithFamilyClientId() {
```
iambmelt: I would consider cleaning this up using `JSONObject` -- since this JSON is currently pretty simple and this is only test code I wouldn't let this stop ship. That said, I'd much prefer to see something like the following

```
JSONObject json = new JSONObject();
json.putString("id_token", TEST_IDTOKEN);
json.putString("access_token", "blah blah");
...
return json.toString();
```

weijjia: can do it.

Issue #982, opened 2017-10-04
[Telemetry] AUTHORITY_TYPE_ADFS is missing EVENT_PREFIX


iambmelt:
`EventStrings` has most fields prepended with an `EVENT_PREFIX` -- as pointed out by @GTValentine `AUTHORITY_TYPE_ADFS` is missing this prefix.

Investigate: Was this intentional? Did iOS do this too? If not, we should fix and modify Kusto query if needed.

See [`EventStrings`](https://github.com/AzureAD/azure-activedirectory-library-for-android/blob/dev/adal/src/telemetry/java/com/microsoft/aad/adal/EventStrings.java#L154)

iambmelt:
Investigation of iOS implementation:

Authority type key: [iOS ](https://github.com/AzureAD/azure-activedirectory-library-for-objc/blob/dev/ADAL/src/telemetry/ADTelemetryEventStrings.m#L42)- [Android](https://github.com/AzureAD/azure-activedirectory-library-for-android/blob/dev/adal/src/telemetry/java/com/microsoft/aad/adal/EventStrings.java#L61)

Authority type value ADFS: [iOS ](https://github.com/AzureAD/azure-activedirectory-library-for-objc/blob/dev/ADAL/src/telemetry/ADTelemetryEventStrings.m#L118)- [Android](https://github.com/AzureAD/azure-activedirectory-library-for-android/blob/dev/adal/src/telemetry/java/com/microsoft/aad/adal/EventStrings.java#L154)

Authority type value AAD: [iOS ](https://github.com/AzureAD/azure-activedirectory-library-for-objc/blob/dev/ADAL/src/telemetry/ADTelemetryEventStrings.m#L117)- [Android](https://github.com/AzureAD/azure-activedirectory-library-for-android/blob/dev/adal/src/telemetry/java/com/microsoft/aad/adal/EventStrings.java#L155)

PR #983, opened 2017-10-05
Corrects telemetry field for Authority type AAD

iambmelt:
Corrects mismatch with iOS telemetry implementation. See [here](https://github.com/AzureAD/azure-activedirectory-library-for-objc/blob/dev/ADAL/src/telemetry/ADTelemetryEventStrings.m#L117) & #982

### DIFF
--- a/adal/src/telemetry/java/com/microsoft/aad/adal/EventStrings.java
+++ b/adal/src/telemetry/java/com/microsoft/aad/adal/EventStrings.java
@@ -152,7 +152,7 @@ final class EventStrings {
 
     // Parameter values
     static final String AUTHORITY_TYPE_ADFS = "adfs";
-    static final String AUTHORITY_TYPE_AAD = EVENT_PREFIX + "aad";
+    static final String AUTHORITY_TYPE_AAD = "aad";
 
     static final String AUTHORITY_VALIDATION_SUCCESS = EVENT_PREFIX + "authority_validation_status_success";
     static final String AUTHORITY_VALIDATION_FAILURE = EVENT_PREFIX + "authority_validation_status_failure";
--- a/adal/src/telemetry/java/com/microsoft/aad/adal/EventStrings.java
+++ b/adal/src/telemetry/java/com/microsoft/aad/adal/EventStrings.java
@@ -154,8 +154,8 @@ final class EventStrings {
     static final String AUTHORITY_TYPE_ADFS = "adfs";
     static final String AUTHORITY_TYPE_AAD = "aad";
 
-    static final String AUTHORITY_VALIDATION_SUCCESS = "authority_validation_status_success";
-    static final String AUTHORITY_VALIDATION_FAILURE = "authority_validation_status_failure";
+    static final String AUTHORITY_VALIDATION_SUCCESS = "yes";
+    static final String AUTHORITY_VALIDATION_FAILURE = "no";
     static final String AUTHORITY_VALIDATION_NOT_DONE = "authority_validation_status_not_done";
 
     // Broker account service related events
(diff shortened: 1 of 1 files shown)


adal/src/telemetry/java/com/microsoft/aad/adal/EventStrings.java
```diff
@@ -152,7 +152,7 @@
 
     // Parameter values
     static final String AUTHORITY_TYPE_ADFS = "adfs";
-    static final String AUTHORITY_TYPE_AAD = EVENT_PREFIX + "aad";
+    static final String AUTHORITY_TYPE_AAD = "aad";
 
     static final String AUTHORITY_VALIDATION_SUCCESS = EVENT_PREFIX + "authority_validation_status_success";
```
GTValentine: >EVENT_PREFIX [](start = 55, length = 12)

Looks like in ios there are no prefixes for all parameter values. Should we remove them for static final String AUTHORITY_VALIDATION_SUCCESS = EVENT_PREFIX + "authority_validation_status_success"; as well?

iambmelt: @GTValentine Yep - can do

iambmelt: Corrected in e36273d8a7c69271b1f8dd07d06135aea255f82e

GTValentine: Are you sure it's okay to just change it though? What about all those existing queries?

iambmelt: While this _is_ reported by our telemetry, there is currently no dashboard that I'm aware of that is using these values.

Digging around in the iOS code, it looks like they have an entirely different value specified to track the status: iOS is using 'yes' and 'no' it appears. This change may be inconsequential from an x-plat perspective unless these things are brought into alignment. I'll post a message into our telemetry channel to see how we want to address this

iambmelt: Per sync with @oldalton - let's update these values to `yes` / `no` to make Android match iOS and not force them to make changes too

iambmelt: Updated in 1fdf9a2e5ed6d98e98f4976979a843381f3ea734

adal/src/telemetry/java/com/microsoft/aad/adal/EventStrings.java
```diff
@@ -152,11 +152,11 @@
...
     static final String AUTHORITY_TYPE_ADFS = "adfs";
-    static final String AUTHORITY_TYPE_AAD = EVENT_PREFIX + "aad";
+    static final String AUTHORITY_TYPE_AAD = "aad";
 
-    static final String AUTHORITY_VALIDATION_SUCCESS = EVENT_PREFIX + "authority_validation_status_success";
-    static final String AUTHORITY_VALIDATION_FAILURE = EVENT_PREFIX + "authority_validation_status_failure";
-    static final String AUTHORITY_VALIDATION_NOT_DONE = EVENT_PREFIX + "authority_validation_status_not_done";
+    static final String AUTHORITY_VALIDATION_SUCCESS = "yes";
+    static final String AUTHORITY_VALIDATION_FAILURE = "no";
+    static final String AUTHORITY_VALIDATION_NOT_DONE = "authority_validation_status_not_done";
```
GTValentine: let's rename it to "maybe" and we will have a full house:p

Issue #985, opened 2017-10-10
Discovery#validateAuthority : missing network checking which results in wrong error code turned


heidijinxujia:
There is a http request happen during performInstanceDiscovery(), but there is no network checking performed before sending the http request.
When there is no network, it is expected to return DEVICE_INTERNET_IS_NOT_AVAILABLE. However, it returns DEVELOPER_AUTHORITY_IS_NOT_VALID_URL instead. We should add the network checking here and give the calling app the right error code.

When the authority is not validated yet, it is not on the list. We do a instance discovery to check the authority. DEVELOPER_AUTHORITY_IS_NOT_VALID_URL  should be returned only when the authority url is not valid. However, when the network is not available, the code returns the same error code DEVELOPER_AUTHORITY_IS_NOT_VALID_URL here. We should fail the request earlier, before sending the instance discovery request.

10-10 16:57:04.046 24851-25122/com.microsoft.aad.adal.userappwithbroker E/HttpWebRequest: SERVER_ERROR:2017-10-10 23:57:04-00200771-43df-45ce-951e-dbf47ddf1473-IOException:Unable to resolve host "login.microsoftonline.com": No address associated with hostname ver:1.13.2-RC1 
10-10 16:57:04.048 24851-25122/com.microsoft.aad.adal.userappwithbroker E/Discovery: DEVELOPER_AUTHORITY_IS_NOT_VALID_URL:2017-10-10 23:57:04-00200771-43df-45ce-951e-dbf47ddf1473-Error when validating authority ver:1.13.2-RC1

heidijinxujia:
@addev-ashish

PR #986, opened 2017-10-11
Add network checking when hitting the network in authority validation| Fix 985

heidijinxujia:
Fix for #985 
As there is a network request happening during authority validation, we should check the network when calling performInstanceDiscovery(). If there is no network, it is expected to return DEVICE_INTERNET_IS_NOT_AVAILABLE or NO_NETWORK_CONNECTION_POWER_OPTIMIZATION if the device is in doze mode. 

In this PR,
1. Added property for Discovery class.
2. Added the network checking in performInstanceDiscovery().
3. Added the unit tests.
4. Fixed the related unit tests.

### DIFF
--- a/adal/src/main/java/com/microsoft/aad/adal/AcquireTokenRequest.java
+++ b/adal/src/main/java/com/microsoft/aad/adal/AcquireTokenRequest.java
@@ -63,7 +63,7 @@ class AcquireTokenRequest {
      * Instance validation related calls are serviced inside Discovery as a
      * module.
      */
-    private Discovery mDiscovery = new Discovery();
+    private Discovery mDiscovery;
 
     /**
      * Event Variable for the acquireToken API called. This will track whether the API succeeded or not.
@@ -76,6 +76,7 @@ class AcquireTokenRequest {
     AcquireTokenRequest(final Context appContext, final AuthenticationContext authContext, final APIEvent apiEvent) {
         mContext = appContext;
         mAuthContext = authContext;
+        mDiscovery = new Discovery(mContext);
 
         if (authContext.getCache() != null && apiEvent != null) {
             mTokenCacheAccessor = new TokenCacheAccessor(authContext.getCache(),
@@ -199,9 +200,17 @@ class AcquireTokenRequest {
                 validateAuthority(authorityUrl, authenticationRequest.getUpnSuffix(), authenticationRequest.isSilent(),
                         authenticationRequest.getCorrelationId());
                 apiEvent.setValidationStatus(EventStrings.AUTHORITY_VALIDATION_SUCCESS);
-            } catch (AuthenticationException ex) {
-                apiEvent.setValidationStatus(EventStrings.AUTHORITY_VALIDATION_FAILURE);
-                throw ex;
+            } catch (final AuthenticationException authenticationException) {
+                if (null != authenticationException.getCode()
+                        && (authenticationException.getCode().equals(ADALError.DEVICE_CONNECTION_IS_NOT_AVAILABLE)
+                        || authenticationException.getCode().equals(ADALError.NO_NETWORK_CONNECTION_POWER_OPTIMIZATION))) {
+                    // The authority validation is not done because of network error.
+                    apiEvent.setValidationStatus(EventStrings.AUTHORITY_VALIDATION_NOT_DONE);
+                } else {
+                    apiEvent.setValidationStatus(EventStrings.AUTHORITY_VALIDATION_FAILURE);
+                }
+
+                throw authenticationException;
             } finally {
                 Telemetry.getInstance().stopEvent(authenticationRequest.getTelemetryRequestId(), apiEvent,
                         EventStrings.AUTHORITY_VALIDATION_EVENT);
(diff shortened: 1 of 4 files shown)


adal/src/main/java/com/microsoft/aad/adal/AcquireTokenRequest.java
```diff
@@ -210,6 +210,9 @@ private void performAuthorityValidation(final AuthenticationRequest authenticati
             // check if it contains authority url as the key, if key exists, it means that the authority url validation has happened
             // for the authority already.
             if (!UrlExtensions.isADFSAuthority(authorityUrl) && !AuthorityValidationMetadataCache.containsAuthorityHost(authorityUrl)) {
+                //Check if there is network connection
+                HttpWebRequest.throwIfNetworkNotAvailable(mContext);
```
addev-ashish: I would put this throw inside PerformInstanceDiscovery after line 213, i think the only reason this is here because the Context object is not available to run the check. I would add that to the constructor of Discovery.

addev-ashish: That has the benefit that if the cache is prepopulated and there is an AT in the cache, we will return that successfully. Let me know if i missed something.

heidijinxujia:
@addev-ashish  Comments addressed, thanks.

adal/src/main/java/com/microsoft/aad/adal/AcquireTokenRequest.java
```diff
@@ -213,6 +214,16 @@ private void performAuthorityValidation(final AuthenticationRequest authenticati
...
                     mDiscovery.validateAuthority(authorityUrl);
                 } catch (final AuthenticationException authenticationException) {
+                    if (null != authenticationException.getCode()
+                            && (authenticationException.getCode().equals(ADALError.DEVICE_CONNECTION_IS_NOT_AVAILABLE)
+                            || authenticationException.getCode().equals(ADALError.NO_NETWORK_CONNECTION_POWER_OPTIMIZATION))) {
+                        Logger.w(TAG, "Connection is not available to validate the authority.", "", authenticationException.getCode());
+                        apiEvent.setValidationStatus(EventStrings.AUTHORITY_VALIDATION_FAILURE);
+                        Telemetry.getInstance().stopEvent(authenticationRequest.getTelemetryRequestId(), apiEvent,
+                                EventStrings.AUTHORITY_VALIDATION_EVENT);
+                        throw authenticationException;
```
heidijinxujia: @iambmelt could you help me to check if the Telemetry event is processed correctly? Thanks.

heidijinxujia: @addev-ashish Previously, all authentication exceptions are ignored here. I am not very clear the previous intention of the implementation here. However, I think the authentication exception should not be ignored if the authority validation failed with network connection error. Instead, the authentication exception with network error code should be thrown to the calling app.

heidijinxujia: Do we have a SPEC here to explain why we ignore the authority validation failure here?

heidijinxujia: I am going to remove this part, we should let the flow continue if the validateAuthority is turned off.

adal/src/main/java/com/microsoft/aad/adal/Discovery.java
```diff
@@ -191,6 +194,15 @@ public void setCorrelationId(final UUID requestCorrelationId) {
     }
 
     /**
+     * Set the context for the tenant discovery call.
+     *
+     * @param context
+     */
+    void setContext(final Context context) {
```
iambmelt: What happens if this setter wasn't called and someone invokes `Discovery#performInstanceDiscovery`?

adal/src/main/java/com/microsoft/aad/adal/Discovery.java
```diff
@@ -212,6 +224,9 @@ private void performInstanceDiscovery(final URL authorityUrl, final String trust
             return;
         }
 
+        //Check if the network connection available
+        HttpWebRequest.throwIfNetworkNotAvailable(mContext);
```
iambmelt: This could be `null`, couldn't it?

adal/src/main/java/com/microsoft/aad/adal/Discovery.java
heidijinxujia: Discovery is an internal class in adal and the Context could not passed into the constructor, So we need to call the setter whenever we create a Discovery instance.

adal/src/main/java/com/microsoft/aad/adal/Discovery.java
heidijinxujia: We need to call the setter when creating the Discovery instance. This is the best solution I could found so far.

heidijinxujia:
@iambmelt Comment addressed.

adal/src/main/java/com/microsoft/aad/adal/Discovery.java
heidijinxujia: Let's get rid of the setter.

Issue #1002, opened 2017-10-19
AuthenticationContextTests#testAcquireTokenUserId() doesn't use mock network resources


iambmelt:
This test passes/fails inconsistently: it does not use mocked network resources.

Failure case:
```
com.microsoft.aad.adal.AuthenticationContextTest > testAcquireTokenUserId[test(AVD) - 5.0.2] FAILED 
    org.junit.ComparisonFailure: authority is same expected:<https://login.[windows.net]/common> but was:<https://login.[microsoftonline.com]/common>
    at org.junit.Assert.assertEquals(Assert.java:115)
:adal:connectedDebugAndroidTest FAILED
FAILURE: Build failed with an exception.
```

iambmelt:
Issue fixed on `release/1.13.2` - not on `dev` yet (10/19/2017)

nazukj:
@iambmelt to confirm merge and close.

iambmelt:
Test was removed in [3688b0d617b6196f75719715dccd80ab4985b0e3](https://github.com/AzureAD/azure-activedirectory-library-for-android/commit/3688b0d617b6196f75719715dccd80ab4985b0e3#diff-1f51ca351931fc74098edcfa59eecbd3)
Change was released in `1.13.2+`

PR #1003, opened 2017-10-19
Update AuthenticationContextTest.java to comment out failing test

iambmelt:
Fixes incorrect authority assertion for `login.windows.net`

See #1002


heidijinxujia:
Can we just delete this line or you think it is better to comment it out?

iambmelt:
@heidijinxujia Let's remove it

PR #1005, opened 2017-10-20
Sample app with Sovereign Cloud support

iambmelt:
This PR adds support to the sample app for sovereign cloud

Authority: `AAD_COMMON`
Login hint: **Must be blank** (AAD issue)
Resource: `SIMPLE`
Client id: `BLACKFOREST`
Extra qp: `instance_aware=true`

### DIFF
--- a/userappwithbroker/src/main/java/com/microsoft/aad/adal/example/userappwithbroker/Constants.java
+++ b/userappwithbroker/src/main/java/com/microsoft/aad/adal/example/userappwithbroker/Constants.java
@@ -60,7 +60,8 @@ public class Constants {
     enum ClientId {
         MSDEVEX("b92e0ba5-f86e-4411-8e18-6b5f928d968a"),
         ONEDRIVE("af124e86-4e96-495a-b70a-90f90ab96707"),
-        OFFICE("d3590ed6-52b3-4102-aeff-aad2292ab01c");
+        OFFICE("d3590ed6-52b3-4102-aeff-aad2292ab01c"),
+        BLACKFOREST("f5d01c1c-abe6-4207-ae2d-5bc9af251724");
 
         private final String text;
         ClientId(String s) {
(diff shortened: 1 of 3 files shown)


heidijinxujia:
LGTM so far, I will try with it!

userappwithbroker/src/main/java/com/microsoft/aad/adal/example/userappwithbroker/Constants.java
```diff
@@ -60,7 +60,8 @@ public String getText() {
     enum ClientId {
         MSDEVEX("b92e0ba5-f86e-4411-8e18-6b5f928d968a"),
         ONEDRIVE("af124e86-4e96-495a-b70a-90f90ab96707"),
-        OFFICE("d3590ed6-52b3-4102-aeff-aad2292ab01c");
+        OFFICE("d3590ed6-52b3-4102-aeff-aad2292ab01c"),
+        BLACKFOREST("f5d01c1c-abe6-4207-ae2d-5bc9af251724");
```
addev-ashish: Can you rename to AppCheck2_BF or something like that, we might have more ClientId from them in the future

iambmelt: Done 👍

heidijinxujia:
@iambmelt I've merged my current change on authority updating of sample app into the PR.

iambmelt:
@heidijinxujia Oooph. Yes, let's call this a bug. IMHO the sample should demonstrate usage of the app in all scenarios. Is this sample app _just_ for test, or do we plan to make this the official new sample?

If this should get fixed, doesn't this affect usage in any scenario where interactive auth happens and there's a possibility that the user account changes between sov/non-sov?

As a side question, it never was clear to me why auto-updating the authority was out-of-scope for the sov-cloud changes? Does anybody remember?

/cc @addev-ashish

iambmelt:
Per team mtg - sov cloud / non-sov cloud account switching is an unsupported scenario atm and changing sample is out of scope.

I'll open an issue (P2) and link back here.

/cc @heidijinxujia @addev-ashish

PR #1008, opened 2017-10-23
fix #867, 1/3, ADAL, pass the setting of expiration buffer if broker is installed

### LINKED ISSUE #867: Outlook: AuthenticationSettings.INSTANCE.setExpirationBuffer should work if broker is installed.
If broker is installed this setting is ignored.

piqiums:
fix issue https://github.com/AzureAD/azure-activedirectory-library-for-android/issues/867
this is the change in ADAL library

### DIFF
--- a/adal/src/main/java/com/microsoft/aad/adal/AuthenticationConstants.java
+++ b/adal/src/main/java/com/microsoft/aad/adal/AuthenticationConstants.java
@@ -283,6 +283,9 @@ public final class AuthenticationConstants {
 
         /** String for request id returned from Evo. **/
         public static final String REQUEST_ID_HEADER = "x-ms-request-id";
+
+        /** Integer for token expiration buffer. see AuthenticationSettings.mExpirationBuffer */
+        public static final String EXPIRATION_BUFFER = "aad.expiration.buffer";
     }
 
     /**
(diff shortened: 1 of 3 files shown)


adal/src/main/java/com/microsoft/aad/adal/AuthenticationConstants.java
```diff
@@ -283,6 +283,9 @@ private AuthenticationConstants() { }
 
         /** String for request id returned from Evo. **/
         public static final String REQUEST_ID_HEADER = "x-ms-request-id";
+
+        /** Integer for token expiration buffer. see AuthenticationSettings.mExpirationBuffer */
+        public static final String EXPIRATION_BUFFER = "aad.expiration.buffer";
```
addev-ashish: This should be in the Broker subclass, any reason you put in the AAD subclass?

piqiums: 'expiration buffer' is a concept not only related with broker, but also available for the case without broker, so I didn't put it into the 'Broker' subclass. Maybe 'AAD' is not the best place, do you have any suggestion? thanks

adal/src/main/java/com/microsoft/aad/adal/AuthenticationConstants.java
```diff
@@ -491,6 +491,9 @@ private AuthenticationConstants() { }
         /** String for ssl prefix. */
         public static final String REDIRECT_SSL_PREFIX = "https://";
 
+        //* Integer for token expiration buffer. see AuthenticationSettings.mExpirationBuffer
```
addev-ashish: nit: can you keep the comment format common

Issue #1009, opened 2017-10-25
NPE in Discovery.java found in ADAL 1.13.2


piqiums:
NPE in ADAL 1.13.2 in Discovery.java
################################################
A customer came across an exception in the latest ADAL 1.13.2 drop that we incorporated into our 1711 CP build when testing MAM enrollment.
 
10-25 07:48:27.146   3390  3432 E ncaughtExceptionHandler: Uncaught exception in app
10-25 07:48:27.146   3390  3432 E ncaughtExceptionHandler: java.lang.NullPointerException: Attempt to invoke virtual method 'boolean com.microsoft.aad.adal.InstanceDiscoveryMetadata.isValidated()' on a null object reference
10-25 07:48:27.146   3390  3432 E ncaughtExceptionHandler:  at com.microsoft.aad.adal.Discovery.performInstanceDiscovery(Discovery.java:232)
10-25 07:48:27.146   3390  3432 E ncaughtExceptionHandler:  at com.microsoft.aad.adal.Discovery.validateAuthority(Discovery.java:132)
10-25 07:48:27.146   3390  3432 E ncaughtExceptionHandler:  at com.microsoft.aad.adal.AcquireTokenRequest.performAuthorityValidation(AcquireTokenRequest.java:223)
10-25 07:48:27.146   3390  3432 E ncaughtExceptionHandler:  at com.microsoft.aad.adal.AcquireTokenRequest.validateAcquireTokenRequest(AcquireTokenRequest.java:169)
10-25 07:48:27.146   3390  3432 E ncaughtExceptionHandler:  at com.microsoft.aad.adal.AcquireTokenRequest.access$100(AcquireTokenRequest.java:46)
10-25 07:48:27.146   3390  3432 E ncaughtExceptionHandler:  at com.microsoft.aad.adal.AcquireTokenRequest$1.run(AcquireTokenRequest.java:108)
10-25 07:48:27.146   3390  3432 E ncaughtExceptionHandler:  at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1113)
10-25 07:48:27.146   3390  3432 E ncaughtExceptionHandler:  at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:588)
10-25 07:48:27.146   3390  3432 E ncaughtExceptionHandler:  at java.lang.Thread.run(Thread.java:818)

PR #1010, opened 2017-10-25
fix NPE in Discovery.java found in ADAL 1.13.2

piqiums:
fix https://github.com/AzureAD/azure-activedirectory-library-for-android/issues/1009
previsouly, it's calling like "AuthorityValidationMetadataCache.getCachedInstanceDiscoveryMetadata(authorityUrl).isValidated();" directly without checking if the key 'authorityUrl' is in the key set. After my fix, it's calling 'AuthorityValidationMetadataCache.isAuthorityValidated(authorityUrl)' which has already checked if the key exists in key set.

### DIFF
--- a/adal/src/main/java/com/microsoft/aad/adal/Discovery.java
+++ b/adal/src/main/java/com/microsoft/aad/adal/Discovery.java
@@ -219,10 +219,8 @@ final class Discovery {
         try {
             queryUrl = buildQueryString(trustedHost, getAuthorizationCommonEndpoint(authorityUrl));
             final Map<String, String> discoveryResponse = sendRequest(queryUrl);
-
             AuthorityValidationMetadataCache.processInstanceDiscoveryMetadata(authorityUrl, discoveryResponse);
-
-            result = AuthorityValidationMetadataCache.getCachedInstanceDiscoveryMetadata(authorityUrl).isValidated();
+            result = AuthorityValidationMetadataCache.isAuthorityValidated(authorityUrl);
         } catch (final IOException | JSONException e) {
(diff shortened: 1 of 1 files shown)


addev-ashish:
please create this PR for release/1.13.2, also the fix is missing the call to isValidated

piqiums:
Hi Ashish, I want to talk about two things here:
1. the 'dev' branch is serving the mainline role. Every time we make a fix, the fix should go to dev branch first, then it should be cherry-picked to the specific branch (in this case it's release/1.13.2 branch) . Otherwise, if we check code into the specific branch first, then very possibly the fix might be forgotten to check into dev again. This has happened several times recently and cost some extra communication and efforts. Please consider my suggestion to make the process smooth and high-efficiency. 
2. You said this fix is missing the call to 'isValidated'. I would suggest you to look in to the method body of "isAuthorityValidated". 
```
    static boolean isAuthorityValidated(final URL authorityUrl) {
        return containsAuthorityHost(authorityUrl) && getCachedInstanceDiscoveryMetadata(authorityUrl).isValidated();
    }
```
Let me explain the code expression by expression, 
```
containsAuthorityHost(authorityUrl)
```
It checks if the underlying ConcurrentHashMap's keyset has the key. ConcurrentHashMap doesn't accept 'null' as key or value. If it has the key, then there's a non-null value mapping to the key, then we don't need to worry about the NPE.
```
getCachedInstanceDiscoveryMetadata(authorityUrl).isValidated()
```
Now it's getting the value mapping to the key and calling 'isValidated' on it. It's doing exactly the same job more defensively. I hope you could dive deep into the code when reviewing PRs.
  thanks

PR #1016, opened 2017-10-26
fix issue #1009, part 2, NPE in Discovery.java found in ADAL 1.13.2

piqiums:
1. add the preferred network and preferred cache
2. when the preferred network in invalid, use the authority url that is passed.

### DIFF
--- a/adal/src/main/java/com/microsoft/aad/adal/AcquireTokenRequest.java
+++ b/adal/src/main/java/com/microsoft/aad/adal/AcquireTokenRequest.java
@@ -248,7 +248,7 @@ class AcquireTokenRequest {
         }
 
         // replace the authority if host is not the same as the original one.
-        if (!authorityUrl.getHost().equalsIgnoreCase(metadata.getPreferredNetwork())) {
+        if (metadata.getPreferredNetwork() != null && !authorityUrl.getHost().equalsIgnoreCase(metadata.getPreferredNetwork())) {
             try {
                 final URL replacedAuthority = Utility.constructAuthorityUrl(authorityUrl, metadata.getPreferredNetwork());
                 request.setAuthority(replacedAuthority.toString());
--- a/adal/src/main/java/com/microsoft/aad/adal/AcquireTokenRequest.java
+++ b/adal/src/main/java/com/microsoft/aad/adal/AcquireTokenRequest.java
@@ -253,7 +253,8 @@ class AcquireTokenRequest {
                 final URL replacedAuthority = Utility.constructAuthorityUrl(authorityUrl, metadata.getPreferredNetwork());
                 request.setAuthority(replacedAuthority.toString());
             } catch (final MalformedURLException ex) {
-                throw new AuthenticationException(ADALError.DEVELOPER_AUTHORITY_IS_NOT_VALID_URL, ex.getMessage(), ex);
+                //Intentionally empty.
+                Logger.i(TAG, "preferred network is invalid", "use exactly the same authority url that is passed");
             }
         }
     }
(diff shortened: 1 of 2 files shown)


adal/src/main/java/com/microsoft/aad/adal/AcquireTokenRequest.java
```diff
@@ -253,7 +253,8 @@ private void updatePreferredNetworkLocation(final URL authorityUrl, final Authen
                 final URL replacedAuthority = Utility.constructAuthorityUrl(authorityUrl, metadata.getPreferredNetwork());
                 request.setAuthority(replacedAuthority.toString());
             } catch (final MalformedURLException ex) {
-                throw new AuthenticationException(ADALError.DEVELOPER_AUTHORITY_IS_NOT_VALID_URL, ex.getMessage(), ex);
```
rygo-msft: From my testing, a null preferred network doesn't actually result in a MalformedURLException. The resulting URL just has the empty string as its authority. There should be an explicit null check around the preferred network.

adal/src/main/java/com/microsoft/aad/adal/Discovery.java
```diff
@@ -228,7 +229,10 @@ private void performInstanceDiscovery(final URL authorityUrl, final String trust
             final Map<String, String> discoveryResponse = sendRequest(queryUrl);
             AuthorityValidationMetadataCache.processInstanceDiscoveryMetadata(authorityUrl, discoveryResponse);
             if (!AuthorityValidationMetadataCache.containsAuthorityHost(authorityUrl)) {
-                AuthorityValidationMetadataCache.updateInstanceDiscoveryMap(authorityUrl.getHost(), new InstanceDiscoveryMetadata(true));
+                ArrayList<String> aliases = new ArrayList<String>();
```
heidijinxujia: Should we do the same change in (com/microsoft/aad/adal/AcquireTokenRequest.java:226)?

heidijinxujia: Never mind, I think we don't need it for invalidated authority.

adal/src/main/java/com/microsoft/aad/adal/AcquireTokenRequest.java
heidijinxujia: It could not be null, it the metadata is validated, the preferred network won't be null. If the metadata has not been validated, the performAuthorityValidation will return before reaching the updatePreferredNetworkLocation code. The MalformedURLException will be thrown if the authority is not a valid url.

addev-ashish: @rygo-msft is correct. if authority is null, we will create an authority url as https:///common. Lets add a null check as well

piqiums: updated per comments.

heidijinxujia: According to the change in Discovery, the preferredNetwork won't be null when isValidated==true. And AcquireTokenRequest.java L251 won't be reached unless isValidated==true. I'm OK to add the null check on the getPreferredNetwork. But it seems redundant to me.

heidijinxujia:
#1009

adal/src/main/java/com/microsoft/aad/adal/AcquireTokenRequest.java
rygo-msft: Based on how the constructor for InstanceDiscoveryMetadata(bool) is written, isValidate can be true with preferred network being false. It may be worth refactoring that constructor since InstanceDiscoveryMetadata(false) is the only valid way to call it.

PR #1020, opened 2017-10-30
Add GDPR Compliant PII/OII Filtering to Telemetry

### LINKED ISSUE #1019: GDPR PII/OII Filtering for Telemetry
The following data fields should not be reported in Telemetry without explicit enabling by the developer

|Property|`EventString`|Notes|
|---|---|---|
|Subject|   |Not tracked by telemetry|
|UPN|   |Not tracked by telemetry|
|Given Name|   |Not tracked by telemetry|
|Family Name|   |Not tracked by telemetry|
|Email|   |Not tracked by telemetry|
|OID|   |Not tracked by telemetry|
|Login Hint|`LOGIN_HINT`|   |
|Home OID|   |Not tracked by telemetry|
|Exception traces containing network responses|   |    |
|AT/RT/IdToken|`USER_ID`|   |
|Tenant Id|`TENANT_ID`|   |

iambmelt:
These changes close #1019 - modified `setProperty(String, String)` to omit telemetry fields declared as excluded by `TelemetryUtils#GDPR_FILTERED_FIELDS`.

A switch is also added to override this exclusion: `Telemetry#setAllowPii(boolean)`

### DIFF
--- a/adal/src/telemetry/java/com/microsoft/aad/adal/TelemetryUtils.java
+++ b/adal/src/telemetry/java/com/microsoft/aad/adal/TelemetryUtils.java
@@ -29,7 +29,10 @@ import java.util.HashSet;
 import java.util.Set;
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
-import static com.microsoft.aad.adal.EventStrings.*;
+
+import static com.microsoft.aad.adal.EventStrings.LOGIN_HINT;
+import static com.microsoft.aad.adal.EventStrings.TENANT_ID;
+import static com.microsoft.aad.adal.EventStrings.USER_ID;
 
 final class TelemetryUtils {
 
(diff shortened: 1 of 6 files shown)


adal/src/telemetry/java/com/microsoft/aad/adal/TelemetryUtils.java
```diff
@@ -24,16 +24,36 @@
 package com.microsoft.aad.adal;
 
 import java.io.Serializable;
+import java.util.Arrays;
+import java.util.HashSet;
+import java.util.Set;
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
+import static com.microsoft.aad.adal.EventStrings.*;
```
addev-ashish: i know you are trying to avoid doing EventStrings.LOGIN_HINT et al, which makes sense here, but we do not want to set precedence for importing .* in our code, can you either remove this and change the fields below or import specific fields

iambmelt: Resolved 👍

PR #1036, opened 2017-11-14
Updated ADAL to leverage classess moved to common; removed classess a…

shoatman:
See common repo issue 5


adal/build.gradle
```diff
@@ -239,7 +239,7 @@ task findbugs(type: FindBugs) {
     description 'Run findbugs'
     group 'verification'
 
-    classes = fileTree("build/intermediates/classes/debug/")
+    classes = fileTree("$project.buildDir/intermediates/classes/debug/")
```
iambmelt: Great - thanks for fixing this 👍

Issue #1046, opened 2017-11-21
Handling HTTP 429 error response 


heidijinxujia:
Client libraries should not swallow silently the "Retry-After" header information and provide it in the Service Exceptions for each platform (for instance in AdalServiceException in the case of ADAL.Net). We discussed with STS backend team possibility of including HTTP 429 error response in ADAL Client libraries retry strategy. 

We got to conclusion, that it is not quite retriable from ADAL client libraries perspective - "Retry-After" value is dynamic, but in majority of cases It will be minutes or seconds, which is not acceptable for API called from interactive applications. 

Taking to account throttling mechanisms used with STS services (gateway throttling by IP + app level throttling by user and application id) there is not much sense to retry on it as a way to prevent DDOSing. Also, addition of 429 to the list of retriable errors, can be a bit misleading for application developers - they might assume that they do not have to handle it. 

So, we will go with approach of exposing throttling errors to Developers along with headers received in HTTP response.

heidijinxujia:
@addev-ashish Is the header information only required in status code 429? How about HTTP_BAD_REQUEST and HTTP_UNAUTHORIZED? Do we also need the header info in AuthenticationResult?

Currently, we only treat HTTP_BAD_REQUEST and HTTP_UNAUTHORIZED as OAuth2 error and exposing the body info (like error code and error description) in AuthenticationResult. And throw AuthenticationException in all other case. (https://github.com/AzureAD/azure-activedirectory-library-for-android/blob/df8e3f0c9f265763cd766c49e443e67b82a3db3e/adal/src/main/java/com/microsoft/aad/adal/Oauth2.java#L701)

PR #1050, opened 2017-11-27
Handling HTTP Error Response | Part 1

heidijinxujia:
Fix for #1046 

- Add AuthenticationServiceException for STS error handling

1. If status code == 200, 400, 401, AuthenticationResult with error_code and error_description will be return
2. If status code == 5XX, ServerRespondingWithRetryableException will be thrown and will retry once if getIsExtendedLifetimeEnabled is true.
3. If status code is not expected as before, AuthenticationServiceException  will be thrown which contains httpWebResponse headers info. 
The client app get get these the status code by calling AuthenticationServiceException.getStatusCode() and get the responseBody by calling AuthenticationServiceException.getBody() and get the headers by calling AuthenticationServiceException.getHeaders().

### DIFF
--- a/adal/src/main/java/com/microsoft/aad/adal/AuthenticationServiceException.java
+++ b/adal/src/main/java/com/microsoft/aad/adal/AuthenticationServiceException.java
@@ -1,0 +1,90 @@
+// Copyright (c) Microsoft Corporation.
+// All rights reserved.
+//
+// This code is licensed under the MIT License.
+//
+// Permission is hereby granted, free of charge, to any person obtaining a copy
+// of this software and associated documentation files(the "Software"), to deal
+// in the Software without restriction, including without limitation the rights
+// to use, copy, modify, merge, publish, distribute, sublicense, and / or sell
+// copies of the Software, and to permit persons to whom the Software is
+// furnished to do so, subject to the following conditions :
+//
+// The above copyright notice and this permission notice shall be included in
+// all copies or substantial portions of the Software.
+//
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+// IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+
+package com.microsoft.aad.adal;
+
+import java.util.List;
+import java.util.Map;
+
+/**
+ * ADAL exception for STS error handling.
+ */
+public class AuthenticationServiceException extends AuthenticationException {
+    static final long serialVersionUID = 1;
+
+    private HttpWebResponse mResponse;
+
+    /**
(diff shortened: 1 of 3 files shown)


adal/src/main/java/com/microsoft/aad/adal/AuthenticationServiceException.java
```diff
@@ -0,0 +1,90 @@
...
+
+    /**
+     * Get http web response status code.
+     * @return int status code
+     */
+    public int getStatusCode() {
+        if (mResponse != null) {
+            return mResponse.getStatusCode();
+        } else {
+            throw new IllegalArgumentException("HttpWebResponse could not be NULL.");
```
iambmelt: I would not expect a method which accepts no arguments to throw an `IllegalArgumentException`

adal/src/main/java/com/microsoft/aad/adal/AuthenticationServiceException.java
```diff
@@ -0,0 +1,90 @@
...
+
+    /**
+     * Get http web response body.
+     * @return String The response body for the network call.
+     */
+    public String getBody() {
+        if (mResponse != null) {
+            return mResponse.getBody();
+        } else {
+            throw new IllegalArgumentException("HttpWebResponse could not be NULL.");
```
iambmelt: Same here: feels like a misuse of `IllegalArgumentException` -- can this even happen if the constructor is validating the inputs?

adal/src/main/java/com/microsoft/aad/adal/AuthenticationServiceException.java
```diff
@@ -0,0 +1,90 @@
...
+        }
+    }
+
+    /**
+     * Get the http web response headers.
+     * @return The response headers for the network call.
+     */
+    public Map<String, List<String>> getHeaders() {
+        if (null == mResponse || null == mResponse.getResponseHeaders()) {
+            throw new IllegalArgumentException("Illegal httpWebResponse.");
```
iambmelt: Same complaint about the Exception here...

addev-ashish: don't throw at all, the app should not crash if the server returned garbage? IllegalArgumentException is a Unchecked exception, the app will crash

adal/src/main/java/com/microsoft/aad/adal/AuthenticationServiceException.java
```diff
@@ -0,0 +1,90 @@
...
+package com.microsoft.aad.adal;
+
+import java.util.List;
+import java.util.Map;
+
+/**
+ * ADAL exception for STS error handling.
+ */
+public class AuthenticationServiceException extends AuthenticationException {
+    static final long serialVersionUID = 1;
```
addev-ashish: why do we need this?

adal/src/main/java/com/microsoft/aad/adal/AuthenticationServiceException.java
```diff
@@ -0,0 +1,90 @@
...
+     * @param httpWebResponse
+     */
+    public AuthenticationServiceException(final HttpWebResponse httpWebResponse) {
+        super(ADALError.SERVER_ERROR, getResponseInfo(httpWebResponse));
+        mResponse = httpWebResponse;
+    }
+
+    private static String getResponseInfo(final HttpWebResponse response) {
+        if (null == response) {
+            throw new IllegalArgumentException("HttpWebResponse could not be NULL.");
```
addev-ashish: Same as below

adal/src/main/java/com/microsoft/aad/adal/AuthenticationServiceException.java
heidijinxujia: For serialization and deserialization .

heidijinxujia:
Discussed with Ashish this morning. 
1. Move this into dev branch
2. Instead of creating new exception, we thought it would be a better way to add a property with type Map in AuthenticaitonException for better user experience.

PR #1061, opened 2017-11-30
Update authenticationsettings to proxy to the same class in the commo…

### LINKED ISSUE #1060: Update AuthenticationSettings to refer/use/proxy to common AuthenticationSettings
ADAL has a settings enum/class called AuthenticationSettings.  Common needs many of the same settings.  For now update ADAL AuthenticatinonSettings enum to refer to the common AuthenticationSettings class and proxy calls from one to the other.

shoatman:
Closes #1060

### DIFF
--- a/adal/src/main/java/com/microsoft/aad/adal/AuthenticationSettings.java
+++ b/adal/src/main/java/com/microsoft/aad/adal/AuthenticationSettings.java
@@ -23,11 +23,6 @@
 
 package com.microsoft.aad.adal;
 
-import com.microsoft.identity.common.adal.internal.*;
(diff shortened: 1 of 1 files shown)


adal/src/main/java/com/microsoft/aad/adal/AuthenticationSettings.java
```diff
@@ -23,7 +23,7 @@
 
 package com.microsoft.aad.adal;
 
-import com.microsoft.identity.common.adal.internal.AuthenticationConstants;
+import com.microsoft.identity.common.adal.internal.*;
```
iambmelt: Can you run 'Optimize Imports' on this before merging?

Issue #1063, opened 2017-12-01
Change Thrown Exception for AuthenticationContext#serialize [missing FRT] to be Checked Exception


iambmelt:
See #1049 

The Exception thrown in `com.microsoft.aad.adal.AuthenticationContext.serialize(AuthenticationContext.java:1238)` will now be an instance of `UsageAuthenticationException`

/cc @addev-ashish @nazukj @caltseng 

Tagging milestone: January Release

Issue #1049, opened 2017-11-25
tokenItem does not contain family refresh token


caltseng:
We have customers running into the above exception. Any idea what is happening here? It seems like there might be an issue with tokenshare. 

ADAL version - 1.13.1

We have received reports of this on devices with OS 6.0.1 and 7.0.

```
java.lang.IllegalArgumentException: tokenItem does not contain family refresh token
	at com.microsoft.aad.adal.AuthenticationContext.serialize(AuthenticationContext.java:1238)
	at com.microsoft.aad.adal.ADALAuthenticationContext.serialize(ADALAuthenticationContext.java:20)
	at com.microsoft.ols.o365auth.olsauth_android.O365Auth$1.getToken(O365Auth.java:151)
	at com.microsoft.tokenshare.TokenSharingService$TokenProviderBinder.getToken(TokenSharingService.java:296)
	at com.microsoft.tokenshare.ITokenProvider$Stub.onTransact(ITokenProvider.java:63)
	at android.os.Binder.execTransact(Binder.java:573)
```

iambmelt:
Under investigation: per conversation w/ @addev-ashish - The `IllegalArgumentException` that is throw in this instance will be changed to a `UsageAuthenticationException`

caltseng:
+ @Brian Schnack<mailto:brschn@microsoft.com>

nazukj:
Waiting for reply. @iambmelt to ping for update.

iambmelt:
While it's still unclear as to how the FRT-less token was persisted for this user, the changes made in #1064 will prevent users from experiencing a crash: if we're unable to find a valid a token for the user they will have to reauthenticate. 

@caltseng do you have strong feelings one way or another about pursuing this issue further to try and pin down a root cause? If not, I suggest we close this issue and reopen if users report experiencing frequent auth prompts.

iambmelt:
Email update from @caltseng 

```
1)	This exception is getting thrown when some other app is asking for a token from token share. This means that this exception does not seem to be affecting StaffHub but may be negatively impacting another Microsoft App. 
2)	Catching the exception will result in getToken() returning null. I have verified in TokenManager that returning null for getToken() is an expected case. This means that other apps should be gracefully handling this, as if there were no other tokens found. 
3)	PR for this change is here: <omitted>

I plan to bundle this in with our next release, given that this seems to be a low-risk fix.  

Thanks, 
CalebT
```

[...]

PR #1064, opened 2017-12-01
Changes the thrown Exception for AuthenticationContext#serialize when FRT is missing

iambmelt:
See #1063, #1049 -- This change modifies the thrown Exception to be an instance of `UsageAuthenticationException` (a checked Exception), rather than an `IllegalArgumentException`

### DIFF
--- a/adal/src/main/java/com/microsoft/aad/adal/AuthenticationContext.java
+++ b/adal/src/main/java/com/microsoft/aad/adal/AuthenticationContext.java
@@ -1254,7 +1254,7 @@ public class AuthenticationContext {
         if (!StringExtensions.isNullOrBlank(tokenItem.getFamilyClientId())) {
             return SSOStateSerializer.serialize(tokenItem);
         } else {
-            throw new IllegalArgumentException("tokenItem does not contain family refresh token");
+            throw new UsageAuthenticationException(ADALError.FAIL_TO_EXPORT, "tokenItem does not contain family refresh token");
         }
     }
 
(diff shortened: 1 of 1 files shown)


iambmelt:
@caltseng -- while this doesn't fix the issue, it will prevent this unchecked Exception from being thrown when `serialize()` fails to find an FRT

adal/src/main/java/com/microsoft/aad/adal/AuthenticationContext.java
```diff
@@ -1254,7 +1254,7 @@ String serialize(final String uniqueUserId) throws AuthenticationException {
         if (!StringExtensions.isNullOrBlank(tokenItem.getFamilyClientId())) {
             return SSOStateSerializer.serialize(tokenItem);
         } else {
-            throw new IllegalArgumentException("tokenItem does not contain family refresh token");
+            throw new UsageAuthenticationException(ADALError.FAIL_TO_EXPORT, "tokenItem does not contain family refresh token");
```
iambmelt: @addev-ashish This seemed like the _most_ appropriate `ADALError`. Lmk if you have an alternative suggestion

iambmelt:
Should update the changelog too

adal/src/main/java/com/microsoft/aad/adal/AuthenticationContext.java
addev-ashish: This looks good to me. A minor nit, can we change the condition around. !IsNull looks like a double negative to me. 

Something like 
if (isNull)
    throw UsageAuthenticationException;
// no need for else
return SSOStateSerializer.serialize(tokenItem);

iambmelt: Sounds great -- will do 👍

iambmelt: Fixed in 8effb3af67d98f73a808ad293bac54d65f430fce

PR #1065, opened 2017-12-04
Returning the HTTP Error Response code back to the apps when ADAL has it

heidijinxujia:
Fix for #1046 
1. Update the AuthenticationException and AuthenticatoinResult
- the developer can get "Retry-After" headers by calling **AuthenticationException.getHttpResponseHeaders()**
- get response body by calling **AuthenticationException.getResponseBody()**. 
To get the the error code, like "invalid_grant" from the response body, you can calling **AuthenticationException.getResponseBody().get(AuthenticationConstants.OAuth2.ERROR)**.
And call **AuthenticationException.getResponseBody().get(AuthenticationConstants.OAuth2.ERROR_DESCRIPTION)** to get the error description.
- the developer can get the response status code by calling AuthenticationException.getServiceStatusCode()

2. Add the service response error handling. 

| Status Code  | Expected Behavior |
| ------------- | ------------- |
| HTTP_OK | return AuthenticationResult with access token |
| HTTP_BAD_REQUEST, HTTP_UNAUTHORIZED | return AuthenticationResult with error, error_codes and error_description, with headers and body as properties |
| HTTP_UNAUTHORIZED ~ 599 | throw ServerRespondingWithRetryableException, with headers and body as properties |
| Other status code | AuthenticationException with headers and body as properties |

3. Add the unit tests as well.


adal/src/main/java/com/microsoft/aad/adal/AcquireTokenSilentHandler.java
```diff
@@ -164,7 +164,7 @@ AuthenticationResult acquireTokenWithRefreshToken(final String refreshToken)
 
             throw new AuthenticationException(
                     ADALError.AUTH_FAILED_NO_TOKEN, ExceptionExtensions.getExceptionMessage(exc),
-                    new AuthenticationException(ADALError.SERVER_ERROR, exc.getMessage()));
+                    new AuthenticationException(ADALError.SERVER_ERROR, exc.getMessage(), exc));
```
iambmelt: Looking at this code path, there might be something misleading going on here... If this catches a lower Exception thrown for `4XX`-level HTTP status codes (say, a `429`), these can't accurately be called instances of `SERVER_ERROR` since `4XX` error codes are reserved for client errors (Bad Request, Unauthorized, Forbidden, Not Found, etc) -- in some instances `INVALID_REQUEST` may be _more_ appropriate but this also seems like the wrong place to start conditionally checking thrown Exceptions...

Since this code path _may_ be rethrowing an `AuthenticationException`, can we check if there's an existing `ADALError` that may be populated on the Exception before rethrowing, rather than assume it's a `SERVER_ERROR`?

heidijinxujia: Good point, this part confused me as well. Currently adal we don't have specific ADALError enum constant for every http status code or status code group like 4XX, we treat all status code (except for 200, 400, 401) as ADALError.SERVER_ERROR ("Server returned an error"). The developer can get the specific error in String by calling AuthenticationException.getResponseBody().get(AuthenticationConstants.OAuth2.ERROR).

The HTTP web response error is a String type, not an ADALError. We will need the full list of error code if we want to subdivide the error code here. Please let me know you thoughts. @iambmelt @addev-ashish I'm open mind on how we label the http status code in AuthenticationException.

iambmelt: For this case, I might suggest something _roughly_ equivalent to the following:

```java
new AuthenticationException(
        null == exc.getCode() ? ADALError.SERVER_ERROR : exc.getCode(),
        exc.getMessage(),
        exc
    ));
```

Clearly this doesn't solve the _larger issue_ of possibly returning the wrong `ADALError` instance, but I can live with this for now.

I'm fine to make an issue to review our thrown errors and revisit as part of another cleanup/refactor effort.

/cc @addev-ashish @nazukj

heidijinxujia: Sorry, @iambmelt I don't understand why we need to do the check here. The AuthenticationException error code should take the outermost layer's value. In this case, in Line 166, the error code is AUTH_FAILED_NO_TOKEN. We are not replacing the error code with ADALError.SERVER_ERROR. I'm confused why we need to add this checking.
```java
new AuthenticationException(
        null == exc.getCode() ? ADALError.SERVER_ERROR : exc.getCode(),
        exc.getMessage(),
        exc
    ));
```
The ADALError.SERVER_ERROR is a general error code for all http error response, not for specific status code. I totally agree the name itself is misleading and we should add more ADALError to represent the status code better. But I'm not sure if this is the purpose of issue #1046 

And if the developer want to get the http web response error code, for example, if we get the http response body like this
{"error":"interaction_required",
"error_description":"AADSTS50131: Windows device is not in required device state: domain_joined\r\nTrace ID: 41f28fdd-8c37-4f3e-8e51-2eeea28e0700\r\nCorrelation ID: 058ff775-5dfe-4e0a-901f-2440103cfb59\r\nTimestamp: 2017-10-11 00:20:23Z",
"error_codes":[50131],
"timestamp":"2017-10-11 00:20:23Z",
"trace_id":"41f28fdd-8c37-4f3e-8e51-2eeea28e0700",
"correlation_id":"058ff775-5dfe-4e0a-901f-2440103cfb59"}
they can call AuthenticationException.getResponseBody().get(AuthenticationConstants.OAuth2.ERROR) and get "interaction_required".

@addev-ashish Please help me to confirm if there is anything I misunderstand. Thanks. :)

iambmelt: Thanks for clarifying, @heidijinxujia -- I'll say :shipit:

heidijinxujia: I filed the issue we discussed here #1068 @iambmelt

adal/src/main/java/com/microsoft/aad/adal/AuthenticationException.java
```diff
@@ -33,6 +41,12 @@
 
     private ADALError mCode;
 
+    private Map<String, String> mResponseBody;
+
+    private int mServiceStatusCode = -1;
+
+    private Map<String, List<String>> mHttpResponseHeaders;
```
heidijinxujia: Hi @iambmelt I am wondering if it will be better if we combine the mResponseBody and mHttpResponseHeaders into a single Map<String, List<String>> map? 
Pros: it will make the class more extendable 
Cons: body and headers will be stored together

How do you think?

iambmelt: I think I'm _personally_ preferential to keeping them separate. Given that headers and the response body are two distinct and separate parts of the HTTP response, it seems logical to me that they would be exposed through separate API calls on this object.

adal/src/main/java/com/microsoft/aad/adal/AuthenticationException.java
```diff
@@ -33,6 +41,12 @@
 
     private ADALError mCode;
 
+    private Map<String, String> mResponseBody;
+
+    private int mServiceStatusCode = -1;
```
addev-ashish: can we make all these as final, or as many as we can

adal/src/main/java/com/microsoft/aad/adal/AuthenticationException.java
```diff
@@ -33,6 +41,12 @@
 
     private ADALError mCode;
 
+    private Map<String, String> mResponseBody;
```
addev-ashish: can we initialize both the maps to null, so that in non-http cases we are not returning uninitialized values.

adal/src/main/java/com/microsoft/aad/adal/AuthenticationException.java
addev-ashish: not the right data structure, the spec says Map<String, String>, use HashMapExtensions.getJsonResponse

adal/src/main/java/com/microsoft/aad/adal/AuthenticationException.java
```diff
@@ -70,6 +84,89 @@ public AuthenticationException(ADALError code, String details) {
...
+        mCode = code;
+
+        if (null == response) {
+            return;
+        }
+
+        mServiceStatusCode = response.getStatusCode();
+
+        if (null != response.getResponseHeaders()) {
+            mHttpResponseHeaders = response.getResponseHeaders();
```
addev-ashish: HashMapExtensions.getJsonResponse

adal/src/main/java/com/microsoft/aad/adal/AuthenticationException.java
```diff
@@ -70,6 +84,89 @@ public AuthenticationException(ADALError code, String details) {
...
+        }
+
+        if (null != response.getBody()) {
```
addev-ashish: use HashMapExtensions.getJsonResponse, see if you can avoid this code duplication

adal/src/main/java/com/microsoft/aad/adal/AuthenticationException.java
```diff
@@ -70,6 +84,89 @@ public AuthenticationException(ADALError code, String details) {
...
+                                   Throwable throwable) {
+        this(code, details, throwable);
+
+        if (null == response) {
+            return;
+        }
+
+        mServiceStatusCode = response.getStatusCode();
+
+        if (null != response.getResponseHeaders()) {
```
addev-ashish: don't duplicate code

adal/src/main/java/com/microsoft/aad/adal/AuthenticationException.java
heidijinxujia: Hi @addev-ashish The spec didn't say the header is in type Map<String, String>. In addition, the responseHeaders in ADAL is defined as Map<String, List<String>>, please refer to HttpURLConnection.getHeaderFields().

heidijinxujia: https://developer.android.com/reference/java/net/URLConnection.html#getHeaderFields()

adal/src/main/java/com/microsoft/aad/adal/AuthenticationException.java
heidijinxujia: HashMapExtensions.getJsonResponse is used to convert the response body, which is a String, into Map<String, String>.

adal/src/main/java/com/microsoft/aad/adal/AuthenticationException.java
nazukj: >mResponseBody [](start = 32, length = 13)

nit: please keep variable names consistent. mHttpResponseBody?

adal/src/main/java/com/microsoft/aad/adal/AuthenticationException.java
```diff
@@ -70,6 +84,89 @@ public AuthenticationException(ADALError code, String details) {
...
         mCode = code;
+
+        if (null == throwable) {
+            return;
+        }
+
+        if (throwable instanceof AuthenticationException) {
+            mServiceStatusCode = ((AuthenticationException) throwable).getServiceStatusCode();
+            mResponseBody = ((AuthenticationException) throwable).getResponseBody();
+            mHttpResponseHeaders = ((AuthenticationException) throwable).getHttpResponseHeaders();
```
nazukj: nit: if its more than once, should we create a variable that casts it to AuthenticationException

adal/src/main/java/com/microsoft/aad/adal/AuthenticationException.java
```diff
@@ -70,6 +84,89 @@ public AuthenticationException(ADALError code, String details) {
...
+        }
+    }
+
+    /**
+     * @param code Resource file related error code. Message will be derived
+     *            from resource with using app context
```
nazukj: >with using [](start = 32, length = 10)

nit: resource using

adal/src/main/java/com/microsoft/aad/adal/ServerRespondingWithRetryableException.java
```diff
@@ -47,4 +47,8 @@ public ServerRespondingWithRetryableException(String msg) {
     public ServerRespondingWithRetryableException(String msg, Throwable throwable) {
         super(null, msg, throwable);
     }
+
+    public ServerRespondingWithRetryableException(String msg, HttpWebResponse response) {
```
nazukj: >public ServerRespondi [](start = 4, length = 21)

please add comment

adal/src/main/java/com/microsoft/aad/adal/AuthenticationException.java
heidijinxujia: The status code cannot be final. Ref Line#93.

adal/src/main/java/com/microsoft/aad/adal/AuthenticationException.java
heidijinxujia: Addressed.

adal/src/main/java/com/microsoft/aad/adal/AuthenticationException.java
heidijinxujia: Sorry, do you mean Line 148 is duplicated? Addressed.

addev-ashish: No, this function and the previous one have the same code repeated. Line 111-122 and 141 to 152 are same

adal/src/main/java/com/microsoft/aad/adal/AuthenticationException.java
```diff
@@ -33,6 +39,12 @@
 
     private ADALError mCode;
 
+    private HashMap<String, String> mHttpResponseBody = null;
```
iambmelt: Can we change this left-side to Map<String, String>? (Coding to interface instead of concrete class)

adal/src/main/java/com/microsoft/aad/adal/AuthenticationException.java
```diff
@@ -33,6 +39,12 @@
...
+
+    private int mServiceStatusCode = -1;
+
+    private HashMap<String, List<String>> mHttpResponseHeaders = null;
```
iambmelt: Same as above

adal/src/main/java/com/microsoft/aad/adal/AuthenticationException.java
```diff
@@ -86,6 +139,73 @@ public String getMessage() {
...
+    public int getServiceStatusCode() {
+        return mServiceStatusCode;
+    }
+
+    /**
+     * Gets the response body that may be returned by the service.
+     *
+     * @return response body map, null if not initialized.
+     */
+    public HashMap<String, String> getHttpResponseBody() {
```
iambmelt: Let's change this to `public Map<String, String> getHttpResponseBody()` - more interface  vs concrete stuff :)

adal/src/main/java/com/microsoft/aad/adal/AuthenticationException.java
```diff
@@ -86,6 +139,73 @@ public String getMessage() {
...
+        return mHttpResponseBody;
+    }
+
+    /**
+     * Get the response headers that indicated an error.
+     *
+     * @return The response headers for the network call, null if not initialized.
+     */
+    public HashMap<String, List<String>> getHttpResponseHeaders() {
```
iambmelt: Same here ^^

adal/src/main/java/com/microsoft/aad/adal/AuthenticationException.java
```diff
@@ -86,6 +139,73 @@ public String getMessage() {
...
+        return mHttpResponseHeaders;
+    }
+
+    void setHttpResponseBody(HashMap<String, String> body) {
```
iambmelt: Same for this param

adal/src/main/java/com/microsoft/aad/adal/AuthenticationException.java
```diff
@@ -86,6 +139,73 @@ public String getMessage() {
...
+        mHttpResponseBody = body;
+    }
+
+    void setHttpResponseHeaders(HashMap<String, List<String>> headers) {
```
iambmelt: first param here

adal/src/main/java/com/microsoft/aad/adal/AuthenticationResult.java
```diff
@@ -95,6 +99,12 @@
 
     private CliTelemInfo mCliTelemInfo;
 
+    private HashMap<String, String> mHttpResponseBody = null;
```
iambmelt: Same here (`Map` vs `HashMap`)

adal/src/main/java/com/microsoft/aad/adal/AuthenticationResult.java
```diff
@@ -95,6 +99,12 @@
...
+
+    private int mServiceStatusCode = -1;
+
+    private HashMap<String, List<String>> mHttpResponseHeaders = null;
```
iambmelt: This one too

adal/src/main/java/com/microsoft/aad/adal/AuthenticationResult.java
```diff
@@ -399,4 +409,48 @@ final CliTelemInfo getCliTelemInfo() {
     final void setCliTelemInfo(final CliTelemInfo cliTelemInfo) {
         mCliTelemInfo = cliTelemInfo;
     }
+
+    void setHttpResponseBody(final HashMap<String, String> body) {
```
iambmelt: This param

adal/src/main/java/com/microsoft/aad/adal/AuthenticationResult.java
```diff
@@ -399,4 +409,48 @@ final CliTelemInfo getCliTelemInfo() {
...
+        mHttpResponseBody = body;
+    }
+
+    public HashMap<String, String> getHttpResponseBody() {
```
iambmelt: Return type

adal/src/main/java/com/microsoft/aad/adal/AuthenticationResult.java
```diff
@@ -399,4 +409,48 @@ final CliTelemInfo getCliTelemInfo() {
...
+        return mHttpResponseBody;
+    }
+
+    void setHttpResponseHeaders(final HashMap<String, List<String>> headers) {
```
iambmelt: param

adal/src/main/java/com/microsoft/aad/adal/AuthenticationResult.java
```diff
@@ -399,4 +409,48 @@ final CliTelemInfo getCliTelemInfo() {
...
+        mHttpResponseHeaders = headers;
+    }
+
+    public HashMap<String, List<String>> getHttpResponseHeaders() {
```
iambmelt: return type

adal/src/main/java/com/microsoft/aad/adal/BrokerProxy.java
```diff
@@ -481,8 +485,20 @@ private AuthenticationResult getResultFromBrokerResponse(final Bundle bundleResu
 
             throw new AuthenticationException(adalErrorCode, msg);
         } else if (!StringExtensions.isNullOrBlank(oauth2ErrorCode) && request.isSilent()) {
-            throw new AuthenticationException(ADALError.AUTH_REFRESH_FAILED_PROMPT_NOT_ALLOWED,
+            final AuthenticationException exception = new AuthenticationException(ADALError.AUTH_REFRESH_FAILED_PROMPT_NOT_ALLOWED,
                     "Received error from broker, errorCode: " + oauth2ErrorCode + "; ErrorDescription: " + oauth2ErrorDescription);
+            final Serializable responseBody = bundleResult.getSerializable(AuthenticationConstants.OAuth2.HTTP_RESPONSE_BODY);
+            final Serializable responseHeaders = bundleResult.getSerializable(AuthenticationConstants.OAuth2.HTTP_RESPONSE_HEADER);
+            if (null != responseBody && responseBody instanceof HashMap) {
```
iambmelt: One params are updated, let's change this check to `instanceof Map`

adal/src/main/java/com/microsoft/aad/adal/BrokerProxy.java
```diff
@@ -481,8 +485,20 @@ private AuthenticationResult getResultFromBrokerResponse(final Bundle bundleResu
...
+                exception.setHttpResponseBody((HashMap)responseBody);
```
iambmelt: Can this cast be avoided? If not, let's change the cast to `Map`

adal/src/main/java/com/microsoft/aad/adal/BrokerProxy.java
```diff
@@ -481,8 +485,20 @@ private AuthenticationResult getResultFromBrokerResponse(final Bundle bundleResu
...
+            }
+
+            if (null != responseHeaders && responseHeaders instanceof HashMap) {
```
iambmelt: Same as above re: `Map` vs. `HashMap`

adal/src/main/java/com/microsoft/aad/adal/BrokerProxy.java
```diff
@@ -481,8 +485,20 @@ private AuthenticationResult getResultFromBrokerResponse(final Bundle bundleResu
...
+                exception.setHttpResponseHeaders((HashMap)responseHeaders);
```
iambmelt: Same as above re: cast

adal/src/main/java/com/microsoft/aad/adal/HashMapExtensions.java
```diff
@@ -110,4 +113,36 @@ private HashMapExtensions() {
...
+            while (i.hasNext()) {
+                final String key = (String) i.next();
+                responseItems.put(key, jsonObject.getString(key));
+            }
+        }
+
+        return responseItems;
+    }
+
+    static HashMap<String, List<String>> jsonStringAsMapList(String headerStr) throws JSONException {
```
iambmelt: Since there's potentially some good reusability on this method, can you add a quick blurb of javadoc to summarize what it does, what it's inputs/outputs are?

adal/src/main/java/com/microsoft/aad/adal/HashMapExtensions.java
```diff
@@ -110,4 +113,36 @@ private HashMapExtensions() {
         return response;
     }
 
+    static HashMap<String, String> jsonStringAsMap(String bodyStr) throws JSONException {
```
iambmelt: Since there's potentially some good reusability on this method, can you add a quick blurb of javadoc to summarize what it does, what it's inputs/outputs are?

adal/src/main/java/com/microsoft/aad/adal/HashMapExtensions.java
```diff
@@ -110,4 +113,36 @@ private HashMapExtensions() {
...
+        final HashMap<String, String> responseItems = new HashMap<>();
```
iambmelt: Let's also consider changing the return type to `Map` -- I know the class is called _HashMapExtensions_, but we should still try to code to an interface where possible :-)

adal/src/main/java/com/microsoft/aad/adal/HashMapExtensions.java
```diff
@@ -110,4 +113,36 @@ private HashMapExtensions() {
...
+                final String key = (String) i.next();
+                responseItems.put(key, jsonObject.getString(key));
+            }
+        }
+
+        return responseItems;
+    }
+
+    static HashMap<String, List<String>> jsonStringAsMapList(String headerStr) throws JSONException {
+        final HashMap<String, List<String>> responseItems = new HashMap<>();
```
iambmelt: Same for this return type -- even if the class name no longer makes sense.

iambmelt:
Can we rename the class `HashMapExtensions` to `MapExtensions` without any problems? It looks like yes: the class is not `public` -- I'm not crazy about a class of extensions for a concrete type.

adal/src/main/java/com/microsoft/aad/adal/AuthenticationException.java
heidijinxujia: We have to use HashMap instead of Map. Because AuthenticationException is Serializable.

adal/src/main/java/com/microsoft/aad/adal/AuthenticationException.java
heidijinxujia: Same above.

adal/src/main/java/com/microsoft/aad/adal/AuthenticationException.java
heidijinxujia: Same above. We need to parse it into json object when passing it between apps with intent bundle.

adal/src/main/java/com/microsoft/aad/adal/AuthenticationException.java
heidijinxujia: Same above.

adal/src/main/java/com/microsoft/aad/adal/AuthenticationException.java
heidijinxujia: Same above.

adal/src/main/java/com/microsoft/aad/adal/AuthenticationException.java
heidijinxujia: Same above.

adal/src/main/java/com/microsoft/aad/adal/AuthenticationResult.java
heidijinxujia: Same above.

adal/src/main/java/com/microsoft/aad/adal/AuthenticationResult.java
heidijinxujia: Same above.

adal/src/main/java/com/microsoft/aad/adal/AuthenticationResult.java
heidijinxujia: Same above.

adal/src/main/java/com/microsoft/aad/adal/AuthenticationResult.java
heidijinxujia: Same above.

adal/src/main/java/com/microsoft/aad/adal/AuthenticationResult.java
heidijinxujia: Same above.

adal/src/main/java/com/microsoft/aad/adal/AuthenticationResult.java
heidijinxujia: Same above.

adal/src/main/java/com/microsoft/aad/adal/BrokerProxy.java
heidijinxujia: Same above.

PR #1066, opened 2017-12-04
Shoatman/unified/dev/1057

shoatman:
Update ADAL to leverage common core for caching.  Changes include:

- TokenCacheAccessor - Update to use common cache when saving tokens (based on whether DefaultTokenCache was in use)
- Updates to tests to fix issues related to when Authority Host / Cloud Metadata data was loaded
- Updates to tests to include raw id_tokens to ensure proper behavior in common (TODO: Add support for ADFS.... we may only want to support common cache in the case of AAD)

### DIFF
--- a/adal/src/androidTest/java/com/microsoft/aad/adal/AuthenticationContextTest.java
+++ b/adal/src/androidTest/java/com/microsoft/aad/adal/AuthenticationContextTest.java
@@ -33,6 +33,7 @@ import android.content.pm.Signature;
 import android.os.Build;
 import android.os.Bundle;
 import android.support.test.InstrumentationRegistry;
+import android.support.test.filters.Suppress;
 import android.support.test.runner.AndroidJUnit4;
 import android.test.UiThreadTest;
 import android.util.Base64;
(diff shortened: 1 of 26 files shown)


adal/src/androidTest/java/com/microsoft/aad/adal/AuthenticationContextTest.java
```diff
@@ -110,7 +114,7 @@
      */
     private static final String VALID_AUTHORITY = "https://login.windows.net/test.onmicrosoft.com";
 
-    protected static final int CONTEXT_REQUEST_TIME_OUT = 20000;
+    protected static final int CONTEXT_REQUEST_TIME_OUT = 20000; //20000
```
iambmelt: nit - stray comment

iambmelt:
We may also want to create an issue for ADFS support since that appears to remain  a TODO item

adal/src/androidTest/java/com/microsoft/aad/adal/AuthenticationContextTest.java
shoatman: I expanded on this comment to provide a tip for future developers.  Thanks for the reminder.

shoatman:
I added an issue to address ADFS scenario.  See #1067

Issue #1080, opened 2017-12-19
upgrading Android Studio to 3.0 


piqiums:
Currently, we're using Android Studio 2.x. We need to use Android Studio 3.0 in soon. If we build directly with Android Studio 3.0, there will be some issue blocking the building. The building should be unblocked.
Also , please update all 3 repos: public adal, ad-accounts and adal unity.

piqiums:
the change in public ADAL is checked in. will do the same for Unity and ad-account

piqiums:
the change for Unity and ad-account is also checked in.

PR #1081, opened 2017-12-19
fix android studio 3.0 build issue by upgrading gradle version from v2.14.1 to v3.5

piqiums:
fix android studio 3.0 build issue by upgrading gradle version from v2.14.1 to v3.5
fix https://github.com/AzureAD/azure-activedirectory-library-for-android/issues/1080

### DIFF
--- a/adal/src/main/java/com/microsoft/aad/adal/WebFingerMetadataRequestor.java
+++ b/adal/src/main/java/com/microsoft/aad/adal/WebFingerMetadataRequestor.java
@@ -99,6 +99,7 @@ class WebFingerMetadataRequestor
      * @return the URL of the WebFinger document
      * @throws MalformedURLException if the URL could not be constructed
      */
+    @SuppressWarnings("PMD")
     static URL buildWebFingerUrl(final URL resource, final DRSMetadata drsMetadata)
             throws MalformedURLException {
         final URL passiveAuthEndpoint = new URL(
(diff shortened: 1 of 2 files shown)


adal/src/main/java/com/microsoft/aad/adal/WebFingerMetadataRequestor.java
```diff
@@ -99,6 +99,7 @@ WebFingerMetadata parseMetadata(final HttpWebResponse webResponse) throws Authen
      * @return the URL of the WebFinger document
      * @throws MalformedURLException if the URL could not be constructed
      */
+    @SuppressWarnings("PMD")
```
heidijinxujia: Do you know what kind of PMD issue reported here?

piqiums: InsufficientStringBufferDeclaration, it's quite trival. 
http://pmd.sourceforge.net/pmd-4.3.0/rules/strings.html ,
 search InsufficientStringBufferDeclaration : 

Failing to pre-size a StringBuffer properly could cause it to re-size many times during runtime. This rule checks the characters that are actually passed into StringBuffer.append(), but represents a best guess "worst case" scenario. An empty StringBuffer constructor initializes the object to 16 characters. This default is assumed if the length of the constructor can not be determined.

This rule is defined by the following Java class: net.sourceforge.pmd.rules.strings.InsufficientStringBufferDeclaration

Example:

                
```
public class Foo {
    void bar() {
        StringBuffer bad = new StringBuffer();
        bad.append("This is a long string, will exceed the default 16 characters");//bad
        StringBuffer good = new StringBuffer(41);
        good.append("This is a long string, which is pre-sized");//good
    }
}
```

piqiums:
the change in public ADAL is checked in. will do the same for Unity and ad-account

Issue #1078, opened 2017-12-19
URL Query Parameter Parsing Ignores Key/Value Pairs When Value is Empty


iambmelt:
See [this method](https://github.com/AzureAD/azure-activedirectory-library-for-android/blob/dev/adal/src/main/java/com/microsoft/aad/adal/HashMapExtensions.java#L65-L94).

[RFC 3986](https://tools.ietf.org/html/rfc3986#section-3.4) doesn't mandate a value for each key. We should account for this.

The Map returned by `URLFormDecodeData` should contain the key with a `null` or `""` value.

PR #1082, opened 2017-12-21
URL Query Parameter Parsing | Fix for #1078

heidijinxujia:
RFC  #1078 

Update the URI parsing to handle the situation when parsing the key with empty value case.

This is because the certAuthorities list returned from server could be null when either no certificate or no permission for ADFS service account for the Device container in AD. We should return the NoDeviceCertResponse instead of throwing the DEVICE_CERTIFICATE_RESPONSE_FAILED AuthenticationException. In addition, we should not store the key value when parsing failed.

### DIFF
--- a/adal/src/androidTest/java/com/microsoft/aad/adal/ChallengeResponseBuilderTests.java
+++ b/adal/src/androidTest/java/com/microsoft/aad/adal/ChallengeResponseBuilderTests.java
@@ -308,14 +308,6 @@ public class ChallengeResponseBuilderTests extends AndroidTestHelper {
 
         try {
             m.invoke(handler, CERT_REDIRECT
-                    + "?Nonce=2&CertAuthoritiesMissing=ABC&Version=1.0&SubmitUrl=1&Context=1");
-            Assert.fail("No exception");
-        } catch (Exception ex) {
-            assertTrue("Argument exception", ex.getCause().getMessage().contains("CertAuthorities"));
-        }
-
-        try {
-            m.invoke(handler, CERT_REDIRECT
                     + "?Nonce=2&CertAuthorities=ABC&Version=1.0&SubmitUrlMissing=1&Context=1");
             Assert.fail("No exception");
         } catch (Exception ex) {
--- a/adal/src/androidTest/java/com/microsoft/aad/adal/ChallengeResponseBuilderTests.java
+++ b/adal/src/androidTest/java/com/microsoft/aad/adal/ChallengeResponseBuilderTests.java
@@ -308,6 +308,13 @@ public class ChallengeResponseBuilderTests extends AndroidTestHelper {
 
         try {
             m.invoke(handler, CERT_REDIRECT
+                    + "?Nonce=2&CertAuthorities=&Version=1.0&SubmitUrl=1&Context=1");
+        } catch (Exception ex) {
+            Assert.fail("No exception");
+        }
+
+        try {
+            m.invoke(handler, CERT_REDIRECT
                     + "?Nonce=2&CertAuthorities=ABC&Version=1.0&SubmitUrlMissing=1&Context=1");
             Assert.fail("No exception");
         } catch (Exception ex) {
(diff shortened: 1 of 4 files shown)


adal/src/main/java/com/microsoft/aad/adal/HashMapExtensions.java
```diff
@@ -71,22 +71,28 @@ private HashMapExtensions() {
...
+                } else if (elements.length == 1) {
+                    try {
+                        key = StringExtensions.urlFormDecode(elements[0].trim());
+                        value = "";
+                    } catch (UnsupportedEncodingException e) {
+                        Logger.i(TAG, ADALError.ENCODING_IS_NOT_SUPPORTED.getDescription(), e.getMessage(), null);
                     }
                 }
+
+                if (!StringExtensions.isNullOrBlank(key)) {
```
iambmelt: If the above `UnsupportedEncodingException` is thrown, won't this try to `put(null, null)`? Would it make more sense to skip these values?

adal/src/androidTest/java/com/microsoft/aad/adal/ChallengeResponseBuilderTests.java
```diff
@@ -306,14 +306,6 @@ public void testGetChallengeResponseInvalidRedirect() throws ClassNotFoundExcept
             assertTrue("Argument exception", ex.getCause().getMessage().contains("Nonce"));
         }
 
-        try {
-            m.invoke(handler, CERT_REDIRECT
-                    + "?Nonce=2&CertAuthoritiesMissing=ABC&Version=1.0&SubmitUrl=1&Context=1");
```
iambmelt: Can you add a test where the `CertAuthorities` query param is returned in the same manner that originally uncovered this error? A test where it takes the form of `&CertAuthorities=` ? (empty list case)

adal/src/main/java/com/microsoft/aad/adal/HashMapExtensions.java
heidijinxujia: `put(null, null)` would never happen with this check on Line 93. If UnsupportedEncodingException is thrown when decoding the value part, we will log the exception and if the key is not null or empty, we will put the key-value pair with empty value, which is requested in this PR.

addev-ashish: but put (key, null) can now happen which would have never happened before.

heidijinxujia: Hi @addev-ashish I think this is by design. The purpose of this PR is to accept empty paras like CAs rather than throw AuthenticationException.

iambmelt: Just confirming: this _is_ by design. I'm OK with `null` but it does mean that we need to be _very intentional_ when checking for the presence of keys by using `map.containsKey()` and not just using `map.get(key) != null`.

Preferentially speaking, would `""` be a better value? @addev-ashish @heidijinxujia

heidijinxujia:
K, I think I get your point here. @iambmelt Do you think for the case where json parse exception thrown, we should not store these key value. I think it would be better to not store instead of (key, NULL) in this case.

iambmelt:
@heidijinxujia Yeah, I'd skip them if they fail to parse. Hopefully this decision is theoretical and this isn't a code path we're walking down too often, as it would indicate a larger problem

heidijinxujia:
@iambmelt  Addressed. Thanks!

adal/src/androidTest/java/com/microsoft/aad/adal/HashMapExtensionTests.java
```diff
@@ -79,10 +79,6 @@ public void testURLFormDecodeNegative() throws IllegalArgumentException,
         assertNotNull(result);
         assertTrue(result.isEmpty());
 
-        result = (HashMap<String, String>) m.invoke(object, "&a=");
```
addev-ashish: can you add some values here that will test the new path? also if a test case can trigger the exception that would be great.

heidijinxujia: K. I will add more unit tests today.

Issue #1087, opened 2018-01-05
Update userappwithbroker for guest account support


addev-ashish:
Add the guest client id and authority. Also store  uid per authority

PR #1088, opened 2018-01-05
Updating userappwtihbroker for support of guest accounts #1087

addev-ashish:
Add authority as a part of UID key
Add constants for guest account

### DIFF
--- a/userappwithbroker/src/main/java/com/microsoft/aad/adal/example/userappwithbroker/Constants.java
+++ b/userappwithbroker/src/main/java/com/microsoft/aad/adal/example/userappwithbroker/Constants.java
@@ -29,6 +29,7 @@ package com.microsoft.aad.adal.example.userappwithbroker;
 
 public class Constants {
     enum AuthorityType {
+        AAD_GUEST("https://login.microsoftonline.com/nomfaad.onmicrosoft.com"),
         AAD_MSDEVEX("https://login.microsoftonline.com/msdevex.onmicrosoft.com"),
         AAD_COMMON("https://login.microsoftonline.com/common");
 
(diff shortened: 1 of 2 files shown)


userappwithbroker/src/main/java/com/microsoft/aad/adal/example/userappwithbroker/Constants.java
```diff
@@ -29,6 +29,7 @@
 
 public class Constants {
     enum AuthorityType {
+        AAD_GUEST("https://login.microsoftonline.com/nomfaad.onmicrosoft.com"),
```
heidijinxujia: nit: It could be more convenient to put the AAD_MSDEVEX as default at the top of the list, in the perspective of manual testing. So we do not need to scroll down each time.

userappwithbroker/src/main/java/com/microsoft/aad/adal/example/userappwithbroker/Constants.java
```diff
@@ -58,6 +59,7 @@ public String getText() {
 
 
     enum ClientId {
+        GUESTCLIENT("ea5c8087-2476-489c-ae03-ad44a2ac399d"),
         MSDEVEX("b92e0ba5-f86e-4411-8e18-6b5f928d968a"),
```
heidijinxujia: nit: Same here. It could be better to set MSDEVEX as default.

userappwithbroker/src/main/java/com/microsoft/aad/adal/example/userappwithbroker/Constants.java
iambmelt: This is a good point, I'd be happy to see this change

addev-ashish: fixed

userappwithbroker/src/main/java/com/microsoft/aad/adal/example/userappwithbroker/Constants.java
addev-ashish: fixed

PR #1089, opened 2018-01-05
Sanity check on the GDPR loggers

heidijinxujia:


adal/src/main/java/com/microsoft/aad/adal/DateTimeAdapter.java
```diff
@@ -121,7 +121,7 @@ public synchronized Date deserialize(JsonElement json, Type typeOfT,
         try {
             return mEnUs24HourFormat.parse(jsonString);
         } catch (final ParseException e) {
-            Logger.e(TAG + methodName, "Could not parse date: " + e.getMessage(), "",
+            Logger.e(TAG + methodName, "Could not parse date. ", e.getMessage(),
```
iambmelt: As a small nit, you could remove the whitespace character after the `.` -- I'm not particularly troubled by this, however.

Issue #1101, opened 2018-01-22
Silent Auth for Sovereign Cloud Account is broken in the Sample App


heidijinxujia:
RFC test case 345163

PR #1103, opened 2018-01-23
Update the sample app to support the silent auth of sov cloud account

heidijinxujia:
RFC #1101 
For the client app, to support the sov cloud accounts, it should maintain the authority returned from AuthenticationResult and use it for the future silent auth purpose.

In the sample app, it demonstrate an example where we use the shared preference to keep the cloud authorities.

### DIFF
--- a/userappwithbroker/src/main/java/com/microsoft/aad/adal/example/userappwithbroker/MainActivity.java
+++ b/userappwithbroker/src/main/java/com/microsoft/aad/adal/example/userappwithbroker/MainActivity.java
@@ -48,6 +48,7 @@ import com.microsoft.aad.adal.AuthenticationContext;
 import com.microsoft.aad.adal.AuthenticationResult;
 import com.microsoft.aad.adal.AuthenticationSettings;
 import com.microsoft.aad.adal.IDispatcher;
+import com.microsoft.aad.adal.Logger;
 import com.microsoft.aad.adal.PromptBehavior;
 import com.microsoft.aad.adal.Telemetry;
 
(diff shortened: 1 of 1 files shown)


userappwithbroker/src/main/java/com/microsoft/aad/adal/example/userappwithbroker/MainActivity.java
```diff
@@ -292,23 +296,28 @@ public void onError(Exception exc) {
      */
     private void saveUserIdFromAuthenticationResult(final AuthenticationResult authResult, final String authority) {
         mSharedPreference = getSharedPreferences(SHARED_PREFERENCE_STORE_USER_UNIQUEID, MODE_PRIVATE);
+        if (null != authResult) {
```
iambmelt: I don't necessarily have a problem with defensively checking if this parameter is `null`, but I will call out that if `onSuccess()` was called with a `null` result, we wouldn't make it this far: it would throw an NPE inside of `onSuccess()` when we log out the accessToken. You could omit this check, since this code should never get hit.

heidijinxujia: Thank you, Brian. Yes, I understand. I was thinking it would be better to check the input object for future usage because it is a single method (I understand it is private method and currently just get called in the onSuccess which has null check on result) and has getAuthority and getUserInfo called on the object. If the other calling app does not has the null check on the input, nullpointerexception will be thrown without the checking.

userappwithbroker/src/main/java/com/microsoft/aad/adal/example/userappwithbroker/MainActivity.java
```diff
@@ -292,23 +296,28 @@ public void onError(Exception exc) {
...
+            final SharedPreferences.Editor prefEditor = mSharedPreference.edit();
+            if (null != authResult.getAuthority()) {
```
addev-ashish: so if authority comes back as null, shouldn't you raise hell here saying this should not happen? Just add a Toast

heidijinxujia: Addressed.

PR #1124, opened 2018-02-20
Fix all checkstyle errors

sgonzalezMSFT:


heidijinxujia:
Changes LGTM. @iambmelt good advice! I think it would be better to open another PR for the final declaration change.

PR #1125, opened 2018-02-20
the log related part of ICS 337720

piqiums:
the log related part of ICS 337720

### DIFF
--- a/adal/src/main/java/com/microsoft/aad/adal/Logger.java
+++ b/adal/src/main/java/com/microsoft/aad/adal/Logger.java
@@ -22,6 +22,7 @@
 // THE SOFTWARE.
 
 package com.microsoft.aad.adal;
+import com.microsoft.aad.adal.BuildConfig;
 
 import android.annotation.SuppressLint;
 import android.util.Log;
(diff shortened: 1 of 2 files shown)


adal/src/main/java/com/microsoft/aad/adal/Logger.java
```diff
@@ -44,7 +45,7 @@
     private LogLevel mLogLevel  = LogLevel.Verbose;
     private ILogger mExternalLogger = null;
     private static final String CUSTOM_LOG_ERROR = "Custom log failed to log message:%s";
-    private boolean mAndroidLogEnabled = false;
+    private boolean mAndroidLogEnabled = BuildConfig.DEBUG;
```
iambmelt: This will leave logging on by 'default' for `debug` builds and allow it to still be settable for `release` builds, am I understanding this change correctly @piqiums ?

piqiums: You're correct. It will be ON for debug and OFF for release by default. yes, it's settable via the method 'Logger.setAndroidLogEnabled'. But I think we should remove this method.

iambmelt: I'd be curious to hear the rationale behind removing it _entirely_ -- I think it would severely hamper our ability to inspect device logs and drill into the verbose errors not captured by our telemetry.

Thanks for clarifying on the above question -- these changes look OK to me. I give the :shipit: once Travis-CI passes

Thanks, @piqiums

PR #1134, opened 2018-03-05
Run findbugs task after assembleDebug

iambmelt:
Ensures `classes` directory isn't empty when `findbugs` task is executed. Runs _after_ `assembleDebug` or `assembleRelease`


iambmelt:
Thanks for reviewing @shoatman -- I'll let CI finish before merging into your PR 👍

PR #1154, opened 2018-03-22
Add the adapter layer to Common-Core logging

heidijinxujia:
- Dev can enable/disable the Common-Core logging by calling setEnableCommonCoreLog(). If the common core logging is enabled, where mEnableCommonCoreLog == true, the adal logging will apply the common-core logging logic. Otherwise, adal logging will follow the native logging pattern.

- Add the commonCoreWrapper() function. In ADAL, we put all the PII related infos into the parameter additionalMessage. If the additionalMessage is not null, and he CommonCore logging is enabled, we will treat the additionaMessage apart from the message and call the PII logging separately.

Here is the mapping table.

| ADAL  | CommonCoreLogging |
| ------------- | ------------- |
| `Logger.v` | `CommonCoreLogger.verbose` + `CommonCoreLogger.verbosePII` if additionalMessage not empty  | 
| `Logger.i` | `CommonCoreLogger.info` + `CommonCoreLogger.infoPII` if additionalMessage not empty  |
| `Logger.w` | `CommonCoreLogger.warn` + `CommonCoreLogger.warnPII` if additionalMessage not empty |
| `Logger.e` | `CommonCoreLogger.error` + `CommonCoreLogger.warnPII` if additionalMessage not empty |
| `Logger.d` | `CommonCoreLogger.verbose` |

Conflicts need to discuss
1. LogLevel debug is deprecated in ADAL. So I did not introduce the debug level into Common Core logging. For all the Logger.d calling, I map it into CommonCoreLogger.verbose. Sounds good?
2. In ADAL logging logic, we did not print the Throwable stacktrace unless the PII logging is enable. However, in MSAL we print the stacktrace anyway. Due to this conflict, my current implementation in the common-core logging is aligned to the MSAL logging logic. However, to the ADAL adapter layer, we only print the stacktrace when the PII logging is enable.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,3 +1,4 @@
 [submodule "common"]
 	path = common
 	url = https://github.com/AzureAD/microsoft-authentication-library-common-for-android.git
+	branch = jinjia/logging
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,4 +1,3 @@
 [submodule "common"]
 	path = common
 	url = https://github.com/AzureAD/microsoft-authentication-library-common-for-android.git
-	branch = jinjia/logging
--- a/adal/src/main/java/com/microsoft/aad/adal/Logger.java
+++ b/adal/src/main/java/com/microsoft/aad/adal/Logger.java
@@ -26,7 +26,6 @@ package com.microsoft.aad.adal;
 import android.util.Log;
 
 import com.microsoft.identity.common.adal.internal.util.StringExtensions;
-import com.microsoft.identity.common.internal.logging.CommonCoreLogger;
 import com.microsoft.identity.common.internal.logging.ILoggerCallback;
 import com.microsoft.identity.common.internal.logging.LoggerSettings;
 
@@ -46,39 +45,42 @@ public class Logger {
 
     /**
      * @return The single instance of {@link Logger}.
+     *
+     * @deprecated use {@link com.microsoft.identity.common.internal.logging.Logger#getInstance()} instead.
      */
+    @Deprecated
     public static Logger getInstance() {
         return sINSTANCE;
     }
 
-
     /**
      * Set the log level for diagnostic purpose. By default, the sdk enables the verbose level logging.
-     *
+     * @deprecated use {@link com.microsoft.identity.common.internal.logging.Logger#setLogLevel(com.microsoft.identity.common.internal.logging.Logger.LogLevel)} instead.
      * @param logLevel The {@link LogLevel} to be enabled for the diagnostic logging.
      */
+    @Deprecated
     public void setLogLevel(final LogLevel logLevel) {
         switch (logLevel) {
             case Error:
-                CommonCoreLogger.getInstance()
-                        .setLogLevel(CommonCoreLogger.LogLevel.ERROR);
+                com.microsoft.identity.common.internal.logging.Logger.getInstance()
+                        .setLogLevel(com.microsoft.identity.common.internal.logging.Logger.LogLevel.ERROR);
                 break;
             case Warn:
-                CommonCoreLogger.getInstance()
-                        .setLogLevel(CommonCoreLogger.LogLevel.WARN);
+                com.microsoft.identity.common.internal.logging.Logger.getInstance()
+                        .setLogLevel(com.microsoft.identity.common.internal.logging.Logger.LogLevel.WARN);
                 break;
             case Info:
-                CommonCoreLogger.getInstance()
-                        .setLogLevel(CommonCoreLogger.LogLevel.INFO);
+                com.microsoft.identity.common.internal.logging.Logger.getInstance()
+                        .setLogLevel(com.microsoft.identity.common.internal.logging.Logger.LogLevel.INFO);
                 break;
             case Verbose:
-                CommonCoreLogger.getInstance()
-                        .setLogLevel(CommonCoreLogger.LogLevel.VERBOSE);
+                com.microsoft.identity.common.internal.logging.Logger.getInstance()
+                        .setLogLevel(com.microsoft.identity.common.internal.logging.Logger.LogLevel.VERBOSE);
                 break;
             case Debug:
                 //The debug level is deprecated and removed in common core.
-                CommonCoreLogger.getInstance()
-                        .setLogLevel(CommonCoreLogger.LogLevel.INFO);
+                com.microsoft.identity.common.internal.logging.Logger.getInstance()
+                        .setLogLevel(com.microsoft.identity.common.internal.logging.Logger.LogLevel.INFO);
                 break;
             default:
                 throw new IllegalArgumentException("Unknown logLevel");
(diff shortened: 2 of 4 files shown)


heidijinxujia:
The TravisCI fails because it could not update with the submodule.

heidijinxujia:
I also finished the E2E testing for `setEnableCommonCoreLog()`.

piqiums:
for the 1st question:
```
LogLevel debug is deprecated in ADAL. So I did not introduce the debug level into Common Core logging. For all the Logger.d calling, I map it into CommonCoreLogger.verbose. Sounds good?
```
I suggest map the Logger.d to Info level since the log of 'd' level is very useful, better map it to a higher level instead of a lower level.

.gitmodules
```diff
@@ -1,3 +1,4 @@
 [submodule "common"]
 	path = common
 	url = https://github.com/AzureAD/microsoft-authentication-library-common-for-android.git
+	branch = jinjia/logging
```
addev-ashish: don't make this a part of the PR

adal/src/main/java/com/microsoft/aad/adal/Logger.java
```diff
@@ -27,6 +27,7 @@
 import android.util.Log;
 
 import com.microsoft.identity.common.adal.internal.util.StringExtensions;
+import com.microsoft.identity.common.internal.logging.*;
```
addev-ashish: nit: *

adal/src/main/java/com/microsoft/aad/adal/Logger.java
```diff
@@ -234,6 +287,77 @@ private void log(String tag, String message, String additionalMessage, LogLevel
         }
     }
 
+    private void commonCoreWrapper(String tag, String message, String additionalMessage, LogLevel logLevel,
+                                   ADALError errorCode, Throwable throwable){
+        if (mEnableCommonCoreLog) {
```
addev-ashish: if you switch this check the code will be simpler

adal/src/main/java/com/microsoft/aad/adal/Logger.java
```diff
@@ -234,6 +287,77 @@ private void log(String tag, String message, String additionalMessage, LogLevel
         }
     }
 
+    private void commonCoreWrapper(String tag, String message, String additionalMessage, LogLevel logLevel,
```
addev-ashish: lets talk about this tomorrow, i am confused why ADAL is calling into the common core

iambmelt:
As part of the changes to introduce a common `Logger` can we mark the ADAL/MSAL specific classes as `@Deprecated`?

Don't see any sense in having duplicative APIs _long term_

adal/src/main/java/com/microsoft/aad/adal/Logger.java
```diff
@@ -40,69 +38,125 @@
...
-    // Disable to log PII by default.
-    private boolean mEnablePII = false;
+    private String mCorrelationId = null;
 
     /**
      * @return The single instance of {@link Logger}.
+     *
+     * @deprecated use {@link com.microsoft.identity.common.internal.logging.Logger#getInstance()} instead.
      */
+    @Deprecated
```
addev-ashish: why is this deprecated? How will the apps get the instance of the logger? apps should not call into common core

adal/src/main/java/com/microsoft/aad/adal/Logger.java
```diff
@@ -40,69 +38,125 @@
...
         return sINSTANCE;
     }
 
     /**
      * Set the log level for diagnostic purpose. By default, the sdk enables the verbose level logging.
-     *
+     * @deprecated use {@link com.microsoft.identity.common.internal.logging.Logger#setLogLevel(com.microsoft.identity.common.internal.logging.Logger.LogLevel)} instead.
      * @param logLevel The {@link LogLevel} to be enabled for the diagnostic logging.
      */
+    @Deprecated
```
addev-ashish: same question here

addev-ashish: if you are expecting apps to call into common core, that is not an acceptable solution

adal/src/main/java/com/microsoft/aad/adal/Logger.java
```diff
@@ -40,69 +38,125 @@
...
     /**
      * Set the custom logger. Configures external logging to configure a callback that
      * the sdk will use to pass each log message. Overriding the logger callback is not allowed.
      *
+     * @deprecated use {@link com.microsoft.identity.common.internal.logging.Logger#setExternalLogger(ILoggerCallback)} instead.
+     *
      * @param externalLogger The reference to the {@link ILogger} that can
      *                       output the logs to the designated places.
      */
+    @Deprecated
```
addev-ashish: should not be deprecated.

iambmelt:
+1 on the _non-deprecation_ of `setLogLevel()`, `getInstance()`, `setAndroidLogEnabled()`, `setEnablePII()`, `enum LogLevel`, and `setExternalLogger()` -- marking other methods such as ... `d()`, `i()`, `v()`, `w()`, and `e()` as `@Deprecated` makes sense to me... [1]

I don't know what to do about `ILogger` exactly.... I think it's a poor API but this may be a problem we can solve later... Thoughts @addev-ashish @heidijinxujia ?

<hr>

**[1]** We don't expect 3rd parties to use these APIs for logging, right? If no, then I don't see any reason _not_ to `@Deprecate` these methods. Since ADAL/MSAL will depend on `common` for the foreseeable future, it makes sense to me that they should _fully delegate_ to the shared logging layer and, over time, migrate off of their own logging APIs entirely (they will still expose their own logging hooks, but there won't be any duplicative implementation of logging functions such as `i()`, `d()`, etc...).

/cc @addev-ashish

heidijinxujia:
@addev-ashish @iambmelt Comments addressed, please have a look. Thanks.

adal/src/main/java/com/microsoft/aad/adal/Logger.java
```diff
@@ -65,7 +54,31 @@ public static Logger getInstance() {
      * @param logLevel The {@link LogLevel} to be enabled for the diagnostic logging.
      */
     public void setLogLevel(final LogLevel logLevel) {
-        this.mLogLevel = logLevel;
+        switch (logLevel) {
```
iambmelt: To lower the [cyclomatic complexity](https://en.wikipedia.org/wiki/Cyclomatic_complexity) of this function, this `switch` statement as well as the one used @ `L99` looks like a good use case for an [`EnumBiMap`](https://google.github.io/guava/releases/18.0/api/docs/com/google/common/collect/EnumBiMap.html) -- since we can't use this structure without adding the `Guava` dependency you _could_ cut down on the code here by initializing a `Map<Enum, Enum>` such that...

```java
Map<Enum, Enum> mLogLevelMap = new HashMap<>() {{
    put(LogLevel.Error, com.microsoft.identity.common.internal.logging.Logger.LogLevel.ERROR);
    put(com.microsoft.identity.common.internal.logging.Logger.LogLevel.ERROR, LogLevel.Error);
    // ... (do this for each LogLevel pairing)
}};

// Now the function @ L57 becomes
public void setLogLevel(final LogLevel logLevel) {
    // Avoid the switch block by calling get() on the Map
    com.microsoft.identity.common.internal.logging.Logger.getInstance()
        .setLogLevel(mLogLevelMap.get(logLevel));
}

// And the function @ L91 becomes
public synchronized void setExternalLogger(ILogger externalLogger) {
    com.microsoft.identity.common.internal.logging.Logger.getInstance().setExternalLogger(new ILoggerCallback() {
        @Override
        public void log(String tag, com.microsoft.identity.common.internal.logging.Logger.LogLevel logLevel, String message, boolean containsPII) {
            if (mExternalLogger != null) {
                if (!LoggerSettings.getAllowPii() && containsPII) {
                    return;
                } else {
                    // Avoid the switch block by calling get() on the Map
                    mExternalLogger.Log(tag, message, null, mLogLevelMap.get(logLevel), null);
                }
            }
        }
    });

    mExternalLogger = externalLogger;
}
```

(As a word of caution, I haven't tested this code)

heidijinxujia: Sorry, I took it back. I reconsidered this implementation. Actually the mapping from adal loglevel to common-core loglevel is only used once in setLogLevel. On the other hand, the mapping from common-core loglevel to adal loglevel also only happen once in setExternalLogger. I do not think it would be a good idea to hold the mixed mapping between common-core and adal.

adal/src/main/java/com/microsoft/aad/adal/Logger.java
```diff
@@ -152,204 +192,225 @@ void Log(String tag, String message, String additionalMessage, LogLevel level,
...
-                // log message as warning to report callback error issue
-                Log.w(tag, String.format(CUSTOM_LOG_ERROR, message));
-            }
+                throw new IllegalArgumentException("Unknown logLevel");
         }
     }
 
     /**
-     * Logs debug message.
+     * @deprecated use {@link com.microsoft.identity.common.internal.logging.Logger#info(String, String, String)} instead.
```
iambmelt: @addev-ashish Do we want to hedge this with some kind of messaging that external devs shouldn't use these functions at all? It's implicit in our design, but not explicitly communicated anywhere 🤔

iambmelt: I'm not sure what the prevailing attitude is right now towards the removal of public API surface. If we're OK with a major-rev then I say sure, let's yank the visibility. If we need to continue to expose these methods, let's deprecate.

@addev-ashish Can you lend some guidance here? Can we be bold and start trimming surface?

PR #1156, opened 2018-03-23
merge unity, part_3

piqiums:
merge unity, part 3.
In this PR, I made AuthenticationRequest a public class, and all its constructors public, since it's used by ad-accounts, and there's a class extending it.  consumer is : ad-accounts/AADAuthenticator/src/main/java/com/microsoft/workaccount/workplacejoin/BrokerContext.java:676.

### DIFF
--- a/adal/src/main/java/com/microsoft/aad/adal/AuthenticationRequest.java
+++ b/adal/src/main/java/com/microsoft/aad/adal/AuthenticationRequest.java
@@ -31,7 +31,7 @@ import java.util.UUID;
 /**
  * Represent request and keeps authorization code and similar info.
  */
-class AuthenticationRequest implements Serializable {
+public class AuthenticationRequest implements Serializable {
 
     private static final long serialVersionUID = 1L;
 
@@ -75,6 +75,8 @@ class AuthenticationRequest implements Serializable {
 
     private transient InstanceDiscoveryMetadata mInstanceDiscoveryMetadata;
 
+    private boolean mSkipCache = false;
+
     /**
      * Developer can use acquireToken(with loginhint) or acquireTokenSilent(with
      * userid), so this sets the type of the request.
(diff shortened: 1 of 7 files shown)


adal/src/main/java/com/microsoft/aad/adal/AuthenticationRequest.java
```diff
@@ -31,7 +31,7 @@
 /**
  * Represent request and keeps authorization code and similar info.
  */
-class AuthenticationRequest implements Serializable {
+public class AuthenticationRequest implements Serializable {
```
piqiums: why make AuthenticationRequest and its constructors public ? since it's used by ad-accounts, and there's a class extending it.  consumer is : ad-accounts/AADAuthenticator/src/main/java/com/microsoft/workaccount/workplacejoin/BrokerContext.java:676

adal/src/main/java/com/microsoft/aad/adal/WebRequestHandler.java
```diff
@@ -54,6 +54,10 @@
 
     private UUID mRequestCorrelationId = null;
 
+    private String mCurrentAdalClientVersion = AuthenticationContext.getVersionName();
+
+    private String mBrokerVersion = null;
```
piqiums: this is a variable I added to make it both work for broker and public ADAL. In broker, we need to add the broker version as a parameter in header. But it's not necessary in public ADAL, so I introduced this variable, if it's not set (in public ADAL) then the broker version won't be passed as a parameter in header.

adal/src/main/java/com/microsoft/aad/adal/AuthenticationRequest.java
addev-ashish: This must not be public. Leave it in package scope and we can work around it in BrokerContext

adal/src/main/java/com/microsoft/aad/adal/AuthenticationRequest.java
```diff
@@ -104,7 +106,21 @@
         mClaimsChallenge = claimsChallenge;
     }
 
-    AuthenticationRequest(String authority, String resource, String client, String redirect,
+    public AuthenticationRequest(String authority, String resource, String client, String redirect,
```
addev-ashish: can we see if we can prevent this explosion of constructors?

adal/src/main/java/com/microsoft/aad/adal/AuthenticationRequest.java
piqiums: changed it back.

adal/src/main/java/com/microsoft/aad/adal/AuthenticationRequest.java
piqiums: since we need to make some change in the ad-accounts, I would keep the constructors in public-ADAL.

addev-ashish:
Please merge after Travis passes.

PR #1157, opened 2018-03-23
exception handling

addev-ashish:
Doing exception handling for unified core

### DIFF
--- a/adal/src/main/java/com/microsoft/aad/adal/ADALError.java
+++ b/adal/src/main/java/com/microsoft/aad/adal/ADALError.java
@@ -28,8 +28,7 @@ import android.content.res.Configuration;
 import android.content.res.Resources;
 
 import com.microsoft.identity.common.exception.CommonCoreBaseException;
-import com.microsoft.identity.common.exception.CommonCoreException;
-import com.microsoft.identity.common.exception.CommonCoreServiceException;
+import com.microsoft.identity.common.exception.CommonCoreExceptionMessage;
 
 import java.util.HashMap;
 import java.util.Map;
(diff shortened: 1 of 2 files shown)


adal/src/main/java/com/microsoft/aad/adal/ADALError.java
```diff
@@ -638,6 +634,38 @@ public String getLocalizedDescription(Context context) {
...
+        exceptionMap.put(CommonCoreServiceException.ACCESS_DENIED, ADALError.AUTH_REFRESH_FAILED_PROMPT_NOT_ALLOWED);
+        exceptionMap.put(CommonCoreServiceException.INVALID_INSTANCE, ADALError.AUTH_REFRESH_FAILED_PROMPT_NOT_ALLOWED);
+        exceptionMap.put(CommonCoreServiceException.INVALID_REQUEST, ADALError.SERVER_INVALID_REQUEST);
+        exceptionMap.put(CommonCoreServiceException.REQUEST_TIMEOUT, ADALError.SERVER_ERROR);
+        exceptionMap.put(CommonCoreServiceException.SERVICE_NOT_AVAILABLE, ADALError.SERVER_ERROR);
+        exceptionMap.put(CommonCoreServiceException.UNKNOWN_ERROR, ADALError.SERVER_ERROR);
+    }
+
+    public static ADALError fromCommon(final CommonCoreBaseException exception) {
+        // What do we do if errorCode is null?
```
heidijinxujia: ADAL did not use AuthenticationException with null errorCode, although it does not have a null check in the constructor.  I think the question should be in which circumstance we will also `null` erroCode in CommonCoreBaseException or we should not. 

There is another question. Which if the map does not contains the key-value pair for CommonCoreBaseException.getErrorCode() mapping to the ADALError? In the current implementation, we need to add the mapping when we introduce a new errorCode in common-core. And it is hard to find the mapping miss until the CommonCoreBaseException is thrown. As a result, we probably return the AuthenticationException with null ADAL ErrorCode. Should we also take into consideration of this case?

adal/src/main/java/com/microsoft/aad/adal/ADALError.java
```diff
@@ -638,6 +634,38 @@ public String getLocalizedDescription(Context context) {
...
+        return exceptionMap.get(exception.getErrorCode());
```
heidijinxujia: Other concern is how will ADAL consume the CommonCoreServiceException error code description. As CommonCoreServiceException errors have a more refined categorization. Will this mapping miss the common core error descriptions? 

For example, how do we distinguish the errors between `CommonCoreServiceException.REQUEST_TIMEOUT`, `CommonCoreServiceException.SERVICE_NOT_AVAILABLE` and `CommonCoreServiceException.UNKNOWN_ERROR` when they are all mapped into `ADALError.SERVER_ERROR`? Will the wrapped AuthenticationException lose this part of info?

adal/src/main/java/com/microsoft/aad/adal/ADALError.java
```diff
@@ -638,6 +634,38 @@ public String getLocalizedDescription(Context context) {
             return resources.getString(resources.getIdentifier(this.name(), "string",
                     context.getPackageName()));
         }
-        return mCommonADALError.getDescription();
+        return mDescription;
+    }
+    private static final Map<String, ADALError> exceptionMap = new HashMap<String,ADALError>();
```
iambmelt: nit: newline between previous method and this one

adal/src/main/java/com/microsoft/aad/adal/ADALError.java
iambmelt: Possibly we create a new generic 'stuff went wrong but didn't report an error code' kind of exception? Something to handle the default case if the `exceptionMap` returns `null` on `get()` -- seems like the most sane thing I can think of, short of throwing a `RuntimeException` to indicate it's a programming error

adal/src/main/java/com/microsoft/aad/adal/ADALError.java
iambmelt: Can we create new `ADALError` to support this case? I don't want us to be digging-in deeper for an API we're not particularly happy with, but I don't know how else we can maintain the specificity _and_ work within the constraint of having to shoehorn whatever error `common` emits into a limited set of `enums`

adal/src/main/java/com/microsoft/aad/adal/ADALError.java
```diff
@@ -36,577 +42,567 @@
     /**
      * Authority validation returned an error.
      */
-    DEVELOPER_AUTHORITY_CAN_NOT_BE_VALIDED(com.microsoft.identity.common.adal.error.ADALError.DEVELOPER_AUTHORITY_CAN_NOT_BE_VALIDED),
+    DEVELOPER_AUTHORITY_CAN_NOT_BE_VALIDED("Authority validation returned an error"),
```
iambmelt: @shoatman Was the developer who had originally migrated these errors into `common` -- I suspect he'll also want to review these changes once he returns.

adal/src/main/java/com/microsoft/aad/adal/ADALError.java
iambmelt: The more I think about it, the more I'm convinced we need to have _something_ happen when `null` is returned from `get()` -- it would be too easy to add a new `CommonCoreServiceException` and fail to update this mapping.

What if we [provided a setter on this enum](https://stackoverflow.com/questions/15302188/are-enums-allowed-to-have-setters-in-java) to track the `String` `errorCode` returned from `common` -- would we still have to maintain this mapping?

iambmelt:
Dropped a couple of inline questions - no big concerns. Think we need a fallback for when `exceptionMap.get()` returns `null`

PR #1159, opened 2018-03-26
Updating reference docs to be clear cache is skipped

danieldobalian:


adal/src/main/java/com/microsoft/aad/adal/AuthenticationContext.java
```diff
@@ -437,11 +437,9 @@ public void acquireToken(Activity activity, String resource, String clientId,
     }
 
     /**
-     * acquireToken will start interactive flow if needed. It checks the cache
-     * to return existing result if not expired. It tries to use refresh token
-     * if available. If it fails to get token with refresh token, behavior will
-     * depend on options. If promptbehavior is AUTO, it will remove this refresh
-     * token from cache and fall back on the UI if activitycontext is not null.
+     * acquireToken will start an interactive auth flow to acquire new tokens 
+     * with the requested claims. If promptbehavior is AUTO, it will remove this refresh
```
iambmelt: This actually seems _less_ clear to me - if AT is called we:

1. [Call acquireToken](https://github.com/AzureAD/azure-activedirectory-library-for-android/blob/477ff358e1cc40587a8a71344650990288defad9/adal/src/main/java/com/microsoft/aad/adal/AuthenticationContext.java#L477)
2. Which calls [performAcquireTokenRequest](https://github.com/AzureAD/azure-activedirectory-library-for-android/blob/477ff358e1cc40587a8a71344650990288defad9/adal/src/main/java/com/microsoft/aad/adal/AcquireTokenRequest.java#L111)
3. Which calls [tryAcquireTokenSilent](https://github.com/AzureAD/azure-activedirectory-library-for-android/blob/477ff358e1cc40587a8a71344650990288defad9/adal/src/main/java/com/microsoft/aad/adal/AcquireTokenRequest.java#L333)
4. Which calls [shouldTrySilentFlow](https://github.com/AzureAD/azure-activedirectory-library-for-android/blob/477ff358e1cc40587a8a71344650990288defad9/adal/src/main/java/com/microsoft/aad/adal/AcquireTokenRequest.java#L352)
5. Which returns `true` [if no claims challenge is present `&& prompt == AUTO`](https://github.com/AzureAD/azure-activedirectory-library-for-android/blob/477ff358e1cc40587a8a71344650990288defad9/adal/src/main/java/com/microsoft/aad/adal/AcquireTokenRequest.java#L388)

What's going on here?

Without digging around in the code exhaustively, I don't remember anything about this removing RT's (unless they've expired/been revoked and cannot resolve the silent call) -- won't we need `FORCE` for this behavior? /cc @addev-ashish

heidijinxujia: ADAL won't remove the tokens unless getting "invalid_grant" from the server.

heidijinxujia: The old javadoc does not see correct to me. If AT/RT exists and we still need the UI prompt. we should call AT with prompt behavior with `FORCE_PROMPT`.

heidijinxujia: If it is a claim challenge call, we do not need `FORCE_PROMPT` behavior. The AT call will skip the silent call and jump into the acquireTokenInteractiveFlow.

danieldobalian: Thanks @iambmelt @heidijinxujia, I'll update the text. I had left the auto part from the first note, but it seems like that piece was also misleading.

danieldobalian: Made a few changes. The promptbehavior option in this specific overload does have an unclear use. We should clarify and add it to the description of prompt parameter before merging this PR.

adal/src/main/java/com/microsoft/aad/adal/AuthenticationContext.java
```diff
@@ -437,12 +437,8 @@ public void acquireToken(Activity activity, String resource, String clientId,
...
 
     /**
-     * acquireToken will start interactive flow if needed. It checks the cache
-     * to return existing result if not expired. It tries to use refresh token
-     * if available. If it fails to get token with refresh token, behavior will
-     * depend on options. If promptbehavior is AUTO, it will remove this refresh
-     * token from cache and fall back on the UI if activitycontext is not null.
-     * Default is AUTO.
+     * acquireToken will start an interactive auth flow to acquire new tokens 
+     * with the requested claims. Bypasses token cache.
```
iambmelt: Was there some specific ask to change this documentation? These calls will only bypass the token cache if `PromptBehavior != Auto` _or_ claims challenges are present.

[Here ](https://github.com/AzureAD/azure-activedirectory-library-for-android/blob/dev/adal/src/main/java/com/microsoft/aad/adal/AcquireTokenRequest.java#L333-L341)is where this call uses the local cache to try and acquireTokens -- [here ](https://github.com/AzureAD/azure-activedirectory-library-for-android/blob/dev/adal/src/main/java/com/microsoft/aad/adal/AcquireTokenRequest.java#L386-L390)is where claims challenges and `PrompBehavior` are evaluated

danieldobalian: Good catch. The code here may be implemented a bit differently than the other platforms and planned behavior ([ADAL .NET](https://github.com/AzureAD/azure-activedirectory-library-for-dotnet/blob/dev/adal/src/Microsoft.IdentityModel.Clients.ActiveDirectory/AuthenticationContext.cs#L391)).  I'm not sure we need to change it, but it does seem confusing to have split behavior in a call intended for getting a token with a superset of claims. IMO, claims in this overload should be non-optional.

danieldobalian: Submitted a new commit with correct behavior. The change is coming from some confusion/ambiguity around this method.

PR #1160, opened 2018-03-28
merge unity, part 4

piqiums:
Merge unity, part 4

### DIFF
--- a/adal/src/androidTest/java/com/microsoft/aad/adal/AuthenticationActivityUnitTest.java
+++ b/adal/src/androidTest/java/com/microsoft/aad/adal/AuthenticationActivityUnitTest.java
@@ -87,6 +87,8 @@ public class AuthenticationActivityUnitTest {
 
     private static final int TEST_CALLING_UID = 333;
 
+    private static final String TEST_AUTHORITY = "https://login.microsoftonline.com/common";
+
     private Intent mIntentToStartActivity;
 
     @SuppressWarnings("checkstyle:visibilitymodifier")
@@ -126,7 +128,7 @@ public class AuthenticationActivityUnitTest {
         Constructor<?> constructor = c.getDeclaredConstructor(String.class, String.class,
                 String.class, String.class, String.class, boolean.class);
         constructor.setAccessible(true);
-        Object o = constructor.newInstance("authority", "client", "resource", "redirect",
+        Object o = constructor.newInstance(TEST_AUTHORITY, "client", "resource", "redirect",
                 "loginhint", false);
         ReflectionUtils.setFieldValue(o, "mRequestId", TEST_REQUEST_ID);
         ReflectionUtils.setFieldValue(o, "mCorrelationId", UUID.randomUUID());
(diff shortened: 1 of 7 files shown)


adal/src/main/java/com/microsoft/aad/adal/Oauth2.java
```diff
@@ -767,7 +834,38 @@ private void stopHttpEvent(final HttpEvent httpEvent) {
...
-        mTokenEndpoint = tokenEndpoint;
+    private void setTokenEndpoint(final String authority) {
+        try {
+            mTokenEndpoint = getUpdatePreferredNetworkLocation(authority) + DEFAULT_TOKEN_ENDPOINT;
+        } catch(AuthenticationException ex) {
+            mTokenEndpoint = authority + DEFAULT_TOKEN_ENDPOINT;
+        }
+    }
+
+    private String getUpdatePreferredNetworkLocation(final String authority) throws AuthenticationException {
```
iambmelt: See `AcquireTokenRequest.java` @ `L259` -- I suggest refactoring for reuse where possible

adal/src/main/java/com/microsoft/aad/adal/Oauth2.java
```diff
@@ -351,6 +389,23 @@ private static void extractJsonObjects(Map<String, String> responseItems, String
         }
     }
 
+    static void extractJsonArray(Map<String, List<String>> responseList, String jsonStr)
```
iambmelt: `HashMapExtensions.java` contains _very_ similar functionality to this already. I'd take a look to see if the current implementation does everything you need already. May require some tweaking of the API

piqiums: good catch. thanks

adal/src/androidTest/java/com/microsoft/aad/adal/OauthTests.java
```diff
@@ -350,21 +350,17 @@ public void testRefreshTokenRequestMessage() throws UnsupportedEncodingException
                 "client 1234567890-+=;'", "redirect 1234567890-+=;'", "loginhint@ggg.com", null,
                 null, null, false);
         final Oauth2 oauth2 = createOAuthInstance(request);
-        assertEquals(
-                "Token request",
-                "grant_type=refresh_token&refresh_token=refreshToken23434%3D&client_id=client+1234567890-%2B%3D%3B%27&resource=resource%2520+",
-                oauth2.buildRefreshTokenRequestMessage("refreshToken23434="));
+        assertTrue("Token request",
```
heidijinxujia: Have you tried to run the unit tests?

adal/src/androidTest/java/com/microsoft/aad/adal/AuthenticationActivityUnitTest.java
```diff
@@ -87,6 +87,8 @@
 
     private static final int TEST_CALLING_UID = 333;
 
+    private static final String TEST_AUTHORITY = "https://login.microsoftonline.com/common";
```
heidijinxujia: What will happen if keeping using the pseudo authority url here? Will the unit test fail?

adal/src/androidTest/java/com/microsoft/aad/adal/OauthTests.java
piqiums: yes, it passed the UT.

adal/src/androidTest/java/com/microsoft/aad/adal/AuthenticationActivityUnitTest.java
piqiums: it will pass the UT. Since there's some code change in checking the authority, so we need to provide valid authority in UT.

PR #1163, opened 2018-04-04
[PAUSED - DO NOT MERGE] Fix the cache key bug when deleting the cache item.

heidijinxujia:
To-do 
1. add unit tests
2. Add logging into the cache flow.

I did a E2E verification on the sample app and the tokens are deleted successfully.

### DIFF
--- a/adal/src/main/java/com/microsoft/aad/adal/TokenCacheAccessor.java
+++ b/adal/src/main/java/com/microsoft/aad/adal/TokenCacheAccessor.java
@@ -381,10 +381,10 @@ class TokenCacheAccessor {
      */
     private List<String> getKeyListToRemoveForRT(final TokenCacheItem cachedItem) {
         final List<String> keysToRemove = new ArrayList<>();
-        keysToRemove.add(CacheKey.createCacheKeyForRTEntry(mAuthority, cachedItem.getResource(), cachedItem.getClientId(), null));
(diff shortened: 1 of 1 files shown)


adal/src/main/java/com/microsoft/aad/adal/TokenCacheAccessor.java
```diff
@@ -381,10 +381,10 @@ private void setItemToCacheForUser(final String resource, final String clientId,
      */
     private List<String> getKeyListToRemoveForRT(final TokenCacheItem cachedItem) {
         final List<String> keysToRemove = new ArrayList<>();
-        keysToRemove.add(CacheKey.createCacheKeyForRTEntry(mAuthority, cachedItem.getResource(), cachedItem.getClientId(), null));
+        keysToRemove.add(CacheKey.createCacheKeyForRTEntry(cachedItem.getAuthority(), cachedItem.getResource(), cachedItem.getClientId(), null));
```
rayluo: Quick question: if this fix is desired, do we also need to change getKeyListToRemoveForMRRT(...) and getKeyListToRemoveForFRT(...)?

iambmelt:
Tagging as DNM (`Do Not Merge`) until we hear back re: if this fix stick

adal/src/main/java/com/microsoft/aad/adal/TokenCacheAccessor.java
heidijinxujia: Yes, we should also do the same change with MRRT and FRT. This PR was paused from last week.

PR #1168, opened 2018-04-13
merge Unity, part 5, make changes according to the method calls in ad-accounts

piqiums:
merge Unity, part 5, make changes according to the method calls in ad-accounts

### DIFF
--- a/adal/src/main/java/com/microsoft/aad/adal/AuthenticationResult.java
+++ b/adal/src/main/java/com/microsoft/aad/adal/AuthenticationResult.java
@@ -61,49 +61,49 @@ public class AuthenticationResult implements Serializable {
         Succeeded,
     }
 
-    private String mCode;
+    protected String mCode;
 
-    private String mAccessToken;
+    protected String mAccessToken;
 
-    private String mRefreshToken;
+    protected String mRefreshToken;
 
     private String mTokenType;
 
-    private Date mExpiresOn;
+    protected Date mExpiresOn;
 
-    private String mErrorCode;
+    protected String mErrorCode;
 
-    private String mErrorDescription;
+    protected String mErrorDescription;
 
-    private String mErrorCodes;
+    protected String mErrorCodes;
 
-    private boolean mIsMultiResourceRefreshToken;
+    protected boolean mIsMultiResourceRefreshToken;
 
-    private UserInfo mUserInfo;
+    protected UserInfo mUserInfo;
 
-    private String mTenantId;
+    protected String mTenantId;
 
-    private String mIdToken;
+    protected String mIdToken;
 
-    private AuthenticationStatus mStatus = AuthenticationStatus.Failed;
+    protected AuthenticationStatus mStatus = AuthenticationStatus.Failed;
 
-    private boolean mInitialRequest;
+    protected boolean mInitialRequest;
     
-    private String mFamilyClientId;
+    protected String mFamilyClientId;
 
-    private boolean mIsExtendedLifeTimeToken = false;
+    protected boolean mIsExtendedLifeTimeToken = false;
 
-    private Date mExtendedExpiresOn;
+    protected Date mExtendedExpiresOn;
 
-    private String mAuthority;
+    protected String mAuthority;
 
-    private CliTelemInfo mCliTelemInfo;
+    protected CliTelemInfo mCliTelemInfo;
 
-    private HashMap<String, String> mHttpResponseBody = null;
+    protected HashMap<String, String> mHttpResponseBody = null;
 
-    private int mServiceStatusCode = -1;
+    protected int mServiceStatusCode = -1;
 
-    private HashMap<String, List<String>> mHttpResponseHeaders = null;
+    protected HashMap<String, List<String>> mHttpResponseHeaders = null;
 
     AuthenticationResult() {
         mCode = null;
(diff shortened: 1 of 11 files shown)


adal/src/main/java/com/microsoft/aad/adal/Discovery.java
```diff
@@ -68,7 +68,7 @@
     /**
      * {@link ReentrantLock} for making sure there is only one instance discovery request sent out at a time.
      */
-    private static volatile ReentrantLock sInstanceDiscoveryNetworkRequestLock;
+    private static volatile ReentrantLock sInstanceDiscoveryNetworkRequestLock = getLock();
```
addev-ashish: you changed the logic here, line 131 was getting the lock earlier

adal/src/telemetry/java/com/microsoft/aad/adal/TelemetryUtils.java
```diff
@@ -34,7 +34,7 @@
 import static com.microsoft.aad.adal.EventStrings.TENANT_ID;
 import static com.microsoft.aad.adal.EventStrings.USER_ID;
 
-final class TelemetryUtils {
+public final class TelemetryUtils {
```
addev-ashish: same

adal/src/main/java/com/microsoft/aad/adal/PackageHelper.java
```diff
@@ -45,7 +45,7 @@
 /**
  * Gets information about calling activity.
  */
-class PackageHelper {
+public class PackageHelper {
```
addev-ashish: this should not be public

adal/src/telemetry/java/com/microsoft/aad/adal/TelemetryUtils.java
```diff
@@ -58,47 +58,47 @@ private static void initializeGdprFilteredFields() {
         );
     }
 
-    static class CliTelemInfo implements Serializable {
+    public static class CliTelemInfo implements Serializable {
```
addev-ashish: same

adal/src/main/java/com/microsoft/aad/adal/StringExtensions.java
```diff
@@ -38,7 +38,7 @@
 import java.util.List;
 import java.util.StringTokenizer;
 
-final class StringExtensions {
+public final class StringExtensions {
```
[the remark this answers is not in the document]
piqiums: I don't know what class you are mentioning here. As far as I could see, there's a class named "StringHelper" that has "isNullOrBlank" method only. It doesn't have any other methods in StringExtension.

adal/src/main/java/com/microsoft/aad/adal/AuthenticationResult.java
```diff
@@ -404,13 +404,13 @@ final void setFamilyClientId(final String familyClientId) {
         mFamilyClientId = familyClientId;
     }
 
-    final void setAuthority(final String authority) {
+    public final void setAuthority(final String authority) {
```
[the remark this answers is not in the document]
piqiums: Maybe setAuthority is sensitive, but this is used by ad-accounts. What do you suggest?

adal/src/main/java/com/microsoft/aad/adal/PackageHelper.java
piqiums: why ?

adal/src/main/java/com/microsoft/aad/adal/AuthorityValidationMetadataCache.java
```diff
@@ -40,7 +40,7 @@
  * Hold the authority validation metadata.
  */
 
-final class AuthorityValidationMetadataCache {
+public final class AuthorityValidationMetadataCache {
```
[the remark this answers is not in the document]
piqiums: "please dont do this" conveys too little information. Would you explain what bad result it would lead to, and point out the right direction?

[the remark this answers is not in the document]
addev-ashish: please don't increase the public API surface of ADAL. I do not want a developer reaching into the authority validation cache and doing anything. We need to look at why you want this as public. Same/similar concern for all the points below.

adal/src/main/java/com/microsoft/aad/adal/StringExtensions.java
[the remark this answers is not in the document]
addev-ashish: Add the function you need to String Helper.

adal/src/main/java/com/microsoft/aad/adal/PackageHelper.java
addev-ashish: We do not want to support this class as a public interface. Lets talk about this in person

adal/src/main/java/com/microsoft/aad/adal/InstanceDiscoveryMetadata.java
```diff
@@ -26,66 +26,45 @@
...
- * preferred cache location and all the associated aliased host related to the passed in authority.
- */
-final class InstanceDiscoveryMetadata {
+class InstanceDiscoveryMetadata {
     private final String mPreferredNetwork;
     private final String mPreferredCache;
     private final List<String> mAliases = new ArrayList();
     private final boolean mIsValidated;
 
-    /**
```
iambmelt: Why are we deleting the javadoc?

PR #1174, opened 2018-04-25
Adding support for automation testing with Appium and Serenity

shoatman:
This is the initial merge of the work to support UI Automation using Appium, the Android Driver, Java Client and Serenity BDD.  See the instructions mark down file at the root of the automation project folder.

Known Issues:

- automation/libs/classes.jar;  Currently support for using client credentials is embedded in the .aar file of common. I've extracted the .jar from that .aar file for use within the automation project.
- no Travis integration/testing yet
- no testing on non windows platforms... for example... the certificate is currently hard-coded to be retrieved from the windows credential store
- labapi and keyvault folders contain generated code.  Please ignore in terms of code review.  Code generated using Swagger

### DIFF
--- a/automation/AutomationTestingInstructions.md
+++ b/automation/AutomationTestingInstructions.md
@@ -36,6 +36,18 @@ Other instructions:
 3. When prompted for credentails the username doesn't matter, just enter something, but the password does. Please use a strong password as this will be used to protect the private key when the certificate is written to disk.  (NOTE: The certificate will be deleted )
 
 ```powershell
+#Attribution: http://www.sherweb.com/blog/powershell-ing-on-windows-server-how-to-import-certificates-using-powershell/
+function Import-NewPfxCertificate {
+param([String]$certPath,[String]$certRootStore = “CurrentUser”,[String]$certStore = “My”,$pfxPass = $null)
+ $pfx = new-object System.Security.Cryptography.X509Certificates.X509Certificate2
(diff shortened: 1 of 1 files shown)


automation/AutomationTestingInstructions.md
```diff
@@ -0,0 +1,77 @@
...
+
+## Emulators and Phsyical Devices
+
+Many (if not all of you) at the time of writing are using a version of Windows 10 for your development. Windows 10 introduced some new security features which take advantage of the Hyper-V.  In order to run emulators locally you'll need to disable those security features.  You can do so by:
+
+1. Downloading the script found here: [https://www.microsoft.com/en-us/download/details.aspx?id=53337](https://www.microsoft.com/en-us/download/details.aspx?id=53337) 
+2. Running the following command from an elevated command prompt:
+
+```powershell
+
```
iambmelt: We'll need to add instructions here :)

shoatman: Yep.  Good catch.

shoatman:
I'm open to suggestion on how to deal with the generated code.  Their should never be a need to fix the generated code.  There may be a need to fix the templates used to generate the code or to re-run the generation as a result of API changes in the lab api.  (I'm guessing Key Vault... will continue to support their API for a much longer period of time)  I'll file an issue to track how to handle the generated code.... Do you think we need to come to a conclusion before merging this?  I ask since I'd like to unblock others on the team to begin writing tests now.

PR #1189, opened 2018-05-16
Updated to only support AAD w/ common cache; also correct an issue wi…

shoatman:
Updated TokeCache accessor to only write via common cache when the authority is not ADFS.  MSAL does not currently support ADFS and will never support ADFS v2.  

Also moved some code to ensure that all code paths benefit from having the updated authority returned on the result (sovereign).  Sent email to suggest that we always return the authority on the result.  I think anyone should be able to assume that the authority on the authentication result is not null.


iambmelt:
The CI build errored-out. Restarting for you @shoatman

PR #1190, opened 2018-05-17
refactor broker validation out of the BrokerProxy so it can be shared

microsoftjoe:
I'd like to use the signature validation code from BrokerProxy for my method in the Company Portal that will provide the enrollment id to the broker.  That method needs to verify that the call is coming from a real broker, so it needs to do the exact same signature validation.
This refactoring moves the broker validation code to a separate class, so it will be shareable.  It also aids with separation of concerns in the BrokerProxy.
If this isn't acceptable, please let me know since I will have to fork the broker validation code into the Company Portal codebase if it can't be shared from the ADAL library.

### DIFF
--- a/adal/src/main/java/com/microsoft/aad/adal/TokenCacheAccessor.java
+++ b/adal/src/main/java/com/microsoft/aad/adal/TokenCacheAccessor.java
@@ -26,6 +26,7 @@ import android.content.Context;
 
 import com.microsoft.aad.adal.AuthenticationResult.AuthenticationStatus;
 import com.microsoft.identity.common.adal.internal.AuthenticationConstants;
+import com.microsoft.identity.common.adal.internal.cache.StorageHelper;
 import com.microsoft.identity.common.adal.internal.util.StringExtensions;
 import com.microsoft.identity.common.internal.cache.ADALOAuth2TokenCache;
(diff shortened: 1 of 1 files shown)


adal/src/main/java/com/microsoft/aad/adal/BrokerValidator.java
```diff
@@ -0,0 +1,181 @@
...
+import java.security.cert.CertificateException;
+import java.security.cert.CertificateFactory;
+import java.security.cert.PKIXParameters;
+import java.security.cert.TrustAnchor;
+import java.security.cert.X509Certificate;
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.List;
+
+public class BrokerValidator {
```
shoatman: I'm thinking that we may want to put this class in our internal package within the common repo.  Would be good for it to be in package: com.microsoft.identity.common.internal.broker (which doesn't exist yet).  As proposed would be part of the public surface area of ADAL.  

Moving this to common means to PRs.  One for common and then one for ADAL to actually use the newly relocated code.

microsoftjoe: I just created a PR in common for the creation of the BrokerValidator class there.

microsoftjoe:
BrokerValidator has been moved to common as requested.

PR #1205, opened 2018-05-24
Shoatman add client info to refresh token request

shoatman:


adal/src/androidTest/java/com/microsoft/aad/adal/Util.java
```diff
@@ -159,10 +159,11 @@ static String getErrorResponseBody(final String errorCode) throws JSONException
 
     static byte[] getPoseMessage(final String refreshToken, final String clientId, final String resource)
```
iambmelt: We should fix this -- not gong to block on this PR

PR #1210, opened 2018-05-30
Addressed cache item deletion issue due to authority not being updated to preferred cache one

kreedula:
Changes to ensure if we delete cache items only if the target refresh token matches.
TODO: Understand and make changes if needed on a MalformedURLException on getAuthorityUrlWithPreferredCache() method.

### DIFF
--- a/adal/src/main/java/com/microsoft/aad/adal/TokenCacheAccessor.java
+++ b/adal/src/main/java/com/microsoft/aad/adal/TokenCacheAccessor.java
@@ -290,8 +290,9 @@ class TokenCacheAccessor {
             try {
                 mAuthority = getAuthorityUrlWithPreferredCache();
             } catch (final MalformedURLException e) {
-                //TODO : Understand what can be done here
-                Logger.d(TAG + methodName, "Authority from preferred caches is invalid");
+                // This should never happen. If the exception is thrown, mAuthority will  have the value passed from the request.
+                com.microsoft.identity.common.internal.logging.Logger.error(TAG, "Authority from preferred cache is invalid", null);
+                com.microsoft.identity.common.internal.logging.Logger.errorPII(TAG, "Failed with Exception", e);
             }
             // remove Item if oauth2_error is invalid_grant
             Logger.v(TAG + methodName, "Received INVALID_GRANT error code, remove existing cache entry.");
@@ -393,9 +394,8 @@ class TokenCacheAccessor {
         }
 
         for (final String key : keys) {
-          // mTokenCacheStore.removeItem(key);
             TokenCacheItem cacheItem = mTokenCacheStore.getItem(key);
-            if(cacheItem!=null && cacheItem.getRefreshToken().equals(toBeDeletedCacheItem.getRefreshToken())) {
+            if (cacheItem != null && cacheItem.getRefreshToken().equals(toBeDeletedCacheItem.getRefreshToken())) {
                 mTokenCacheStore.removeItem(key);
             }
         }
--- a/adal/src/main/java/com/microsoft/aad/adal/TokenCacheAccessor.java
+++ b/adal/src/main/java/com/microsoft/aad/adal/TokenCacheAccessor.java
@@ -289,9 +289,9 @@ class TokenCacheAccessor {
             // make sure to set authority to the one in preferred cache
             try {
                 mAuthority = getAuthorityUrlWithPreferredCache();
-            }catch (MalformedURLException e){
+            } catch (final MalformedURLException e) {
                 //TODO : Understand what can be done here
-                Logger.d(TAG + methodName,  "Authority from preferred caches is invalid");
+                Logger.d(TAG + methodName, "Authority from preferred caches is invalid");
             }
             // remove Item if oauth2_error is invalid_grant
             Logger.v(TAG + methodName, "Received INVALID_GRANT error code, remove existing cache entry.");
@@ -393,6 +393,7 @@ class TokenCacheAccessor {
         }
 
         for (final String key : keys) {
+          // mTokenCacheStore.removeItem(key);
             TokenCacheItem cacheItem = mTokenCacheStore.getItem(key);
             if(cacheItem!=null && cacheItem.getRefreshToken().equals(toBeDeletedCacheItem.getRefreshToken())) {
                 mTokenCacheStore.removeItem(key);
--- a/adal/src/main/java/com/microsoft/aad/adal/TokenCacheAccessor.java
+++ b/adal/src/main/java/com/microsoft/aad/adal/TokenCacheAccessor.java
@@ -286,6 +286,13 @@ class TokenCacheAccessor {
                 throw new AuthenticationException(ADALError.DEVELOPER_AUTHORITY_IS_NOT_VALID_URL, e.getMessage(), e);
             }
         } else if (AuthenticationConstants.OAuth2ErrorCode.INVALID_GRANT.equalsIgnoreCase(result.getErrorCode())) {
+            // make sure to set authority to the one in preferred cache
+            try {
+                mAuthority = getAuthorityUrlWithPreferredCache();
+            }catch (MalformedURLException e){
+                //TODO : Understand what can be done here
+                Logger.d(TAG + methodName,  "Authority from preferred caches is invalid");
+            }
             // remove Item if oauth2_error is invalid_grant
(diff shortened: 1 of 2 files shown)


adal/src/main/java/com/microsoft/aad/adal/TokenCacheAccessor.java
```diff
@@ -286,6 +286,13 @@ void updateCachedItemWithResult(final String resource, final String clientId, fi
                 throw new AuthenticationException(ADALError.DEVELOPER_AUTHORITY_IS_NOT_VALID_URL, e.getMessage(), e);
             }
         } else if (AuthenticationConstants.OAuth2ErrorCode.INVALID_GRANT.equalsIgnoreCase(result.getErrorCode())) {
+            // make sure to set authority to the one in preferred cache
+            try {
+                mAuthority = getAuthorityUrlWithPreferredCache();
+            }catch (MalformedURLException e){
```
heidijinxujia: nit: final

heidijinxujia: and missing space before {

adal/src/main/java/com/microsoft/aad/adal/TokenCacheAccessor.java
```diff
@@ -286,6 +286,13 @@ void updateCachedItemWithResult(final String resource, final String clientId, fi
...
+                //TODO : Understand what can be done here
+                Logger.d(TAG + methodName,  "Authority from preferred caches is invalid");
```
heidijinxujia: This should not happen, but if MalformedURLException is thrown, we should throw AuthenticationException with invalid url error code.

adal/src/main/java/com/microsoft/aad/adal/TokenCacheAccessor.java
kreedula: Addressed them. Thanks for pointing out

adal/src/main/java/com/microsoft/aad/adal/TokenCacheAccessor.java
kreedula: I agree this should never happen. But in this case we changing the authority url to preferred cache authority url and caller would have no knowledge about it. Not sure if throwing AuthenticationException is a right thing to do here as the url passed in by the caller is a valid one

heidijinxujia: When storing the cache item, are we always using getAuthorityUrlWithPreferredCache() or using authority from request as well?

kreedula:
@iambmelt Thank you for mentioned about the test case. The test case itself contradicts the change in the PR. The test case validates that both RT and MRRT tokens ( with different refresh token values ) should be cleared if we get an exception with `invalid_grant`. However per our discussion, we should only clear the tokens which match the target token. I have made appropriate changes to the test case here. Please have a look and let me know your thoughts

adal/src/main/java/com/microsoft/aad/adal/TokenCacheAccessor.java
kreedula: Only if the InstanceDiscoveryMetadata of the authority is successful , we update it here : [AcquireTokenRequest.java#L258](https://github.com/AzureAD/azure-activedirectory-library-for-android/blob/962ce25365320e9eb58a90f7070f46520e2c5733/adal/src/main/java/com/microsoft/aad/adal/AcquireTokenRequest.java#L258). However even if it fails, we could store the cache items with the request's authority url. But in this case `getAuthorityUrlWithPreferredCache()` would return us the request authority url itself , so I think we should be fine here.

However there is a case where we update the authority to cloudInstanceHostName here at [Oauth2.java#L287](https://github.com/AzureAD/azure-activedirectory-library-for-android/blob/962ce25365320e9eb58a90f7070f46520e2c5733/adal/src/main/java/com/microsoft/aad/adal/Oauth2.java#L287) and store the tokens in the cache with it. I am not sure if this case is handled as for the deletion to work `getAuthorityUrlWithPreferredCache()` should return the url of `cloudInstanceHostName. 

@iambmelt and @addev-ashish Correct em if I am wrong, but I think this could be potential issue here

adal/src/main/java/com/microsoft/aad/adal/TokenCacheAccessor.java
```diff
@@ -286,6 +286,13 @@ void updateCachedItemWithResult(final String resource, final String clientId, fi
                 throw new AuthenticationException(ADALError.DEVELOPER_AUTHORITY_IS_NOT_VALID_URL, e.getMessage(), e);
             }
         } else if (AuthenticationConstants.OAuth2ErrorCode.INVALID_GRANT.equalsIgnoreCase(result.getErrorCode())) {
+            // make sure to set authority to the one in preferred cache
+            try {
+                mAuthority = getAuthorityUrlWithPreferredCache();
```
iambmelt: Discussed in standup, going to leave as-is. If this `Exception` is thrown (it shouldn't), we'll fall-back on the passed-in Authority value and keep 🚚in

adal/src/main/java/com/microsoft/aad/adal/TokenCacheAccessor.java
```diff
@@ -286,6 +286,13 @@ void updateCachedItemWithResult(final String resource, final String clientId, fi
...
+            } catch (final MalformedURLException e) {
+                //TODO : Understand what can be done here
+                Logger.d(TAG + methodName, "Authority from preferred caches is invalid");
```
iambmelt: Let's do the following logging here so that we receive the `Exception` if PII logging is enabled:

```java
com.microsoft.identity.common.internal.logging.Logger.error(TAG, "Authority from preferred caches is invalid", null)
com.microsoft.identity.common.internal.logging.Logger.errorPII(TAG, "Failed with Exception", e);
```

adal/src/main/java/com/microsoft/aad/adal/TokenCacheAccessor.java
```diff
@@ -347,46 +354,50 @@ void updateTokenCacheUsingCommonCache(final String resource, final String client
...
                 break;
             default:
                 throw new AuthenticationException(ADALError.INVALID_TOKEN_CACHE_ITEM);
         }
 
         for (final String key : keys) {
-            mTokenCacheStore.removeItem(key);
+          // mTokenCacheStore.removeItem(key);
+            TokenCacheItem cacheItem = mTokenCacheStore.getItem(key);
+            if(cacheItem!=null && cacheItem.getRefreshToken().equals(toBeDeletedCacheItem.getRefreshToken())) {
```
iambmelt: Please run a formatter on this with `ctrl +alt + l`

adal/src/main/java/com/microsoft/aad/adal/TokenCacheAccessor.java
```diff
@@ -347,46 +354,50 @@ void updateTokenCacheUsingCommonCache(final String resource, final String client
...
-                keys = getKeyListToRemoveForFRT(tokenCacheItem);
+                keys = getKeyListToRemoveForFRT(toBeDeletedCacheItem);
                 break;
             default:
                 throw new AuthenticationException(ADALError.INVALID_TOKEN_CACHE_ITEM);
         }
 
         for (final String key : keys) {
-            mTokenCacheStore.removeItem(key);
+          // mTokenCacheStore.removeItem(key);
```
iambmelt: Let's remove the commented-out code

kreedula:
Thanks @iambmelt . Addressed all comments in the latest commit : https://github.com/AzureAD/azure-activedirectory-library-for-android/pull/1210/commits/f2e48f0089f5e4e4ab9601c5032cdb8fda359898

kreedula:
Sure. Thanks @iambmelt

heidijinxujia:
@kreedula Have you run a E2E test on this change?

kreedula:
@heidijinxujia Yes. I have an automation test which verifies that the token cache items are deleted when we expire AT and RT. The test was failing before this change but now it passes as we are setting the right authority.

iambmelt:
@kreedula https://github.com/AzureAD/azure-activedirectory-library-for-android/pull/1219 should be our replacement for this. We should be good to close this without merging. Let me know if you have any concerns

PR #1213, opened 2018-05-30
Test cases related to Access Token and Refresh Token expiry

kreedula:
Added a test case AcquireTokenSilentAfterExpireAT
Added a test case AcquireTokenAfterExpireAT
Changes to the automation app to create cache keys w.r.t to new Common cache changes
Added the token result to User object. It is used to expire Access token
Made changes to use expected token count based on Common vs Legacy cache


automation/src/test/java/com/microsoft/identity/common/test/automation/AcquireTokenAfterExpireAT.java
```diff
@@ -0,0 +1,149 @@
...
+
+
+    @Test
+    public void should_be_able_to_new_access_token_after_expiry_on_silent() {
+
+        givenThat(james).wasAbleTo(
+                acquireToken,
+                clickDone,
+                readCache);
+        then(james).should(seeThat(TokenCacheItemCount.displayed(), is(6)));
```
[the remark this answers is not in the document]
kreedula: I have added the logic to differentiate between the token cache count of Common and Legacy cache. Unfortunately we were using a just a number "6" in all the test cases to compare. So had to make the change to all the test cases. Now the logic is at one place if we want to change it in the future.

automationtestapp/src/main/java/com/microsoft/aad/automation/testapp/MainActivity.java
```diff
@@ -192,6 +194,14 @@ private void processReadCacheRequest() {
         return allItems;
     }
 
+    private boolean isCommonCacheUsed(){
```
kreedula: @shoatman Please review the code and let me know if there is any other scenerio where we use common cache apart from this? From `TokenCacheAccessor` I see we update `mUseCommonCache` to true only in this case.

automation/src/test/java/com/microsoft/identity/common/test/automation/AcquireTokenSilentAfterExpireATAndRT.java
```diff
@@ -0,0 +1,151 @@
...
+
+@RunWith(SerenityParameterizedRunner.class)
+public class AcquireTokenSilentAfterExpireATAndRT {
+
+    @TestData
+    public static Collection<Object[]> FederationProviders(){
+
+
+        return Arrays.asList(new Object[][]{
+                {"ADFSv2"}/*,
```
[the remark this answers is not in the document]
kreedula: Thanks for pointing out. It was accidental, addressed it

automationtestapp/src/main/java/com/microsoft/aad/automation/testapp/MainActivity.java
shoatman: We probably need something based on version rather than on whether the default token cache is being used.  If we want to test versions of ADAL that do not contain common cache... for example.

kreedula: As discussed, I looking for a specific class in common using reflection in addition to the default token cache to determine if we are saving using common

kreedula:
@shoatman Thank you for you review and suggestion, I made the required change and also resolved conflicts with your force refresh change. Please have a look when you get a chance.

PR #1215, opened 2018-05-31
Updates README instructions for submodule/common use.

iambmelt:
This change updates the `README` with instructions for including ADAL in your project.

The include-as-zip option has been removed because it's incredibly _easy_ to **mess up**.
Brief explanation:
- Downloading as zip via GitHub doesn't include the submodule, only an empty directory
- The submodule/branch/SHA being used by ADAL is contingent on the branch being used
- The submodule ref in `.gitmodules` may not necessarily `ref` to `HEAD`, so you'd have to navigate to the proper commit in the version tree before download **OR** `git clone` it (which is what we were trying to prevent having to do in the first place)

`git clone` avoids all of these headaches. Just do that!

### DIFF
--- a/README.md
+++ b/README.md
@@ -69,7 +69,7 @@ To build with Gradle,
 * Go to the root folder where you cloned this repo
 * To run the sample app, connect the test device and run the command: `./gradlew :userappwithbroker:installDebug`
 * You should see app 'Fancy ADAL Test App' installed on the test device
-* Select an authority, [optionally] enter a login hint and/or query parameters, and clikc `Acquire Token` to enter credentials with AAD
+* Select an authority, [optionally] enter a login hint and/or query parameters, and click `Acquire Token` to enter credentials with AAD
 
 To build with Maven, you can use the pom.xml at top level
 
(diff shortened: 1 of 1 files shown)


iambmelt:
The `README` still mentions the `hello` sample app. This doesn't exist anymore and the `README` should no longer mention it.

README.md
```diff
@@ -67,18 +67,18 @@ To build with Gradle,
 * Clone this repo in to a directory of your choice
 * Setup emulator with SDK 23
 * Go to the root folder where you cloned this repo
-* To run the sample app, connect the test device and run the command: `./gradlew :sample:installDebug`
-* You should see app 'hello' installed in the test device
-* Enter test user credentials to try
+* To run the sample app, connect the test device and run the command: `./gradlew :userappwithbroker:installDebug`
+* You should see app 'Fancy ADAL Test App' installed on the test device
+* Select an authority, [optionally] enter a login hint and/or query parameters, and clikc `Acquire Token` to enter credentials with AAD
```
kreedula: Typo clikc -> click

iambmelt: Oops. Thanks @kreedula  😄

iambmelt: Fixed via 831a256ed22530a555228ee898f70923b8ab7f0c

iambmelt:
Since these instructions mention the POM, I'll investigate reintroducing it today

PR #1216, opened 2018-06-04
Shoatman force refresh

shoatman:
Adding forceRefresh parameter to acquire token silent overloads.  New parameter instructs ADAL to:

- If broker/user in broker... use broker to refresh access token
- If no broker or not user in the broker... use the local refresh token to refresh the access token


adal/src/main/java/com/microsoft/aad/adal/AuthenticationContext.java
```diff
@@ -726,6 +726,93 @@ public void onError(Exception exc) {
...
+     *                 {@link AuthenticationResult #getUserInfo()}
+     * @param forceRefresh when true, access token is renewed using broker if available; otherwise, uses local refresh token
+     * @return A {@link Future} object representing the
+     * {@link AuthenticationResult} of the call. It contains Access
+     * Token,the Access Token's expiration time, Refresh token, and
+     * {@link UserInfo}.
+     * @throws AuthenticationException If silent request fails to get the token back.
+     * @throws InterruptedException    If the main thread is interrupted before or during the activity.
+     */
+    public AuthenticationResult acquireTokenSilentSync(String resource, String clientId, String userId, boolean forceRefresh)
```
iambmelt: Can the alternative implementations of `acquireTokenSilentSync` call this (passing `false`) to achieve their current behavior? If yes, can we turn those implementations into thin wraps around this method? (or possibly an internal call which allows for the `apiEvent` to be passed?)

adal/src/main/java/com/microsoft/aad/adal/AuthenticationContext.java
```diff
@@ -726,6 +726,93 @@ public void onError(Exception exc) {
...
+
+        final String methodName = ":acquireTokenSilentSync";
+        checkPreRequirements(resource, clientId);
+        checkADFSValidationRequirements(null);
+        final AtomicReference<AuthenticationResult> authenticationResult = new AtomicReference<>();
+        final AtomicReference<Exception> exception = new AtomicReference<>();
+        final CountDownLatch latch = new CountDownLatch(1);
+
+        final String requestId = Telemetry.registerNewRequest();
+        final APIEvent apiEvent = createApiEvent(mContext, clientId, requestId, EventStrings.ACQUIRE_TOKEN_SILENT_SYNC);
```
iambmelt: I suspect that this call should get its own `API_ID` -- see [this doc](https://microsoft.sharepoint.com/:w:/r/teams/aad/devex/_layouts/15/Doc.aspx?sourcedoc=%7BAAE38591-8A29-4A50-AE00-B9F66505B6D2%7D&file=Telemetry%20DDS.docx&action=default&mobileredirect=true) (requires corpnet) for a table of API IDs and their meanings. If iOS is adding this functionality, we'll need to collaborate with them to ensure we're using the same values

adal/src/main/java/com/microsoft/aad/adal/AuthenticationContext.java
```diff
@@ -845,6 +932,47 @@ public void acquireTokenSilentAsync(String resource,
...
+                                        String userId,
+                                        boolean forceRefresh,
+                                        AuthenticationCallback<AuthenticationResult> callback) {
+        if (!checkPreRequirements(resource, clientId, callback) || !checkADFSValidationRequirements(null, callback)) {
+            // AD FS validation cannot be perfomed, stop executing
+            return;
+        }
+
+        final String requestId = Telemetry.registerNewRequest();
+        final APIEvent apiEvent = createApiEvent(mContext, clientId, requestId,
```
iambmelt: [Potentially] Same idea here for API ID / telemetry

adal/src/main/java/com/microsoft/aad/adal/AuthenticationContext.java
```diff
@@ -726,6 +726,93 @@ public void onError(Exception exc) {
...
+                    public void onError(Exception exc) {
+                        exception.set(exc);
+                        latch.countDown();
+                    }
+                });
+
+        latch.await();
+
+        Exception e = exception.get();
+        if (e != null) {
```
iambmelt: Haven't traced the entirety of this code to be sure, but I suspect you'll need to set the success state on the `apiEvent` in `onSuccess` and `onError` -- if there is an `Exception` and message, you'll want to set these values too. See `setWasApiCallSuccessful` and `stopTelemetryAndFlush`

adal/src/main/java/com/microsoft/aad/adal/AuthenticationContext.java
```diff
@@ -726,6 +726,93 @@ public void onError(Exception exc) {
...
+            if (e instanceof AuthenticationException) {
+                throw (AuthenticationException) e;
+            } else if (e instanceof RuntimeException) {
+                throw (RuntimeException) e;
```
iambmelt: Is there any sense in wrapping this inside of an `AuthenticationException` to rethrow? Are we expecting devs to catch instances of `RuntimeException` or let it crash? (This was TBD, as I recall)

shoatman: This is existing pattern.  I'll take a look.

adal/src/main/java/com/microsoft/aad/adal/AuthenticationContext.java
shoatman: Yep.

shoatman:
Closing will re-open post rebase

PR #1232, opened 2018-06-22
merge the change to deprecating Unity-fork to dev branch

piqiums:
merge the change deprecating Unity-fork to dev branch.


adal/src/main/java/com/microsoft/aad/adal/ADALError.java
```diff
@@ -344,11 +344,18 @@
      */
     BROKER_PACKAGE_NAME_NOT_FOUND("Broker is not installed in your system"),
 
+    BROKER_PRT_FAILED("Broker is failed to get PRT"),
```
iambmelt: Grammatical issue with this error, should be `Broker failed to get PRT`

adal/src/main/java/com/microsoft/aad/adal/AuthenticationRequest.java
```diff
@@ -201,6 +207,10 @@ public String getResource() {
     public String getClientId() {
         return mClientId;
     }
+    
+    public void setClientId(String id) {
```
iambmelt: Nit about `final` params

adal/src/main/java/com/microsoft/aad/adal/AuthorityValidationMetadataCache.java
```diff
@@ -42,7 +42,7 @@
  * Hold the authority validation metadata.
  */
 
-final class AuthorityValidationMetadataCache {
+class AuthorityValidationMetadataCache {
```
iambmelt: Does this class now get extended somewhere?

iambmelt:
@piqiums Can you bring the `common` submodule up to date? I'll continue reviewing afterwards

adal/src/main/java/com/microsoft/aad/adal/ADALError.java
piqiums: this is moved from Unity. good catch. I will edit this.

adal/build.gradle
```diff
@@ -207,6 +207,7 @@ task javadoc(type: Javadoc) {
         }
     }
     destinationDir = reporting.file("$project.buildDir/outputs/jar/javadoc/")
+    failOnError false
```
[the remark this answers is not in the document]
iambmelt: Since we can't break javadoc, we need to remove this and troubleshoot why the build is failing without it

adal/src/main/java/com/microsoft/aad/adal/AuthenticationResult.java
```diff
@@ -69,55 +69,55 @@
         Succeeded,
     }
 
-    private String mCode;
+    protected String mCode;
```
iambmelt: I'm not seeing usages in the code that explain why this became `protected` -- is this still needed?

adal/src/main/java/com/microsoft/aad/adal/AuthenticationResult.java
```diff
@@ -440,13 +444,13 @@ final void setFamilyClientId(final String familyClientId) {
         mFamilyClientId = familyClientId;
     }
 
-    final void setAuthority(final String authority) {
+    public final void setAuthority(final String authority) {
```
iambmelt: This doesn't look like it needs to change; I see usages inside of `OAuth2` and nowhere else... can this be left at default visibility?

adal/src/main/java/com/microsoft/aad/adal/AuthenticationResult.java
```diff
@@ -440,13 +444,13 @@ final void setFamilyClientId(final String familyClientId) {
...
         if (!StringExtensions.isNullOrBlank(authority)) {
             mAuthority = authority;
         }
     }
 
-    final CliTelemInfo getCliTelemInfo() {
+    public final CliTelemInfo getCliTelemInfo() {
```
iambmelt: Same note as above, is `public` vis required?

adal/src/main/java/com/microsoft/aad/adal/AuthorityValidationMetadataCache.java
```diff
@@ -57,8 +57,8 @@
 
     private static ConcurrentMap<String, InstanceDiscoveryMetadata> sAadAuthorityHostMetadata = new ConcurrentHashMap<>();
 
-    private AuthorityValidationMetadataCache() {
-        // DateExtensions class, no public constructor
+    protected AuthorityValidationMetadataCache() {
```
iambmelt: Why does a utility class need a `protected` constructor?

adal/src/main/java/com/microsoft/aad/adal/AuthorityValidationMetadataCache.java
```diff
@@ -69,7 +69,7 @@ static boolean isAuthorityValidated(final URL authorityUrl) {
         return containsAuthorityHost(authorityUrl) && getCachedInstanceDiscoveryMetadata(authorityUrl).isValidated();
     }
 
-    static InstanceDiscoveryMetadata getCachedInstanceDiscoveryMetadata(final URL authorityUrl) {
+    public static InstanceDiscoveryMetadata getCachedInstanceDiscoveryMetadata(final URL authorityUrl) {
```
iambmelt: This class isn't `public` but these methods are.... 🤔

adal/src/main/java/com/microsoft/aad/adal/WebRequestHandler.java
```diff
@@ -0,0 +1,124 @@
...
+        return headers;
+    }
+
+    /**
+     * Sets correlationId.
+     *
+     * @param requestCorrelationId {@link UUID}
+     */
+    public void setRequestCorrelationId(UUID requestCorrelationId) {
+        this.mRequestCorrelationId = requestCorrelationId;
```
iambmelt: Since `this` isn't needed, let's remove and make the parameter to this method `final`

adal/src/main/java/com/microsoft/aad/adal/WebRequestHandler.java
```diff
@@ -0,0 +1,124 @@
...
+    }
+
+    @Override
+    public void setClientVersion(String clientVersion) {
```
iambmelt: `final` param

adal/src/main/java/com/microsoft/aad/adal/WebRequestHandler.java
```diff
@@ -0,0 +1,124 @@
...
+        if (!StringExtensions.isNullOrBlank(clientVersion)) {
+            mCurrentClientVersion = clientVersion;
+        }
+    }
+
+    public void setBrokerVersion(String brokerVersion) {
```
iambmelt: nit: `final` param

adal/src/main/java/com/microsoft/aad/adal/AuthenticationResult.java
piqiums: this method is called in Ad-accounts, so I need to make it public as it was in Unity-fork

adal/src/main/java/com/microsoft/aad/adal/AuthenticationResult.java
piqiums: this method is called in BrokerClient in Ad-accounts, so I need to make it public as it was in Unity-fork

adal/src/main/java/com/microsoft/aad/adal/AuthenticationResult.java
piqiums: this variable is has no other usage. I've changed it back to private. But other variables needs to be updated to 'protected' since BrokerTokenResult in Ad-accounts is extending AuthenticationResult and will use the member variables directly.

adal/src/main/java/com/microsoft/aad/adal/AuthorityValidationMetadataCache.java
piqiums: changed back to private

adal/src/main/java/com/microsoft/aad/adal/AuthorityValidationMetadataCache.java
piqiums: these methods are used in Ad-accounts, but I double checked they're used in the same package "com.microsoft.aad.adal", so it's OK to remove the added 'public'

adal/src/main/java/com/microsoft/aad/adal/Oauth2.java
```diff
@@ -241,6 +265,13 @@ public String buildRefreshTokenRequestMessage(String refreshToken)
                     StringExtensions.urlFormEncode(mRequest.getResource()));
         }
 
+        // sending redirect uri for the refresh token request if it's provided
+        if (!StringExtensions.isNullOrBlank(mRequest.getRedirectUri())
```
addev-ashish: can we do a quick test about what happens if redirecturi is garbage for the non-broker scenario?

adal/src/main/java/com/microsoft/aad/adal/WebRequestHandler.java
```diff
@@ -0,0 +1,124 @@
...
+import java.util.Map;
+import java.util.UUID;
+
+
+/**
+ * It uses one time async task. WebRequest are wrapped here to prevent multiple
+ * reuses for same tasks. Each request returns a handler for cancel action. Call
+ * this from UI thread to correctly create async task and execute.
+ */
+public class WebRequestHandler implements IWebRequestHandler {
```
addev-ashish: i am a little confused here, don't we have an existing implementation of this class in ADAL?

adal/src/telemetry/java/com/microsoft/aad/adal/TelemetryUtils.java
```diff
@@ -36,7 +36,7 @@
 import static com.microsoft.aad.adal.EventStrings.TENANT_ID;
 import static com.microsoft.aad.adal.EventStrings.USER_ID;
 
-final class TelemetryUtils {
+public final class TelemetryUtils {
```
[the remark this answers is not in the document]
piqiums: This is used in Ad-accounts in another package, so it needs to be public

adal/src/main/java/com/microsoft/aad/adal/WebRequestHandler.java
piqiums: good catch. this is already existing in 'common' submodule. I will delete this

adal/src/main/java/com/microsoft/aad/adal/WebRequestHandler.java
piqiums: this is dup to the one in 'common' submodule, I will remove this file.

adal/src/main/java/com/microsoft/aad/adal/WebRequestHandler.java
piqiums: this is dup to the one in 'common' submodule, I will remove this file.

adal/src/main/java/com/microsoft/aad/adal/WebRequestHandler.java
piqiums: this is dup to the one in 'common' submodule, I will remove this file.

adal/build.gradle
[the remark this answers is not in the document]
piqiums: done

iambmelt:
@piqiums Please update your submodule to remove conflicts

adal/src/main/java/com/microsoft/aad/adal/HashMapExtensions.java
```diff
@@ -109,11 +109,11 @@ private HashMapExtensions() {
      * get key value pairs from response.
      *
      * @param webResponse HttpWebResponse to convert to a map
-     * @return Map
+     * @return HashMap
      * @throws JSONException
      */
-    static Map<String, String> getJsonResponse(HttpWebResponse webResponse) throws JSONException {
-        final Map<String, String> response = new HashMap<>();
+    static HashMap<String, String> getJsonResponse(HttpWebResponse webResponse) throws JSONException {
```
iambmelt: I'm assuming this is using `HashMap` because it implements `Serializable` and this is going to be on an `Intent` -- where is this happening though?

adal/src/main/java/com/microsoft/aad/adal/IJWSBuilder.java
```diff
@@ -49,4 +53,9 @@
      */
     String generateSignedJWT(String nonce, String submitUrl, RSAPrivateKey privateKey,
                              RSAPublicKey pubKey, X509Certificate x509Certificate) throws AuthenticationException;
+
+
+    String generateJWT(HashMap<String, String> header,
```
iambmelt: Can this just be a `Map`?

adal/src/main/java/com/microsoft/aad/adal/JWSBuilder.java
```diff
@@ -100,6 +106,44 @@ private JwsHeader() {
...
+            int expTimeInSeconds) throws JSONException, UnsupportedEncodingException {
+        Logger.v(TAG, "Generating JWT.");
+        JSONObject headerJson = generateJson(header, expTimeInSeconds);
+        JSONObject bodyJson = generateJson(body, expTimeInSeconds);
+        String signingInput = StringExtensions.encodeBase64URLSafeString(headerJson.toString().getBytes(AuthenticationConstants.ENCODING_UTF8))
+                + "." + StringExtensions.encodeBase64URLSafeString(bodyJson.toString().getBytes(AuthenticationConstants.ENCODING_UTF8));
+        return signingInput;
+    }
+
+    private JSONObject generateJson(HashMap<String, String> values, int expireSeconds)
```
iambmelt: Input parameter should be `Map`

adal/src/main/java/com/microsoft/aad/adal/JWSBuilder.java
```diff
@@ -100,6 +106,44 @@ private JwsHeader() {
...
+    /**
+     * Generate generic String key/value pair JWT.
+     *
+     * @param header
+     * @param body
+     * @return String Base64URLSafe(header)+Base64URLSafe(body)
+     * @throws JSONException
+     * @throws UnsupportedEncodingException
+     */
+    public String generateJWT(HashMap<String, String> header, HashMap<String, String> body,
```
iambmelt: Input parameter should be `Map` unless there's some specific reason to use `HashMap`

adal/src/main/java/com/microsoft/aad/adal/StringExtensions.java
```diff
@@ -39,7 +41,8 @@
 import java.util.StringTokenizer;
 
 
-final class StringExtensions {
+
+public final class StringExtensions {
```
iambmelt: This entire class also exists in `common` -- can it be removed?

adal/src/main/java/com/microsoft/aad/adal/TokenCacheItem.java
```diff
@@ -398,6 +400,19 @@ public final boolean isExtendedLifetimeValid() {
         return false;
     }
 
+    protected final Date getTokenUpdateTime() {
+        return this.mTokenUpdatedTime;
+    }
+
+    final void setTokenUpdateTime(final Date tokenUpateTime) {
+        if (null == tokenUpateTime) {
+            this.mTokenUpdatedTime= null;
```
iambmelt: Run formatter - weird spacing

adal/src/main/java/com/microsoft/aad/adal/HashMapExtensions.java
piqiums: in BrokerProxy, please check the code here. https://github.com/AzureAD/azure-activedirectory-library-for-android/blob/dev/adal/src/main/java/com/microsoft/aad/adal/BrokerProxy.java#L485

adal/src/main/java/com/microsoft/aad/adal/IJWSBuilder.java
piqiums: since we're using HashMap for the responseHeader and responseBody, just as you mentioned as above, I think we'd better to keep it consistent to avoid any unexpected error.

adal/src/main/java/com/microsoft/aad/adal/JWSBuilder.java
piqiums: ditto.

adal/src/main/java/com/microsoft/aad/adal/JWSBuilder.java
piqiums: ditto

adal/src/main/java/com/microsoft/aad/adal/AuthenticationResult.java
```diff
@@ -71,53 +71,53 @@
 
     private String mCode;
 
-    private String mAccessToken;
+    protected String mAccessToken;
```
iambmelt: Instead of making these fields `protected`, can we add `protected` setters?

adal/src/main/java/com/microsoft/aad/adal/Discovery.java
```diff
@@ -59,7 +59,7 @@
  * sends common as a tenant name. Discovery checks only authorization endpoint.
  * It does not do tenant verification. Initialize and call from UI thread.
  */
-final class Discovery {
+class Discovery {
```
iambmelt: Is there a reason this is no longer `final`?

piqiums: in Ad-accounts, there's a class name "BrokerDiscovery" that extends the Discovery class, so it could not be final

adal/src/main/java/com/microsoft/aad/adal/BasicWebViewClient.java
```diff
@@ -52,9 +52,9 @@
 
     public static final String BLANK_PAGE = "about:blank";
 
-    private final String mRedirect;
-    private final AuthenticationRequest mRequest;
-    private final Context mCallingContext;
+    protected String mRedirect;
```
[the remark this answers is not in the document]
piqiums: From my perspective, ideally, when we decide if it's a private or protected, we should not look into the existing code, we should decide private/protected only on if we want the subclass access that field.  Actually, exposing a protected member variable is the same with exposing a protected setter. I would suggest that we could check in the code first and then we should walk thru the class hierarchy of WebViewClient later to decide access modifier.

adal/src/main/java/com/microsoft/aad/adal/AuthenticationResult.java
piqiums: From my perspective, ideally, when we decide if it's a private or protected, we should not look into the existing code, we should decide private/protected only on if we want the subclass access that field. Actually, exposing a protected member variable is the same with exposing a protected setter. I would suggest that we could check in the code first and then we should walk thru the class hierarchy of WebViewClient later to decide access modifier.

adal/src/main/java/com/microsoft/aad/adal/JWSBuilder.java
piqiums: done

adal/src/main/java/com/microsoft/aad/adal/TokenCacheItem.java
piqiums: done

adal/src/main/java/com/microsoft/aad/adal/Oauth2.java
iambmelt: @piqiums Have you run this test? Please sync with @addev-ashish on the outcome

adal/src/main/java/com/microsoft/aad/adal/AuthenticationConstants.java
```diff
@@ -1,846 +0,0 @@
-// Copyright (c) Microsoft Corporation.
```
iambmelt: @piqiums This class should not have been deleted - it is part of the public API surface.

We'll likely have to revert this PR, can sync in scrum

PR #1239, opened 2018-06-29
Fixed bug and added test for ATSSync overload

shoatman:
Fixed a copy/paste bug introduced during refactoring to address code review.  Added to test to provided coverage for this overload.


shoatman:
@iambmelt - Thanks for the review.  I updated the submodule pointer.  Good catch.  The builds are now passing as well.... had to restart one after updating the sub-module.  Fun.

Issue #1191, opened 2018-05-17
update ADAL's targetSdkVersion to 26


piqiums:
On 12/19/2017 Google announced the following:

In the second half of 2018, Play will require that new apps and app updates target a recent Android API level. This will be required for new apps in August 2018, and for updates to existing apps in November 2018. This is to ensure apps are built on the latest APIs optimized for security and performance.

source: https://android-developers.googleblog.com/2017/12/improving-app-security-and-performance.html

August 2018: New apps required to target API level 26 (Android 8.0) or higher.
November 2018: Updates to existing apps required to target API level 26 or higher.
2019 onwards: Each year the targetSdkVersion requirement will advance. Within one year following each Android dessert release, new apps and app updates will need to target the corresponding API level or higher.

 Hence, we need to update the targetSdkVersion to 26 for ADAL

piqiums:
working in progress

iambmelt:
Closed by #1240 , thanks @piqiums @kreedula

PR #1240, opened 2018-07-07
Updated Adal target version to 27

kreedula:
Per this issue : https://github.com/AzureAD/azure-activedirectory-library-for-android/issues/1191
Updated the target 27 which is the latest stable version

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -16,7 +16,7 @@ android:
     - platform-tools
     - tools
     - build-tools-27.0.3
-    - android-25
+    - android-27
     - android-21
     - extra
     - extra-android-m2repository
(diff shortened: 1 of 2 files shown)


kreedula:
Thanks @iambmelt . I have updated the `.travis.yml` file.

.travis.yml
```diff
@@ -16,7 +16,7 @@ android:
     - platform-tools
     - tools
     - build-tools-27.0.3
-    - android-25
+    - android-27
     - android-21
```
addev-ashish: why is this line still there?

kreedula: Missed to remove android-21, removed in latest commit

kreedula: Actually, android-21 is for the emulator config. Added it back with a comment

PR #1245, opened 2018-07-11
Version update changes to 1.14.1

kreedula:


kreedula:
Thanks @iambmelt . adalLegacy property in versions.gradle is not part of 1.14.0 as it is added by Shane after the release.

PR #1255, opened 2018-07-18
Removed common dependency from automation test app

kreedula:


kreedula:
Closing PR as we decided it's not needed

PR #1259, opened 2018-07-19
make the AuthenticationConstants in ADAL a subclass of the one in common-core

piqiums:
make the AuthenticationConstants in ADAL a subclass of the one in common-core

### DIFF
--- a/adal/src/main/java/com/microsoft/aad/adal/AuthenticationConstants.java
+++ b/adal/src/main/java/com/microsoft/aad/adal/AuthenticationConstants.java
@@ -1,0 +1,37 @@
+// Copyright (c) Microsoft Corporation.
+// All rights reserved.
+//
+// This code is licensed under the MIT License.
+//
+// Permission is hereby granted, free of charge, to any person obtaining a copy
+// of this software and associated documentation files(the "Software"), to deal
+// in the Software without restriction, including without limitation the rights
+// to use, copy, modify, merge, publish, distribute, sublicense, and / or sell
+// copies of the Software, and to permit persons to whom the Software is
+// furnished to do so, subject to the following conditions :
+//
+// The above copyright notice and this permission notice shall be included in
+// all copies or substantial portions of the Software.
+//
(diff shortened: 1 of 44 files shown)


heidijinxujia:
Did I understand wrong? I remember Shane said we were not going to make AuthenticationConstants in ADAL to inherit from common constants class. Because it will be a breaking change at ADAL surface. @shoatman Please correct me if wrong.

adal/src/main/java/com/microsoft/aad/adal/AuthenticationConstants.java
```diff
@@ -0,0 +1,37 @@
...
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+
+package com.microsoft.aad.adal;
+
+/**
+ * {@link AuthenticationConstants} contains all the constant value the SDK is using.
+ */
+public final class AuthenticationConstants extends com.microsoft.identity.common.adal.internal.BaseAuthenticationConstants {
```
iambmelt: @shoatman had strongly advised against implementing this change via inheritance. While it is quite convenient to relocate the constants in this way, to his point, changing the inheritance hierarchy of this class could be construed as a major-version breaking change. As such, I'm going to request changes :-/

adal/src/main/java/com/microsoft/aad/adal/TokenCacheAccessor.java
```diff
@@ -25,7 +25,7 @@
 import android.content.Context;
 
 import com.microsoft.aad.adal.AuthenticationResult.AuthenticationStatus;
-import com.microsoft.identity.common.adal.internal.AuthenticationConstants;
+import com.microsoft.aad.adal.AuthenticationConstants;
```
iambmelt: This appears to be unused

PR #1260, opened 2018-07-19
add AuthenticationConstants back

piqiums:
add AuthenticationConstants back


piqiums:
@iambmelt  After talking with Shane, he thinks the change make ADAL/AuthenticationConstants inheriting from common/AuthenticationConstants is a breaking change since the direct super class is changed. In Shane's opinion, as long as we have AuthenticationConstants in ADAL as a legacy, then it's OK.

PR #1261, opened 2018-07-20
Cherrypick Auth Dialog crash change and update to 1.15.0-RC

kreedula:


kreedula:
Thank you @iambmelt and @shoatman

Issue #1267, opened 2018-07-20
Javadocs on dialog based AcquireTokens could be improved


danieldobalian:
The javadocs for the two dialog-based acquiretoken overloads are not very intuitive, do not express limitations of broker interaction in these scenarios, and do not describe the functional behavior of the method.

danieldobalian:
Addressed in https://github.com/AzureAD/azure-activedirectory-library-for-android/pull/1268

PR #1268, opened 2018-07-20
Improving dialog-based acquireToken javadocs

danieldobalian:
Issue: https://github.com/AzureAD/azure-activedirectory-library-for-android/issues/1267

Improving dialog javadocs by explaining usage, limitations (in this case these overloads do not work with brokered flows), and overall improve the diction.

### DIFF
--- a/adal/src/main/java/com/microsoft/aad/adal/AuthenticationContext.java
+++ b/adal/src/main/java/com/microsoft/aad/adal/AuthenticationContext.java
@@ -560,14 +560,16 @@ public class AuthenticationContext {
     }
 
     /**
-     * This uses new dialog based prompt. It will create a handler to run the
-     * dialog related code. It will start interactive flow if needed. It checks
-     * the cache to return existing result if not expired. It tries to use
-     * refresh token if available. If it fails to get token with refresh token,
-     * behavior will depend on options. If promptbehavior is AUTO, it will
-     * remove this refresh token from cache and fall back on the UI. Default is
-     * AUTO.
-     *
+     * acquireToken will authorize an end user to call the specified resource. 
+     * The minted access token will be sent to the {@link AuthenticationCallback}
+     * and can be used to call the specified protected resource. 
+     * By default, acquireToken will attempt to fulfill the request silently, and 
+     * perform an interactive login if needed or explicitly specified in 
+     * the request.  This overload uses an 
+     * [AlertDialog](https://developer.android.com/guide/topics/ui/dialogs)
+     * in cases UI is required. This overload does not support any flow
+     * requiring a 
+     * [token broker](https://github.com/AzureAD/azure-activedirectory-library-for-android/wiki/Broker).
      * @param resource             required resource identifier.
      * @param clientId             required client identifier.
      * @param redirectUri          Optional. It will use packagename and provided suffix
@@ -602,9 +604,15 @@ public class AuthenticationContext {
     }
 
     /**
-     * acquireToken will start an interactive auth flow to acquire new tokens 
-     * with the requested claims. Bypasses token cache if promptbehavior is not AUTO or claims are passed. This overload uses new dialog based prompt. 
-     * It will create a handler to run the dialog related code. 
+     * acquireToken will authorize an end user to call the specified resource. 
+     * The minted access token will be sent to the {@link AuthenticationCallback}
+     * and can be used to call the specified protected resource. 
+     * Bypasses token cache if @param prompt is not AUTO or claims are passed.  
+     * This overload uses an 
+     * [AlertDialog](https://developer.android.com/guide/topics/ui/dialogs)
+     * in cases UI is required. This overload does not support any flow
+     * requiring a 
+     * [token broker](https://github.com/AzureAD/azure-activedirectory-library-for-android/wiki/Broker).
      *
      * @param resource             required resource identifier.
      * @param clientId             required client identifier.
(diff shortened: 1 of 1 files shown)


adal/src/main/java/com/microsoft/aad/adal/AuthenticationContext.java
```diff
@@ -560,14 +560,16 @@ public void acquireToken(final IWindowComponent fragment, final String resource,
...
-     * This uses new dialog based prompt. It will create a handler to run the
-     * dialog related code. It will start interactive flow if needed. It checks
-     * the cache to return existing result if not expired. It tries to use
-     * refresh token if available. If it fails to get token with refresh token,
-     * behavior will depend on options. If promptbehavior is AUTO, it will
-     * remove this refresh token from cache and fall back on the UI. Default is
-     * AUTO.
-     *
+     * acquireToken will authorize an end user to call the specified resource. 
+     * The minted access token will be sent to the {@link AuthenticationCallback}
```
iambmelt: Note about clarity: "minted" may confuse users into thinking that our library _creates_ access tokens when, in reality, they are given to us.

How about:
- "The access token returned from the STS..."
- "The acquired access token...."
- "The access token"

adal/src/main/java/com/microsoft/aad/adal/AuthenticationContext.java
```diff
@@ -602,9 +604,15 @@ public void acquireToken(String resource, String clientId, @Nullable String redi
     }
 
     /**
-     * acquireToken will start an interactive auth flow to acquire new tokens 
-     * with the requested claims. Bypasses token cache if promptbehavior is not AUTO or claims are passed. This overload uses new dialog based prompt. 
-     * It will create a handler to run the dialog related code. 
+     * acquireToken will authorize an end user to call the specified resource. 
+     * The minted access token will be sent to the {@link AuthenticationCallback}
```
iambmelt: Another spot: "minted" feels misleading

adal/src/main/java/com/microsoft/aad/adal/AuthenticationContext.java
```diff
@@ -560,14 +560,16 @@ public void acquireToken(final IWindowComponent fragment, final String resource,
...
-     * behavior will depend on options. If promptbehavior is AUTO, it will
-     * remove this refresh token from cache and fall back on the UI. Default is
-     * AUTO.
-     *
+     * acquireToken will authorize an end user to call the specified resource. 
+     * The minted access token will be sent to the {@link AuthenticationCallback}
+     * and can be used to call the specified protected resource. 
+     * By default, acquireToken will attempt to fulfill the request silently, and 
+     * perform an interactive login if needed or explicitly specified in 
+     * the request.  This overload uses an
```
iambmelt: There are two-spaces after the period in the javadoc. Just a formatting thing.

adal/src/main/java/com/microsoft/aad/adal/AuthenticationContext.java
```diff
@@ -602,9 +604,15 @@ public void acquireToken(String resource, String clientId, @Nullable String redi
...
-     * acquireToken will start an interactive auth flow to acquire new tokens 
-     * with the requested claims. Bypasses token cache if promptbehavior is not AUTO or claims are passed. This overload uses new dialog based prompt. 
-     * It will create a handler to run the dialog related code. 
+     * acquireToken will authorize an end user to call the specified resource. 
+     * The minted access token will be sent to the {@link AuthenticationCallback}
+     * and can be used to call the specified protected resource. 
+     * Bypasses token cache if @param prompt is not AUTO or claims are passed.  
+     * This overload uses an 
+     * [AlertDialog](https://developer.android.com/guide/topics/ui/dialogs)
+     * in cases UI is required. This overload does not support any flow
```
iambmelt: "in cases UI is required"

should it be "in cases [when/where] UI is required"

iambmelt: We may want to say `interaction` instead of `UI` -- that's what we really mean, right? ;-)

adal/src/main/java/com/microsoft/aad/adal/AuthenticationContext.java
```diff
@@ -560,14 +560,16 @@ public void acquireToken(final IWindowComponent fragment, final String resource,
...
-     * AUTO.
-     *
+     * acquireToken will authorize an end user to call the specified resource. 
+     * The minted access token will be sent to the {@link AuthenticationCallback}
+     * and can be used to call the specified protected resource. 
+     * By default, acquireToken will attempt to fulfill the request silently, and 
+     * perform an interactive login if needed or explicitly specified in 
+     * the request.  This overload uses an 
+     * [AlertDialog](https://developer.android.com/guide/topics/ui/dialogs)
+     * in cases UI is required. This overload does not support any flow
```
iambmelt: Same note as the other overload.... seems like a word is missing before "UI"... may be more clear to say "interaction"

adal/src/main/java/com/microsoft/aad/adal/AuthenticationContext.java
danieldobalian: Good call out. Fixed.

adal/src/main/java/com/microsoft/aad/adal/AuthenticationContext.java
danieldobalian: Bad habit :)

PR #1283, opened 2018-08-01
Removed commit logic and all version tasks except Nightly

kreedula:
Tasks for RC , major , minor and patch would require manual intervention to update change logs and approving PR, so removed them


iambmelt:
@kreedula So the plan is, CI will call `versionNightly`, run the build, and publish the generated artifact to the nightly stream, correct?

kreedula:
@iambmelt Yes. that's right

kreedula:
Thank you @iambmelt 👍

PR #1284, opened 2018-08-01
Added support to read vsts access token from env variables for Nightly builds

kreedula:

### DIFF
--- a/adal/build.gradle
+++ b/adal/build.gradle
@@ -250,8 +250,8 @@ publishing {
             name "vsts-maven-adal-android"
             url "https://identitydivision.pkgs.visualstudio.com/_packaging/AndroidADAL/maven/v1"
             credentials {
-                username System.getenv("VSTS_ENV_USER_NAME") != null ? System.getenv("VSTS_ENV_USER_NAME") : project.findProperty("vstsUsername")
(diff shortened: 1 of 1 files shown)


adal/build.gradle
```diff
@@ -250,8 +250,8 @@ publishing {
             name "vsts-maven-adal-android"
             url "https://identitydivision.pkgs.visualstudio.com/_packaging/AndroidADAL/maven/v1"
             credentials {
-                username project.findProperty("vstsUsername")
-                password project.findProperty("vstsMavenAccessToken")
+                username System.getenv("VSTS_ENV_USER_NAME") != null ? System.getenv("VSTS_ENV_USER_NAME") : project.findProperty("vstsUsername")
```
iambmelt: How do you feel about giving this a 'schema', in case more artifact streams are added later?

As an idea:
`ENV_<host>_<stream_type>_<stream_name>_<credential_type>`

In this case, something like:
`ENV_VSTS_MVN_ANDROIDADAL_USERNAME`
`ENV_VSTS_MVN_ANDROIDADAL_ACCESSTOKEN`

kreedula: Good suggestion @iambmelt 👍

PR #1287, opened 2018-08-07
AzureActiveDirectoryAuthorizationRequest constructor modification

kreedula:


iambmelt:
@kreedula Conceptually, these changes look OK to me. Once the test is fixed, I think we should be in good shape --  I suspect the test needs to be updated with a redirectUri.

kreedula:
Thanks @iambmelt . with the change in common to remove validation check for redirect uri , the tests are passing now

Issue #1288, opened 2018-08-09
Fix incorrect Exception returned on IOException in Discovery.java


kreedula:
Please ensure the following information is provided when logging a new issue:
ADAL throws DEVELOPER_AUTHORITY_IS_NOT_VALID_INSTANCE when it encounters IOException in performInstanceDiscovery() in Discovery.java.
We should break it up into SockettimeoutException and IOException,

Excepted Behavior : ADAL throws the same error that is thrown when network errors are hit.

kreedula:
Fixed via https://github.com/AzureAD/azure-activedirectory-library-for-android/pull/1289

addev-ashish:
The fix does not address the developer experience around this issue. we should catch sockettimeoutexception and return the network error that we return when network is not available

iambmelt:
Closing as fixed

PR #1296, opened 2018-08-15
Fix incorrect exception handling in Discovery.java

kreedula:
Issue : https://github.com/AzureAD/azure-activedirectory-library-for-android/issues/1288

Also made changes to fix adal compilation issues with common

Updated submodule to latest commit on dev

### DIFF
--- a/adal/src/main/java/com/microsoft/aad/adal/Discovery.java
+++ b/adal/src/main/java/com/microsoft/aad/adal/Discovery.java
@@ -262,8 +262,8 @@ class Discovery {
             Logger.e(TAG + methodName, "Error when validating authority. ", "", ADALError.DEVELOPER_AUTHORITY_IS_NOT_VALID_INSTANCE, e);
             throw new AuthenticationException(ADALError.DEVELOPER_AUTHORITY_IS_NOT_VALID_INSTANCE, e.getMessage(), e);
         } catch (SocketTimeoutException e){
-            Logger.e(TAG + methodName, "Error when validating authority. ", "", ADALError.DEVICE_CONNECTION_IS_NOT_AVAILABLE, e);
-            throw new AuthenticationException(ADALError.DEVICE_CONNECTION_IS_NOT_AVAILABLE, e.getMessage(), e);
+            Logger.e(TAG + methodName, "Error when validating authority. ", "", ADALError.SOCKET_TIMEOUT_EXCEPTION, e);
+            throw new AuthenticationException(ADALError.SOCKET_TIMEOUT_EXCEPTION, e.getMessage(), e);
         } catch (IOException e){
             Logger.e(TAG + methodName, "Error when validating authority. ", "", ADALError.IO_EXCEPTION, e);
             throw new AuthenticationException(ADALError.IO_EXCEPTION, e.getMessage(), e);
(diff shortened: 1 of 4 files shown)


adal/src/main/java/com/microsoft/aad/adal/Discovery.java
```diff
@@ -257,9 +258,15 @@ private void performInstanceDiscovery(final URL authorityUrl, final String trust
...
             }
             result = AuthorityValidationMetadataCache.isAuthorityValidated(authorityUrl);
-        } catch (final IOException | JSONException e) {
-            Logger.e(TAG + methodName, "Error when validating authority. ", "", ADALError.DEVELOPER_AUTHORITY_IS_NOT_VALID_URL, e);
+        } catch (JSONException e) {
+            Logger.e(TAG + methodName, "Error when validating authority. ", "", ADALError.DEVELOPER_AUTHORITY_IS_NOT_VALID_INSTANCE, e);
             throw new AuthenticationException(ADALError.DEVELOPER_AUTHORITY_IS_NOT_VALID_INSTANCE, e.getMessage(), e);
+        } catch (SocketTimeoutException e){
+            Logger.e(TAG + methodName, "Error when validating authority. ", "", ADALError.DEVICE_CONNECTION_IS_NOT_AVAILABLE, e);
+            throw new AuthenticationException(ADALError.DEVICE_CONNECTION_IS_NOT_AVAILABLE, e.getMessage(), e);
```
iambmelt: Let's create an entirely new error for this, such as: `SOCKET_TIMEOUT`

`DEVICE_CONNECTION_IS_NOT_AVAILABLE` is same `AdalError` that will be returned by `HttpWebRequest#throwIfNetworkNotAvailable` -- while these circumstances both indicate some kind of network connectivity trouble, the actual error mode [can be quite different](https://stackoverflow.com/questions/7360520/connectiontimeout-versus-sockettimeout).

In the case of socket timeouts, we want to track (via logging/telemetry) that the device _did_ have a connection, but that it was the Idp/STS who 'dropped' the connection

kreedula:
Updated to latest common and Made changes to fix adal compilation issues with common

adal/src/main/java/com/microsoft/aad/adal/TokenCacheAccessor.java
```diff
@@ -339,13 +339,12 @@ void updateTokenCacheUsingCommonCache(final AuthenticationRequest request, final
...
         AzureActiveDirectoryOAuth2Strategy strategy = ad.createOAuth2Strategy(config);
-        AzureActiveDirectoryAuthorizationRequest aadAuthRequest = new AzureActiveDirectoryAuthorizationRequest(
-                null, request.getClientId(), request.getRedirectUri(), null, request.getResource(),
-                new URL(mAuthority), request.getLoginHint(), request.getCorrelationId(), null,
-                request.getExtraQueryParamsAuthentication(), null,
-                request.getResource(), null, request.getClaimsChallenge());
-
-        mCommonCache.saveTokens(strategy, aadAuthRequest, tokenResponse);
+        AzureActiveDirectoryAuthorizationRequest aadAuthRequest = new AzureActiveDirectoryAuthorizationRequest.Builder(
+                request.getClientId(),
```
iambmelt: See #1307

PR #1308, opened 2018-09-05
Public API changes to add claims for aqcuireTokenSilent calls for LLT

kreedula:
Added two new overloads for both async and sync acquire token calls


iambmelt:
@kreedula Does this contain the same changes as #1313?

kreedula:
@iambmelt , this was the initial PR just with Public API changes. This is no longer needed, closing it.

PR #1313, opened 2018-09-10
Adal changes to support claims on Acquire Token Silent

kreedula:
- Added two new API's for Acquire Token Silent calls
- Passing Claims parameter with token request as well if present
- Changes are made to Acquire Token call to use RT to get a new AT if claims parameter is present
- Unit tests to verify if the token request is constructed correctly with claims

### DIFF
--- a/adal/src/main/java/com/microsoft/aad/adal/AuthenticationContext.java
+++ b/adal/src/main/java/com/microsoft/aad/adal/AuthenticationContext.java
@@ -669,7 +669,32 @@ public class AuthenticationContext {
      */
     public AuthenticationResult acquireTokenSilentSync(String resource, String clientId, String userId)
             throws AuthenticationException, InterruptedException {
-        return acquireTokenSilentSync(resource, clientId, userId, false, EventStrings.ACQUIRE_TOKEN_SILENT_SYNC);
+        return acquireTokenSilentSync(resource, clientId, userId, false, null, EventStrings.ACQUIRE_TOKEN_SILENT_SYNC);
+    }
+
+    /**
+     * This is sync function. It will first look at the cache and automatically
+     * checks for the token expiration. Additionally, if no suitable access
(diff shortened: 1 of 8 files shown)


adal/src/telemetry/java/com/microsoft/aad/adal/EventStrings.java
```diff
@@ -172,12 +172,16 @@
 
     static final String ACQUIRE_TOKEN_SILENT_SYNC_FORCE_REFRESH = "13";
 
+    static final String ACQUIRE_TOKEN_SILENT_SYNC_CLAIMS_CHALLENGE = "15";
```
iambmelt: Has our API telemetry doc been updated with these values?

kreedula: Thanks @iambmelt . Updated the doc

adal/src/main/java/com/microsoft/aad/adal/AuthenticationContext.java
```diff
@@ -669,7 +669,32 @@ public void acquireToken(final String resource, final String clientId, @Nullable
...
+     * checks for the token expiration. Additionally, if no suitable access
+     * token is found in the cache, but refresh token is available, the function
+     * will use the refresh token automatically. This method will not show UI
+     * for the user. If prompt is needed, the method will return an exception
+     *
+     * @param resource required resource identifier.
+     * @param clientId required client identifier.
+     * @param userId   UserID obtained from
+     *                 {@link AuthenticationResult #getUserInfo()}
+     * @param claims   The claims challenge returned from middle tier service, will be added as query string
```
iambmelt: Can you document whether `claims` is `@Nullable` or not? What happens if I pass `""`?

kreedula:
Thanks @iambmelt, improved javadocs as recommended

PR #1314, opened 2018-09-17
Remove RC tag from release 1.15.1

kreedula:


iambmelt:
@kreedula If it's nothing to worry about, LMK and I'll approve

kreedula:
Thanks @iambmelt , looks like submodule update issue. updated common to 0.0.4 here

PR #1317, opened 2018-09-18
Added null check to fix findbugs issue

kreedula:
Find bugs is complaining when we try to assemble the build. Added a null check to fix


kreedula:
Thank you @iambmelt

PR #1328, opened 2018-09-25
Fix compile issue with latest common and update common

kreedula:


adal/src/main/java/com/microsoft/aad/adal/CoreAdapter.java
```diff
@@ -39,7 +38,7 @@ private CoreAdapter() {
     }
 
     /**
-     * Gets the supplied {@link UserInfo} as an {@link Account}.
+     * Gets the supplied {@link UserInfo} as an {@link AccountRecord}.
```
iambmelt: Oops. I made the same mistake you did while reading this PR -- this is the wrong class. This needs to be `BaseAccount`

kreedula: Thank you for pointing out @iambmelt . I pushed a PR with the fix

PR #1340, opened 2018-10-08
[1.15.1-hf1] Picks up common/#271 - updates imports

iambmelt:
See title.

/cc @titorodi


iambmelt:
Merge https://github.com/AzureAD/microsoft-authentication-library-common-for-android/pull/271

Then update this submodule pointer

Then merge this PR

adal/src/main/java/com/microsoft/aad/adal/AuthenticationContext.java
```diff
@@ -1422,7 +1422,7 @@ public static String getVersionName() {
         // Package manager does not report for ADAL
         // AndroidManifest files are not merged, so it is returning hard coded
         // value
-        return "1.15.1";
+        return "1.15.1-hf1";
```
titorodi: Fine for AAD / ESTS right? and their versioning? Since we shipped with -RC I assume this is fine as it would have blocked TrueMAM CA.

changelog.txt
```diff
@@ -1,3 +1,8 @@
+Version 1.15.1-hf1
```
titorodi: Hm, this will get merged later to dev and 1.15.2... so this will be accounted for I think?

adal/src/main/java/com/microsoft/aad/adal/AuthenticationContext.java
iambmelt: Not aware of any issues -- @addev-ashish we're cool, yeah?

changelog.txt
titorodi: Should it call out that it fixes support for certain ADFS CertAuthorities configurations? Otherwise it seems like a strange reason to hotfix.

iambmelt:
Thank you for the reviews @titorodi @kreedula @addev-ashish -- I'll merge this PR once `common` hits central

PR #1365, opened 2018-11-20
Fix the userID sharedpreference for V2 Broker

heidijinxujia:

### DIFF
--- a/adal/src/main/java/com/microsoft/aad/adal/AcquireTokenRequest.java
+++ b/adal/src/main/java/com/microsoft/aad/adal/AcquireTokenRequest.java
@@ -24,6 +24,7 @@
 package com.microsoft.aad.adal;
 
 
+import android.accounts.AccountManager;
 import android.content.Context;
 import android.content.Intent;
 import android.content.pm.PackageInfo;
@@ -36,6 +37,7 @@ import android.util.Log;
 
 import com.microsoft.identity.common.adal.internal.AuthenticationConstants;
 import com.microsoft.identity.common.adal.internal.util.StringExtensions;
+import com.microsoft.identity.common.internal.broker.BrokerResult;
 import com.microsoft.identity.common.internal.providers.microsoft.azureactivedirectory.AzureActiveDirectory;
 import com.microsoft.identity.common.internal.providers.microsoft.azureactivedirectory.AzureActiveDirectoryCloud;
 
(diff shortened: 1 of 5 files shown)


userappwithbroker/src/main/java/com/microsoft/aad/adal/example/userappwithbroker/Constants.java
```diff
@@ -79,7 +79,7 @@ public String getText() {
     }
 
     enum RedirectUri {
-        Regular("msauth://com.microsoft.aad.adal.userappwithbroker/IcB5PxIyvbLkbFVtBI%2FitkW%2Fejk%3D"),
+        Regular("msauth://com.microsoft.aad.adal.userappwithbroker/AyptwvRgqjbFju1%2FfsnGRmFOxhU%3D"),
```
iambmelt: Is this a mistake?

adal/src/main/java/com/microsoft/aad/adal/AcquireTokenRequest.java
```diff
@@ -686,6 +688,65 @@ private void verifyBrokerRedirectUri(final AuthenticationRequest request) throws
...
+                new AuthenticationResult(
+                        accessToken,
+                        null,
+                        expire,
+                        false,
+                        userinfo,
+                        tenantId,
+                        idtoken,
+                        null,
+                        ""
```
iambmelt: Does the `clientId` get set somewhere else? Why `""` and not `null`?

Also, it looks like the `clientId` may be available from `waitingRequest.getRequest().getClientId()` --  if that value is populated, shouldn't it passed into this method such that it can be used to construct the `AuthenticationResult`?

adal/src/main/java/com/microsoft/aad/adal/AcquireTokenRequest.java
```diff
@@ -686,6 +688,65 @@ private void verifyBrokerRedirectUri(final AuthenticationRequest request) throws
...
+        if (null != data.getExtras()) {
+            userinfo = UserInfo.getUserInfoFromBrokerResult(data.getExtras());
+        }
+
+        // grab the fields tracked by x-ms-clitelem
+        final String serverErrorCode = data.getStringExtra(AuthenticationConstants.Broker.CliTelemInfo.SERVER_ERROR);
+        final String serverSubErrorCode = data.getStringExtra(AuthenticationConstants.Broker.CliTelemInfo.SERVER_SUBERROR);
+        final String refreshTokenAge = data.getStringExtra(AuthenticationConstants.Broker.CliTelemInfo.RT_AGE);
+        final String speRingInfo = data.getStringExtra(AuthenticationConstants.Broker.CliTelemInfo.SPE_RING);
+//
```
iambmelt: Let's remove this if it's no longer needed

PR #1372, opened 2019-01-10
Deleted IBrokerAccountService aidl file from adal as it's moved to common

kreedula:
- As part of changes to make `ad-accounts` work with just `common`, ` IBrokerAccountService.aidl` has been moved to common so that it can be accessed in `ad-accounts`. Deleting this file from `adal` repo to avoid multidex issues if an app declares both `adal`  and `ad-accounts` as dependencies

- Updated submodule pointer to latest dev

- Added buildToolsVersion to adal and test apps to align with `common`'s 28.0.3


iambmelt:
@kreedula Build is now passing for me locally 👍

kreedula:
Thanks for validating @iambmelt . Second set of eyes always good

PR #1396, opened 2019-02-12
Improvements to SPE Ring Support w/ Broker

iambmelt:
Expands: https://github.com/AzureAD/microsoft-authentication-library-common-for-android/pull/370/

Adds SPE Ring / Client Telemetry Bundle parsing to failure cases.

### DIFF
--- a/adal/src/main/java/com/microsoft/aad/adal/AuthenticationActivity.java
+++ b/adal/src/main/java/com/microsoft/aad/adal/AuthenticationActivity.java
@@ -70,6 +70,11 @@ import java.security.cert.X509Certificate;
 import java.util.Locale;
 import java.util.UUID;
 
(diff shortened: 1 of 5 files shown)


adal/src/main/java/com/microsoft/aad/adal/AuthenticationActivity.java
```diff
@@ -1098,10 +1120,33 @@ protected void onPostExecute(TokenTaskResult result) {
                             userinfo.getDisplayableId());
                 }
 
+                if (null != result.mTaskResult.getCliTelemInfo()) {
```
kreedula: Just to confirm, these changes in this class are ADAL only related changes right? I mean broker has nothing to do with it ?

iambmelt: That's correct - that's for double checking.

PR #1410, opened 2019-02-27
updating minimum recommended version

danieldobalian:
Updating min recommended version

### DIFF
--- a/README.md
+++ b/README.md
@@ -17,9 +17,9 @@ Note: A corpnet account is required to view the VSTS build.
 
 ## Versions
 
-Current version - 1.16.1
+Current version - 1.16.+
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -19,7 +19,7 @@ Note: A corpnet account is required to view the VSTS build.
 
 Current version - 1.16.1
 
-Minimum recommended version - 1.1.16
+Minimum recommended version - 1.15.0
```
iambmelt: `1.15.0` isn't a bad choice, if we can set this value free of any other considerations, we _could_ choose `1.15.1-hf1` which shipped with better sovereign support and a fix for PkeyAuth

danieldobalian: Good call, changed to `1.15.+`.

PR #1438, opened 2019-06-05
Replace brokerRedirectPrint powershell script

cofitz1:
Replace brokerRedirectPrint powershell script with one that only requires the APK as a parameter.  The current version requires the user to make many manual script updates and to provide access to the keystore files, aliases, and passwords.  We can leverage `aapt` and `keytool` to obtain the required information for building the redirectURI.

If multiple certificates are found the owner information is also displayed so that the user can pick the correct one.

@titorodi 
@microsoftjoe 
@shoatman


GetBrokerRedirectURI.ps1
```diff
@@ -0,0 +1,103 @@
...
+    $signatures = @{}
+    $owner = ""
+    ForEach ($line in $output) {
+        if ($line.contains($ownerPrefix)) {
+            $owner = $line.replace($ownerPrefix, "").trim()
+        }
+
+        if ($line.contains($shaPrefix)) {
+            $base64 = GetBase64FromSHA1Hex($line.replace($shaPrefix, "").trim())
+            $signatures.add($owner, $base64)
```
microsoftjoe: this assumes that the owner line always comes before the sha1 line -- is that always true, even when there are multiples?

cofitz1: Yup - I confirmed with an APK that has this scenario.  The owner always appears before the signatures.

GetBrokerRedirectURI.ps1
```diff
@@ -0,0 +1,103 @@
...
+
+<#
+Extract the APK package name from `aapt dump badging`.
+#>
+function GetPackageName {
+    Param(
+        [string]$apkFile
+    )
+
+    $output = aapt dump badging $apkFile
```
shoatman: What assumptions is the script making about the availability of tools to the shell?  You may want to add the following:
`get-command aapt
if(!$?){
 write-error "The appt command was not found.  Please add your android build tools to the path"
}
`
$? is an automatic variable which indicates whether the last command was successful or not.

GetBrokerRedirectURI.ps1
```diff
@@ -0,0 +1,103 @@
...
+)
+
+<#
+Extract all of the signature owner and SHA1 hex bytes from `keytool -printcert -jarfile`.
+#>
+function GetSHA1Hex {
+    Param(
+        [string]$apkFile
+        )
+    $output = keytool -printcert -jarfile $apkFile
```
shoatman: Same comment here relative to use of a command.  Verify it's available.

GetBrokerRedirectURI.ps1
```diff
@@ -0,0 +1,103 @@
...
+        }
+    }
+    
+    if ($signatures.Keys.count -eq 0) {
+        write-error "No signatures found."
+        exit
+    }
+    
+    if ($signatures.Keys.count -gt 1) {
+        write-warning "Multiple signatures detected, please select the correct one."
```
shoatman: Is there enough information provided by the output to make this determination?

cofitz1: I believe that it is sufficient - I can add some more info on what to do in the warning though

GetBrokerRedirectURI.ps1
cofitz1: Adding `VerifyDependencies` function that checks if get-command is successful

GetBrokerRedirectURI.ps1
cofitz1: see other thread

PR #1447, opened 2019-08-06
AndroidX migration

rpdome:
This is done to unblock android_complete. (AndroidX is a project-wide settings, and therefore requires every sub-project to be migrated).


iambmelt:
Tests are currently failing but I am OK with merging for now and fixing the tests in a follow-up PR.

@rpdome @kreedula any thoughts?

rpdome:
adal test cases are now running locally. Travis still breaks as Mockito still complains.
automation/automationtestapp still breaks - decide not to touch it for now.

PR #1448, opened 2019-08-07
Update compile Travis version 27 -> 28

iambmelt:
See title + diff, unblocks #1447


iambmelt:
Thanks @kreedula @rpdome - I'll merge this PR into yours for now to unblock the AndroidX migrations for dev but we should take a TODO to fix the tests for real

PR #1449, opened 2019-08-09
Update submodule to latest, Fix CI build

iambmelt:
Work in progress.

Goals of this PR:
- Fix unit tests to use `androidx testrunner`
- Update submodule to `common/dev` latest

### DIFF
--- a/adal/src/androidTest/java/com/microsoft/aad/adal/AuthenticationContextTest.java
+++ b/adal/src/androidTest/java/com/microsoft/aad/adal/AuthenticationContextTest.java
@@ -35,7 +35,6 @@ import android.os.Build;
 import android.os.Bundle;
 import android.util.Base64;
 
-import androidx.test.InstrumentationRegistry;
 import androidx.test.annotation.UiThreadTest;
 import androidx.test.core.app.ApplicationProvider;
 import androidx.test.ext.junit.runners.AndroidJUnit4;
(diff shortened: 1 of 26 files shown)


adal/src/androidTest/java/com/microsoft/aad/adal/AuthenticationContextTest.java
```diff
@@ -33,10 +33,12 @@
 import android.content.pm.Signature;
 import android.os.Build;
 import android.os.Bundle;
+import android.util.Base64;
+
 import androidx.test.InstrumentationRegistry;
```
rpdome: given that we seems to be using androidx.test.platform.app.InstrumentationRegistry, can we remove this for good?

rpdome: ps. I have zero idea why there are InstrumentationRegistry in two packages.

iambmelt: Yeah, missed this import - I'll run the tool

gradle/versions.gradle
```diff
@@ -8,7 +8,7 @@ ext {
     buildToolsVersion = "28.0.3"
 
     // Plugins
-    gradleVersion = '3.2.1'
+    gradleVersion = '3.3.1'
```
rpdome: Was going to do this. Nice catch. Where can I add thumbs up emoji.

adal/src/androidTest/java/com/microsoft/aad/adal/AuthenticationContextTest.java
iambmelt: Resolved in https://github.com/AzureAD/azure-activedirectory-library-for-android/pull/1449

gradle/versions.gradle
iambmelt: Right here! 💯

PR #1467, opened 2019-10-18
Update submodule, update ADFSv2 Client Id for Lab 2 Test Fixture, Fix Tests (logger based)

iambmelt:
See title.

Also Fix logger based tests in ADAL (See comments below)

### DIFF
--- a/adal/src/androidTest/java/com/microsoft/aad/adal/AuthenticationActivityUnitTest.java
+++ b/adal/src/androidTest/java/com/microsoft/aad/adal/AuthenticationActivityUnitTest.java
@@ -497,6 +497,12 @@ public class AuthenticationActivityUnitTest {
                     e.printStackTrace();
                 }
 
+                try {
+                    Thread.sleep(1000);
+                } catch (InterruptedException e) {
+                    e.printStackTrace();
+                }
+
                 // get field value to check
                 assertTrue("verify log message",
                         logResponse.getMessage().contains("Webview onResume register broadcast"));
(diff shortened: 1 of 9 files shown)


iambmelt:
Build is failing
>* What went wrong:
A problem occurred evaluating project ':common'.
> Cannot get property 'robolectricVersion' on extra properties extension as it does not exist

@shahzaibj if you have the cycles, feel free to address

shahzaibj:
> Build is failing
> 
> > * What went wrong:
> >   A problem occurred evaluating project ':common'.
> >   Cannot get property 'robolectricVersion' on extra properties extension as it does not exist
> 
> @shahzaibj if you have the cycles, feel free to address

Thanks @iambmelt, looking into it.

shahzaibj:
@iambmelt I've fixed issues related to robolectric, however, the build is still failing as there seem to be a lot of other errors (warnings) related to unchecked calls and other things that are occurring in various files in `common` (thrown as adal does a build of common). Let's take a look at this tomorrow to see why this is happening.

shahzaibj:
@iambmelt Travis build is showing 50 warnings related to unchecked etc. but that is not the reason the build is failing, the build is failing because a few of the tests are failing specifically in the `AuthenticationContextTest` class. I believe some tests got broken when adal is updated with the submodule.

shahzaibj:
The tests that are broken are the ones that use the logger to assert for specific messages, and these test are now broken as logging is now performed on a background task. This change was implemented in [common PR 632](https://github.com/AzureAD/microsoft-authentication-library-common-for-android/pull/632)

shahzaibj:
@iambmelt Read my previous comment to learn why test are broken. 

I've fixed them by adding a thread sleep before we call assert on a log object. This may not be the best fix but it works, other options would be to look into moving all these tests to Robolectric or find some other way of testing these scenarios without using logger (or any background thread based object).

In my opinion, this fix is sufficient and we can increase the thread sleep if needed.

PR #1473, opened 2019-10-30
Version update to 1.16.3-hf1

kreedula:


iambmelt:
@shahzaibj Yes, this is our hotfix postfix -- in the event we hotfix a hotfix, they are each assigned a number. So a hotfix of `1.16.3-hf1` would be `1.16.3-hf2`

PR #1481, opened 2019-11-16
Fix test app, pass authority in silent response

rpdome:
- In test app, we should always be saving the authority, not idp.
- Silent call should return authority.


rpdome:
this fix is also part of #1504

PR #1498, opened 2020-02-27
Fix for sovereign caching authority mismatch

iambmelt:
This PR corrects an issue whereby the ADAL cache's authority property is not updated to match the `cloud_instance_hostname` returned from sovereign HRD

### DIFF
--- a/adal/src/main/java/com/microsoft/aad/adal/TokenCacheAccessor.java
+++ b/adal/src/main/java/com/microsoft/aad/adal/TokenCacheAccessor.java
@@ -29,12 +29,12 @@ import com.microsoft.identity.common.adal.internal.AuthenticationConstants;
 import com.microsoft.identity.common.adal.internal.cache.StorageHelper;
 import com.microsoft.identity.common.adal.internal.util.StringExtensions;
 import com.microsoft.identity.common.internal.cache.ADALOAuth2TokenCache;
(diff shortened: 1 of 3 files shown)


adal/src/main/java/com/microsoft/aad/adal/TokenCacheAccessor.java
```diff
@@ -306,11 +304,17 @@ void updateCachedItemWithResult(final AuthenticationRequest request, final Authe
     /**
      * Update token cache with returned auth result.
      */
-    void updateTokenCache(final AuthenticationRequest request,  final AuthenticationResult result) throws MalformedURLException {
+    void updateTokenCache(final AuthenticationRequest request, final AuthenticationResult result) throws MalformedURLException {
         if (result == null || StringExtensions.isNullOrBlank(result.getAccessToken())) {
             return;
         }
 
+        if (!StringExtensions.isNullOrBlank(result.getAuthority())
```
cofitz1: Should there also be a test update?

Om83: +1

iambmelt: Not so much an "update" because no tests exist for this class _at all_.

However, I've taken the liberty of adding a test in 4c76469f479696f115ac4d734c8a2790cd417d54

PR #1500, opened 2020-02-27
Added debug broker redirect uri; set default client id; authority alw…

shoatman:
- Made adfsv4 the default client id
- Commented out code for determining authority based on the looked up user from shared preferences. Code now always uses the specified authority (UI drop down)
- Added LABSDEBUG redirect URI which matches the signature from the debug.keystore included with android-complete


rpdome:
I need to make some change on top of this for sov cloud testing, so I'll merge it now.

PR #1504, opened 2020-02-28
Add 'other authority' to test app for testing sovereign cloud scenario.

rpdome:
to make sure we can use the authority returned from an instance_aware call is usable in silent flow.


shahzaibj:
It is not clear to me how the authority here is coming from instance_aware, it seems to coming from request options

rpdome:
@shahzaibj ADAL requires the user to save the authority from the interactive call (instance_aware), and then use it in the silent call.

PR #1505, opened 2020-02-28
Bumping to 2.0.0-RC1

rpdome:
We should not return user cancel.


rpdome:
Dropping the change. This could be a server side regression.

PR #1515, opened 2020-03-13
Added getBrokerUsers call before silent to get uid based on login_hint

kreedula:
To get this test working : https://identitydivision.visualstudio.com/Engineering/_workitems/edit/926501

### DIFF
--- a/userappwithbroker/src/main/java/com/microsoft/aad/adal/example/userappwithbroker/MainActivity.java
+++ b/userappwithbroker/src/main/java/com/microsoft/aad/adal/example/userappwithbroker/MainActivity.java
@@ -192,14 +192,19 @@ public class MainActivity extends AppCompatActivity implements NavigationView.On
                                 break;
                             }
                         }
+
+                        if(!TextUtils.isEmpty(uId)) {
+                            callAcquireTokenSilent(requestOptions.getDataProfile().getText(),
+                                    uId,
+                                    requestOptions.getClientId().getText());
+                        }else {
+                            showMessage("No uId matching the provided upn, cannot proceed with silent call");
+                        }
+
                     } catch (OperationCanceledException | AuthenticatorException | IOException e) {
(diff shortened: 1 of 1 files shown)


userappwithbroker/src/main/java/com/microsoft/aad/adal/example/userappwithbroker/MainActivity.java
```diff
@@ -170,9 +177,32 @@ public void onAcquireTokenSilentClicked(final AcquireTokenFragment.RequestOption
...
+                        final UserInfo[] userInfos = mAuthContext.getBrokerUsers();
+
+                        for (UserInfo userInfo : userInfos) {
+                            if (userInfo.getDisplayableId().equalsIgnoreCase(requestOptions.getLoginHint())) {
+                                uId = userInfo.getUserId();
+                                break;
+                            }
+                        }
+                    } catch (OperationCanceledException | AuthenticatorException | IOException e) {
+                        Logger.e(TAG, "Unable to getbrokerusers from broker " + e.getMessage());
```
iambmelt: Can we `Toast` this message?

If this fails and its not hooked up to a debugger, it'll just look like nothing happened

iambmelt: Do we still want to try the silent call if `getBrokerUsers` fails like this? Seems like we should check for a non-error result before proceeding

kreedula: Only proceeding with silent if uId is non-null else showing a toast

shahzaibj:
Did this test case ever used to work? If yes, then how?

kreedula:
> Did this test case ever used to work? If yes, then how?

Looking at the code, i don't think it could have ever worked

PR #1518, opened 2020-03-18
Merge release 3.0.0 to  master

kreedula:

### DIFF
--- a/userappwithbroker/build.gradle
+++ b/userappwithbroker/build.gradle
@@ -70,7 +70,7 @@ dependencies {
     implementation "com.google.android.material:material:$rootProject.ext.materialVersion"
 
     localImplementation(project(':adal'))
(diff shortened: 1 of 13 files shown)


userappwithbroker/build.gradle
```diff
@@ -70,7 +70,7 @@ dependencies {
     implementation "com.google.android.material:material:$rootProject.ext.materialVersion"
 
     localImplementation(project(':adal'))
-    distImplementation("com.microsoft.aad:adal:1.16.3") {
+    distImplementation("com.microsoft.aad:adal:1.16.3-hf2") {
```
shahzaibj: Should we use the latest here too?

kreedula: guess so, but let's defer it for now, we can make a change to dev, since this is a merge PR.

PR #1520, opened 2020-03-18
add dual screen support

rpdome:
Also change from ProgressDialog to ProgressBar, because there is no easier/cleaner way to change the position of the ProgressDialog in dual screen.


shahzaibj:
@rpdome The build is failing. Looks like one of the tests is referencing to the older webview resource id.

`adal/src/androidTest/java/com/microsoft/aad/adal/AuthenticationActivityUnitTest.java:166: error: cannot find symbol
        final WebView webview = mActivityRule.getActivity().findViewById(R.id.webView1);`

PR #1521, opened 2020-03-20
Return ADALError.MDM_REQUIRED after the user interacts with MDM link

rpdome:
Server will attach a query string (ismdmurl=1) along with the browser link to tell us that this link belongs to Intune (MDM) flow.

WebView will detect that query string and return AuthenticationConstants.UIResponse.BROWSER_CODE_MDM (instead of BROWSER_CODE_CANCEL)

### DIFF
--- a/adal/src/main/java/com/microsoft/aad/adal/BasicWebViewClient.java
+++ b/adal/src/main/java/com/microsoft/aad/adal/BasicWebViewClient.java
@@ -38,12 +38,14 @@ import androidx.annotation.NonNull;
 import androidx.annotation.Nullable;
 
 import com.microsoft.aad.adal.ChallengeResponseBuilder.ChallengeResponse;
+import com.microsoft.identity.common.adal.internal.AuthenticationConstants;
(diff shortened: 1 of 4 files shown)


adal/src/main/java/com/microsoft/aad/adal/BasicWebViewClient.java
```diff
@@ -56,6 +57,8 @@
 import static com.microsoft.aad.adal.AuthenticationConstants.OAuth2.CODE;
 import static com.microsoft.aad.adal.AuthenticationConstants.UIResponse.BROWSER_CODE_AUTHENTICATION_EXCEPTION;
 import static com.microsoft.aad.adal.AuthenticationConstants.UIResponse.BROWSER_CODE_ERROR;
+import static com.microsoft.identity.common.adal.internal.AuthenticationConstants.Broker.BROWSER_DEVICE_CA_URL_QUERY_STRING_PARAMETER;
```
Om83: This looks to be a verbose import, why can't we just import AuthenticationConstants?

PR #1528, opened 2020-04-09
Update readme

rpdome:


README.md
```diff
@@ -129,6 +129,10 @@ You can get the binaries from Maven central repo. AAR package can be included as
 ```gradle
 repositories {
     mavenCentral()
+    maven {
+        // Surface Duo SDK dependency.
+        url 'https://pkgs.dev.azure.com/MicrosoftDeviceSDK/DuoSDK-Public/_packaging/Duo-SDK-Feed%40Local/maven/v1'
```
iambmelt, line 134: We might want to call out that this is preview

iambmelt: And that including this is required in your `build.gradle` is (or is not) required as of ADAL version x.x.x

rpdome: This is their public feed - and it's not just for preview. (future versions will be published here too. AFAIK there is no plan to publish them to maven central).

iambmelt: Good to know. Thanks @rpdome

PR #1537, opened 2020-06-19
Add SAML assertion

paivybhav:
ADAL SDK on Android does not support silent token acquision based on SAML assertion, which is needed for MSA guest in AAD tenant related use cases.

Changes made :
* new method acquireTokenSilentSyncWithAssertion in AuthenticaitonContext which will be used by developers
* new constants needed for creating saml assertion based requests in AuthenticationConstants
* added methods to acquireTokenRequest, AcquireTokenSilentHandler and OAuth2 for new code flow 
* Changed the Authentication Request class for passing the assertion
* new apiEvent String for saml assertion
* new test case in AuthenticationContextTest for verification

### DIFF
--- a/adal/src/main/java/com/microsoft/aad/adal/AcquireTokenRequest.java
+++ b/adal/src/main/java/com/microsoft/aad/adal/AcquireTokenRequest.java
@@ -466,6 +466,7 @@ class AcquireTokenRequest {
             throws AuthenticationException {
 
         final boolean requestEligibleForBroker = mBrokerProxy.verifyBrokerForSilentRequest(authenticationRequest);
+        final String methodName = ":acquireTokenSilentFlow";
 
         //1. if forceRefresh == true OR claimsChallenge is not null AND the request is eligible for the broker
         if ((authenticationRequest.getForceRefresh() || authenticationRequest.isClaimsChallengePresent()) && requestEligibleForBroker) {
@@ -479,11 +480,14 @@ class AcquireTokenRequest {
         }
 
         //3. try SAML Assertion
-        if (authenticationRequest.getSamlAssertion() != null) {
+        if (authenticationRequest.getSamlAssertion() != null && authenticationRequest.getAssertionType() != null) {
             final AuthenticationResult authResultFromSaml = tryAcquireTokenSilentWithAssertion(authenticationRequest);
             if (isAccessTokenReturned(authResultFromSaml)) {
+                Logger.v(TAG + methodName, "Access token has been acquired using the saml assertion.");
                 return authResultFromSaml;
-            }            
+            } else {
+                Logger.w(TAG + methodName, "Failed to acquire tokens using saml assertion.");
+            }
         }
 
         //4. We couldn't get locally...If eligible return via broker... otherwise return local result
--- a/adal/src/main/java/com/microsoft/aad/adal/AcquireTokenRequest.java
+++ b/adal/src/main/java/com/microsoft/aad/adal/AcquireTokenRequest.java
@@ -458,8 +458,9 @@ class AcquireTokenRequest {
 
     /**
      * Handles the silent flow. Will always lookup local cache. If there is a valid AT in local cache, will use it. If
-     * AT in local cache is already expired, will try RT in the local cache. If RT requst failed, and if we can switch
-     * to broker for auth, will switch to broker for authentication.
+     * AT in local cache is already expired, will try RT in the local cache. If RT requst failed, then use saml assertion passed in the 
+     * request to acquire RT and AT. If this too fails and if we can switch to broker for auth, 
+     * will switch to broker for authentication.
      */
     private AuthenticationResult acquireTokenSilentFlow(final AuthenticationRequest authenticationRequest)
             throws AuthenticationException {
@@ -477,7 +478,15 @@ class AcquireTokenRequest {
             return authResult;
         }
 
-        //3. We couldn't get locally...If eligible return via broker... otherwise return local result
+        //3. try SAML Assertion
+        if(authenticationRequest.getSamlAssertion()){
+            final AuthenticationResult authResultFromSaml = tryAcquireTokenSilentWithAssertion(authenticationRequest);
+            if (isAccessTokenReturned(authResultFromSaml)) {
+                return authResultFromSaml;
+            }            
+        }
+
+        //4. We couldn't get locally...If eligible return via broker... otherwise return local result
         if (requestEligibleForBroker) {
             return tryAcquireTokenSilentWithBroker(authenticationRequest);
         } else {
@@ -498,6 +507,21 @@ class AcquireTokenRequest {
 
         return acquireTokenSilentHandler.getAccessToken();
     }
+
+    /**
+     * Try acquire token using saml assertion.
+     */
+    private AuthenticationResult tryAcquireTokenSilentWithAssertion(final AuthenticationRequest authenticationRequest)
+            throws AuthenticationException {
+        final String methodName = ":tryAcquireTokenSilentWithAssertion";
+        Logger.v(TAG + methodName, "Try to silently get token using SAML Assertion.");
+        final AcquireTokenSilentHandler acquireTokenSilentHandler = new AcquireTokenSilentHandler(mContext,
+                authenticationRequest, mTokenCacheAccessor);
+        removeTokensForUser(authenticationRequest);
+
+        return acquireTokenSilentHandler.getAccessTokenUsingAssertion();
+    }
+
 
     /**
      * Try acquire token silent with broker.
--- a/adal/src/main/java/com/microsoft/aad/adal/AcquireTokenRequest.java
+++ b/adal/src/main/java/com/microsoft/aad/adal/AcquireTokenRequest.java
@@ -479,7 +479,7 @@ class AcquireTokenRequest {
         }
 
         //3. try SAML Assertion
-        if(authenticationRequest.getSamlAssertion() != null) {
+        if (authenticationRequest.getSamlAssertion() != null) {
             final AuthenticationResult authResultFromSaml = tryAcquireTokenSilentWithAssertion(authenticationRequest);
             if (isAccessTokenReturned(authResultFromSaml)) {
                 return authResultFromSaml;
@@ -520,7 +520,6 @@ class AcquireTokenRequest {
         
         return acquireTokenSilentHandler.getAccessTokenUsingAssertion();
     }
-
 
     /**
      * Try acquire token silent with broker.
--- a/adal/src/main/java/com/microsoft/aad/adal/AcquireTokenRequest.java
+++ b/adal/src/main/java/com/microsoft/aad/adal/AcquireTokenRequest.java
@@ -479,7 +479,7 @@ class AcquireTokenRequest {
         }
 
         //3. try SAML Assertion
-        if(authenticationRequest.getSamlAssertion()){
+        if(authenticationRequest.getSamlAssertion() != null) {
             final AuthenticationResult authResultFromSaml = tryAcquireTokenSilentWithAssertion(authenticationRequest);
             if (isAccessTokenReturned(authResultFromSaml)) {
                 return authResultFromSaml;
@@ -517,8 +517,7 @@ class AcquireTokenRequest {
         Logger.v(TAG + methodName, "Try to silently get token using SAML Assertion.");
         final AcquireTokenSilentHandler acquireTokenSilentHandler = new AcquireTokenSilentHandler(mContext,
                 authenticationRequest, mTokenCacheAccessor);
-        removeTokensForUser(authenticationRequest);
-
+        
         return acquireTokenSilentHandler.getAccessTokenUsingAssertion();
     }
 
--- a/adal/src/main/java/com/microsoft/aad/adal/AcquireTokenRequest.java
+++ b/adal/src/main/java/com/microsoft/aad/adal/AcquireTokenRequest.java
@@ -345,7 +345,7 @@ class AcquireTokenRequest {
         // 1) throw AuthenticationException if
         //    a) No access token is returned from local flow and no prompt is allowed
         //    b) Server error or network errors when sending post request to token endpoint with grant_type
-        //       as refresh_token in local flow
+        //       as refresh_token in local flow or with grant type as assertion
         //    c) Broker returns ERROR_CODE_BAD_ARGUMENTS, ACCOUNT_MANAGER_ERROR_CODE_BAD_AUTHENTICATION
         //       or ERROR_CODE_UNSUPPORTED_OPERATION
         // 2) Non-null AuthenticationResult if we could try silent request, and silent request either successfully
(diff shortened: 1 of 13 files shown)


adal/src/main/java/com/microsoft/aad/adal/AcquireTokenSilentHandler.java
```diff
@@ -118,6 +118,78 @@ AuthenticationResult getAccessToken() throws AuthenticationException {
...
 
+    AuthenticationResult getAccessTokenUsingAssertion() throws AuthenticationException {
+        final String methodName = ":getAccessTokenUsingAssertion";
+        final AuthenticationResult result = acquireTokenWithAssertion();
+        
+        if (result != null && !StringExtensions.isNullOrBlank(result.getAccessToken())) {
+            try {
+                mTokenCacheAccessor.updateTokenCache(mAuthRequest, result);
+            } catch (MalformedURLException e) {
+                Logger.v(TAG + methodName, "Access token fetched but unable to update token cache");
```
iambmelt, line 129: Let's log this error as a warning, instead of as verbose

adal/src/main/java/com/microsoft/aad/adal/AcquireTokenSilentHandler.java
```diff
@@ -118,6 +118,78 @@ AuthenticationResult getAccessToken() throws AuthenticationException {
...
+                    "Request: " + mAuthRequest.getLogInfo()
+                            + " " + ExceptionExtensions.getExceptionMessage(exc)
+                            + " " + Log.getStackTraceString(exc),
+                    ADALError.AUTH_FAILED_NO_TOKEN,
+                    null);
+
+            throw new AuthenticationException(
+                    ADALError.AUTH_FAILED_NO_TOKEN, ExceptionExtensions.getExceptionMessage(exc),
+                    new AuthenticationException(ADALError.SERVER_ERROR, exc.getMessage(), exc));
+        } catch (final IOException | AuthenticationException exc) {
```
iambmelt, line 175: The `catch` block for the `IOExcepton`, `AuthenticationException` appears to contain the same logic as `ServerRespondingWithRetryableException`; is there any reason we should not collapse these into the same `catch` block?

adal/src/main/java/com/microsoft/aad/adal/AcquireTokenSilentHandler.java
```diff
@@ -118,6 +118,78 @@ AuthenticationResult getAccessToken() throws AuthenticationException {
...
+        final String samlAssertion = mAuthRequest.getSamlAssertion();
+        final String assertionType = mAuthRequest.getAssertionType();
+
+        try {
+            final JWSBuilder jwsBuilder = new JWSBuilder();
+            final Oauth2 oauthRequest = new Oauth2(mAuthRequest, mWebRequestHandler, jwsBuilder);
+
+            result = oauthRequest.refreshTokenUsingAssertion(samlAssertion, assertionType);
+            if (result != null && StringExtensions.isNullOrBlank(result.getRefreshToken())) {
+                Logger.i(TAG + methodName, "Refresh token is not returned or empty", "");
```
iambmelt, line 158: This seems more like a `warning` than an `info` statement, right?

adal/src/main/java/com/microsoft/aad/adal/Oauth2.java
```diff
@@ -544,6 +571,34 @@ public AuthenticationResult refreshToken(String refreshToken) throws IOException
         return postMessage(requestMessage, headers);
     }
 
+    public AuthenticationResult refreshTokenUsingAssertion(String samlAssertion, String assertionType)
```
iambmelt, line 574: Can we add nullability (`@Nullable` and/or `@NonNull`) and `final` on the parameters to this method?

Nullability annotations should be applied based on whether you expect/support providing `null` as params to this function

adal/src/main/java/com/microsoft/aad/adal/AcquireTokenRequest.java
```diff
@@ -477,7 +478,15 @@ private AuthenticationResult acquireTokenSilentFlow(final AuthenticationRequest
...
         }
 
-        //3. We couldn't get locally...If eligible return via broker... otherwise return local result
+        //3. try SAML Assertion
+        if (authenticationRequest.getSamlAssertion() != null) {
+            final AuthenticationResult authResultFromSaml = tryAcquireTokenSilentWithAssertion(authenticationRequest);
+            if (isAccessTokenReturned(authResultFromSaml)) {
+                return authResultFromSaml;
+            }            
+        }
```
iambmelt, line 487: This behavior where we fall-thru and try the broker isn't 100% clear to me... want to confirm....

If the caller has supplied a SAML assertion in their request, and no access token was returned are we falling-thru in hopes that the broker has an RT/FRT/PRT for this user already?

Looking to confirm that the decision to proceed here and attempt the request in the broker in intentional

adal/src/main/java/com/microsoft/aad/adal/AcquireTokenRequest.java
```diff
@@ -477,7 +478,15 @@ private AuthenticationResult acquireTokenSilentFlow(final AuthenticationRequest
             return authResult;
         }
 
-        //3. We couldn't get locally...If eligible return via broker... otherwise return local result
+        //3. try SAML Assertion
+        if (authenticationRequest.getSamlAssertion() != null) {
+            final AuthenticationResult authResultFromSaml = tryAcquireTokenSilentWithAssertion(authenticationRequest);
+            if (isAccessTokenReturned(authResultFromSaml)) {
```
iambmelt, line 484: Can we add some logging here to capture both when the access token is returned and when it is not? `Logger.v` seems appropriate for the success case; `Logger.w` for the unsuccessful case

adal/src/main/java/com/microsoft/aad/adal/AuthenticationContext.java
```diff
@@ -645,20 +645,70 @@ private AuthenticationResult acquireTokenSilentSync(final String resource,
...
+     *                 {@link AuthenticationResult #getUserInfo()}
+     * @param forceRefresh when true, access token is renewed using broker if available; otherwise, uses local refresh token
+     * @return A {@link Future} object representing the
+     * {@link AuthenticationResult} of the call. It contains Access
+     * Token,the Access Token's expiration time, Refresh token, and
+     * {@link UserInfo}.
+     * @throws AuthenticationException If silent request fails to get the token back.
+     * @throws InterruptedException    If the main thread is interrupted before or during the activity.
+     */
+    public AuthenticationResult acquireTokenSilentSyncWithAssertion(final String assertion,
```
iambmelt, line 675: Can we add nullability annotations here too? `@NonNull` if the param _cannot_ be `null`, `@Nullable` if it can.

adal/src/main/java/com/microsoft/aad/adal/AuthenticationContext.java
```diff
@@ -645,20 +645,70 @@ private AuthenticationResult acquireTokenSilentSync(final String resource,
...
+                                                                    final String userId,
+                                                                    final boolean forceRefresh,
+                                                                    final String claims)
+            throws AuthenticationException, InterruptedException {
+        final String methodName = ":acquireTokenSilentSyncWithAssertion";
+        return acquireTokenSilentSync(assertion, assertionType, resource, clientId, userId,
+                forceRefresh, claims, EventStrings.ACQUIRE_TOKEN_WITH_SAML_ASSERTION);
+    }
+
+    private AuthenticationResult acquireTokenSilentSync(final String assertion,
```
iambmelt, line 688: Can we add nullability annotations here too? `@NonNull` if the param _cannot_ be `null`, `@Nullable` if it can.

adal/src/telemetry/java/com/microsoft/aad/adal/EventStrings.java
```diff
@@ -186,6 +186,8 @@
 
     static final String ACQUIRE_TOKEN_WITH_REFRESH_TOKEN_2 = "5";
 
+    static final String ACQUIRE_TOKEN_WITH_SAML_ASSERTION = "20";
```
iambmelt, line 189: Does this match the API ID used by iOS? (Asking because I'm unsure)

adal/src/main/java/com/microsoft/aad/adal/AcquireTokenSilentHandler.java
paivybhav: Sure, will change that

adal/src/main/java/com/microsoft/aad/adal/AcquireTokenSilentHandler.java
paivybhav: Yeah, there was some other code there(which was distinguishing the two catch blocks) initially, which was not needed later. Now these 2 catch blocks are doing the same thing, didnt realize that when I made changes. 

They can be collapsed into one catch block, I have changed it.

paivybhav: > The `catch` block for the `IOExcepton`, `AuthenticationException` appears to contain the same logic as `ServerRespondingWithRetryableException`; 

Yeah, there was some other code there(which was distinguishing the two catch blocks) initially, which was not needed later. Now these 2 catch blocks are doing the same thing, didnt realize that when I made changes.

> is there any reason we should not collapse these into the same `catch` block?

They can be collapsed into one catch block, I have changed it.

paivybhav: > is there any reason we should not collapse these into the same `catch` block?

The types in multi-catch are needed to be disjoint ServerRetryableException is a subclass of AuthenticationException.

adal/src/main/java/com/microsoft/aad/adal/AcquireTokenSilentHandler.java
paivybhav: Yes, that's right. Will change it

adal/src/main/java/com/microsoft/aad/adal/Oauth2.java
paivybhav: > Can we add nullability (`@Nullable` and/or `@NonNull`) and `final` on the parameters to this method?

Yes both the parameters are expected to not be null. Will add that

adal/src/main/java/com/microsoft/aad/adal/AcquireTokenRequest.java
paivybhav: > If the caller has supplied a SAML assertion in their request, and no access token was returned are we falling-thru in hopes that the broker has an RT/FRT/PRT for this user already?

I understand what you are saying. But there is a possibilty that there are no refresh tokens available in the cache, and the broker could provide them. In the former code flow, when the refresh token isnt found in the cache the application would try and use the broker to fetch tokens. So I am ensuring that this fall through behavior is present because those use-cases would not have any assertions. 

Now about trying saml assertion before the broker code flow, if the saml assertion is present than its very likely well be able to get the tokens using that, if its not present then the former fall through behavior would come into the picture. 
I maybe missing out something here, correct me if I am wrong

adal/src/main/java/com/microsoft/aad/adal/AcquireTokenRequest.java
paivybhav: Sure, will add it.

adal/src/telemetry/java/com/microsoft/aad/adal/EventStrings.java
paivybhav: I checked the iOS corresponding [file](https://github.com/AzureAD/microsoft-authentication-library-common-for-objc/blob/4dfd66fc0cf2b8aa8be3872d2ad36c5c90e1ee19/IdentityCore/src/telemetry/MSIDTelemetryEventStrings.m), they dont seem to be using API ID, also they dont have anything for saml assertion here. 

I'm not sure about this, what do you suggest we do?

adal/src/main/java/com/microsoft/aad/adal/AcquireTokenSilentHandler.java
paivybhav: > This seems more like a `warning` than an `info` statement, right?

The corresponding line in the refreshtoken method at line 215 is also logging as an info, do you want me to change that too?

adal/src/telemetry/java/com/microsoft/aad/adal/EventStrings.java
paivybhav: > I'm not sure about this, what do you suggest we do?

I looked it up, the API ID is 6. You can see it [here](https://github.com/AzureAD/azure-activedirectory-library-for-objc/blob/dev/ADAL/src/ADAuthenticationContext.m) on line 315.

adal/src/main/java/com/microsoft/aad/adal/AcquireTokenRequest.java
iambmelt: Just wanted to clarify this was the desired behavior; sounds like 'yes'. Resolving this question, thanks

adal/src/main/java/com/microsoft/aad/adal/AcquireTokenSilentHandler.java
iambmelt: Yes please 👍

adal/src/main/java/com/microsoft/aad/adal/AcquireTokenSilentHandler.java
iambmelt: Thanks for clarifying.... So if `ServerRetryableException` is a subclass of `AuthenticationException` and these blocks contain the same logic can we remove the `ServerRespondingWithRetryableException` block entirely and still have the same behavior? Looks like yes, but perhaps I misunderstood

adal/src/telemetry/java/com/microsoft/aad/adal/EventStrings.java
iambmelt: Brilliant; let's use the same value for Android and I can update the associated telemetry documentation on our end

iambmelt: Doc updated; located [here ](https://microsoft.sharepoint.com/:w:/r/teams/aad/devex/_layouts/15/Doc.aspx?sourcedoc=%7BAAE38591-8A29-4A50-AE00-B9F66505B6D2%7D&file=Telemetry%20DDS.docx&action=default&mobileredirect=true)(corpnet access reqd)

paivybhav: Have changed the API ID to 6, will be able to see the changes after the next commit. Will mark this conversation as resolved.

adal/src/main/java/com/microsoft/aad/adal/AcquireTokenSilentHandler.java
paivybhav: Yeah you're right, I'll add the related changes and marking this as resolved.

PR #1538, opened 2020-06-24
Enable SSO for acquire token by refresh token API (for Scotty Flow)

shahzaibj:
In this PR, we've enabled the Scotty flow. Essentially we've enabled SSO via calling acquireTokenByRefreshToken APIs.

Previously ADAL did not store tokens in cache for the acquireTokenByRefreshToken API, however, we have to start saving tokens in cache for this API so that next time when someone calls acquireTokenSilent for this API, then they can get silent SSO.

To achieve this, we've done the following in this API:

- Add optional `scope` parameter to internal Acquire Token Request API
- Pass `openid` scope for AcquireTokenByRefreshToken API
- When the response comes back, save the tokens in cache

### DIFF
--- a/adal/src/main/java/com/microsoft/aad/adal/AcquireTokenRequest.java
+++ b/adal/src/main/java/com/microsoft/aad/adal/AcquireTokenRequest.java
@@ -31,6 +31,7 @@ import android.content.pm.PackageManager;
 import android.os.Bundle;
 import android.os.Handler;
 import android.os.Looper;
+import android.text.TextUtils;
 import android.util.Log;
 
 import androidx.annotation.Nullable;
@@ -174,7 +175,7 @@ class AcquireTokenRequest {
                     final String correlationId = String.format(" CorrelationId: %s", authenticationRequest.getCorrelationId().toString());
 
                     if (result == null) {
-                        Logger.e(TAG + methodName, "Returned result with exchanging auth code for token is null" + correlationId, "",
+                        Logger.e(TAG + methodName, "Returned result with exchanging refresh token for access token is null" + correlationId, "",
                                 ADALError.AUTH_REFRESH_FAILED);
                         throw new AuthenticationException(
                                 ADALError.AUTH_REFRESH_FAILED, correlationId);
@@ -192,8 +193,9 @@ class AcquireTokenRequest {
                                 " Access Token not returned from server ");
                     }
 
-                    if (result.getIdToken() != null && !result.isExtendedLifeTimeToken()) {
-                        final String rawIdToken = result.getIdToken();
+                    final String rawIdToken = result.getIdToken();
+
+                    if (!TextUtils.isEmpty(rawIdToken)) {
                         final IdToken idTokenRecord = new IdToken(rawIdToken);
                         final UserInfo userInfo = new UserInfo(idTokenRecord);
 
@@ -202,6 +204,16 @@ class AcquireTokenRequest {
                         tokenCacheItem.setUserInfo(userInfo);
                         tokenCacheItem.setTenantId(idTokenRecord.getTenantId());
 
+                        // This method is used in ADAL to store tokens in cache during a silent aka
+                        // refresh token request. We are going to be using the same method for this
+                        // API as well for consistency.
+                        //
+                        // To get this to work, we created a TokenCacheItem object above and
+                        // populate it with an ID Token and UserInfo object as in this API ADAL does
+                        // not do a cache look up and does not have the ID Token in cache already.
+                        // We got an ID Token from eSTS and will be using that ID Token to create
+                        // a UserInfo object and seed it into the cache along with the other tokens.
+                        // This allows us to achieve SSO for this API.
                         mTokenCacheAccessor.updateCachedItemWithResult(authenticationRequest,
                                 result, tokenCacheItem);
                     }
--- a/adal/src/main/java/com/microsoft/aad/adal/AcquireTokenRequest.java
+++ b/adal/src/main/java/com/microsoft/aad/adal/AcquireTokenRequest.java
@@ -146,9 +146,15 @@ class AcquireTokenRequest {
     }
 
     /**
-     * Developer is using refresh token call to do refresh without cache usage.
-     * App context or activity is not needed. Async requests are created, so this
-     * needs to be called at UI thread.
+     * This API allows to obtain a new access token in exchange for a refresh token. The refresh
+     * token must be provided to this API. The tokens obtained via this API may or may not be saved
+     * in ADAL's cache depending on two things: The openid scope was requested in the token, and a
+     * resource was supplied to the {@link AuthenticationRequest}. If both of those conditions are
+     * true, then tokens are saved in the cache, else not.
+     *
+     * @param refreshToken          the refresh token to use when exchanging for an access token
+     * @param authenticationRequest the {@link AuthenticationRequest} to use when requesting token
+     * @param externalCallback      the callback to which the result should be posted
      */
     void refreshTokenWithoutCache(final String refreshToken, final AuthenticationRequest authenticationRequest,
                                   final AuthenticationCallback<AuthenticationResult> externalCallback) {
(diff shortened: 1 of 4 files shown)


adal/src/main/java/com/microsoft/aad/adal/AcquireTokenRequest.java
```diff
@@ -168,11 +168,47 @@ public void run() {
...
                     final AcquireTokenSilentHandler acquireTokenSilentHandler = new AcquireTokenSilentHandler(mContext,
                             authenticationRequest, mTokenCacheAccessor);
-                    final AuthenticationResult authResult
+                    final AuthenticationResult result
                             = acquireTokenSilentHandler.acquireTokenWithRefreshToken(refreshToken);
+
+                    final String correlationId = String.format(" CorrelationId: %s", authenticationRequest.getCorrelationId().toString());
+
+                    if (result == null) {
+                        Logger.e(TAG + methodName, "Returned result with exchanging auth code for token is null" + correlationId, "",
```
shahzaibj, line 177: This should be exchanging refresh token

adal/src/main/java/com/microsoft/aad/adal/AcquireTokenRequest.java
```diff
@@ -168,11 +168,47 @@ public void run() {
...
+
+                        TokenCacheItem tokenCacheItem = new TokenCacheItem();
+                        tokenCacheItem.setRawIdToken(rawIdToken);
+                        tokenCacheItem.setUserInfo(userInfo);
+                        tokenCacheItem.setTenantId(idTokenRecord.getTenantId());
+
+                        mTokenCacheAccessor.updateCachedItemWithResult(authenticationRequest,
+                                result, tokenCacheItem);
+                    }
+
```
shahzaibj, line 208: Add comments here

shahzaibj: Addressed

adal/src/main/java/com/microsoft/aad/adal/AcquireTokenRequest.java
shahzaibj: Addressed

iambmelt:
For reviewers:
- This PR can be viewed with the whitespace (formatting) changes omitted: [view link](https://github.com/AzureAD/azure-activedirectory-library-for-android/pull/1538/files?w=1)

adal/src/main/java/com/microsoft/aad/adal/AuthenticationContext.java
```diff
@@ -1155,46 +1160,47 @@ public void startActivityForResult(Intent intent, int requestCode) {
...
-            "access_token": {
-                "xms_cc": {
-                    "values": ["CP1", "CP2"]
-                }
-            }
-        }
+     *                       <p>
+     *                       Sample input claim :
+     *                       {
+     *                       "userinfo":
```
iambmelt, line 1170: I see the formatter didn't like this inline doc: can we revert the formatter change and wrap the original formatting in a `<pre>` block?

iambmelt: This will tell the formatter to leave "as-is" so we can run it in the future and not have to battle with this whitespace change

adal/src/main/java/com/microsoft/aad/adal/AuthenticationRequest.java
```diff
@@ -187,6 +189,26 @@
...
+     * Create an Authentication Request to obtain a token. This overload also accepts a scope param.
+     *
+     * @param authority                 Authority URL
+     * @param resource                  Resource that is requested
+     * @param clientId                  ClientId for the app
+     * @param correlationId             Correlation Id used for logging & telemetry
+     * @param isExtendedLifetimeEnabled a boolean indicating if extended lifetime enabled
+     * @param scope                     the scope requested
+     */
+    AuthenticationRequest(String authority, String resource, String clientId,
```
iambmelt, line 202: There's lots of tech debt around nullability annotations in ADAL.... since we're adding new API here, can we add it to this method?

adal/src/main/java/com/microsoft/aad/adal/AcquireTokenRequest.java
```diff
@@ -168,11 +169,58 @@ public void run() {
...
+                    if (StringExtensions.isNullOrBlank(result.getAccessToken())) {
+                        Logger.e(TAG + methodName, "Access Token not returned from server", "", ADALError.AUTH_FAILED_NO_TOKEN);
+                        throw new AuthenticationException(ADALError.AUTH_FAILED_NO_TOKEN,
+                                " Access Token not returned from server ");
+                    }
+
+                    final String rawIdToken = result.getIdToken();
+
+                    if (!TextUtils.isEmpty(rawIdToken)) {
```
iambmelt, line 198: A comment here about whether or not this case where we didn't get an IdToken is expected, unexpected, valid, or invalid might be useful

adal/src/main/java/com/microsoft/aad/adal/AcquireTokenRequest.java
```diff
@@ -168,11 +169,58 @@ public void run() {
...
+                        // refresh token request. We are going to be using the same method for this
+                        // API as well for consistency.
+                        //
+                        // To get this to work, we created a TokenCacheItem object above and
+                        // populate it with an ID Token and UserInfo object as in this API ADAL does
+                        // not do a cache look up and does not have the ID Token in cache already.
+                        // We got an ID Token from eSTS and will be using that ID Token to create
+                        // a UserInfo object and seed it into the cache along with the other tokens.
+                        // This allows us to achieve SSO for this API.
+                        mTokenCacheAccessor.updateCachedItemWithResult(authenticationRequest,
```
iambmelt, line 217: Slightly confused by this; the javadoc for this method describes the cache as unused, but now we're updating the cache

```java
    /**
     * Developer is using refresh token call to do refresh without cache usage.
     * App context or activity is not needed. Async requests are created, so this
     * needs to be called at UI thread.
     */
```

iambmelt: The repercussions of this change aren't entirely clear to me. Previously this method was used in the `acquireTokenByRefreshToken` flow where IIRC _nothing_ was cached. Now that caching is being added, won't this change the flow of the code thru ADAL for consumers of this API upon subsequent use due to the newly cached RT now being saved in the cache? Is there any scenario we may not be considering whereby customers may be _relying_ on ADAL not caching this token?

adal/src/main/java/com/microsoft/aad/adal/AcquireTokenRequest.java
```diff
@@ -168,11 +169,58 @@ public void run() {
...
+                    final AuthenticationResult result
                             = acquireTokenSilentHandler.acquireTokenWithRefreshToken(refreshToken);
+
+                    final String correlationId = String.format(" CorrelationId: %s", authenticationRequest.getCorrelationId().toString());
+
+                    if (result == null) {
+                        Logger.e(TAG + methodName, "Returned result with exchanging refresh token for access token is null" + correlationId, "",
+                                ADALError.AUTH_REFRESH_FAILED);
+                        throw new AuthenticationException(
+                                ADALError.AUTH_REFRESH_FAILED, correlationId);
```
iambmelt, line 181: Do we want to include some additional message here that _nothing_ was received?

adal/src/main/java/com/microsoft/aad/adal/AcquireTokenRequest.java
iambmelt: Synced offline; this change should be fine. But docs should be updated

shahzaibj: Addressed. Updated docs.

adal/src/main/java/com/microsoft/aad/adal/AcquireTokenRequest.java
shahzaibj: Addressed

adal/src/main/java/com/microsoft/aad/adal/AcquireTokenRequest.java
shahzaibj: Addressed

adal/src/main/java/com/microsoft/aad/adal/AuthenticationRequest.java
shahzaibj: Addressed

adal/src/main/java/com/microsoft/aad/adal/AuthenticationContext.java
shahzaibj: Addressed

Issue #1073, opened 2017-12-12
Explicitly disable Content Provider access from WebViews


iambmelt:
Flagged by [`linkedin/qark`](https://github.com/linkedin/qark)

>While not a vulnerability by itself, it appears this app does not explicitly disable Content Provider access from WebViews. If the WebViews take in untrusted input, this can allow for data theft.<br /><br />To validate this vulnerability, load the following url in this WebView: http://www.secbro.com/qark/poc/WV_CPA_WARNING.html

See [WebSettings#setAllowContentAccess(boolean)](https://developer.android.com/reference/android/webkit/WebSettings.html#setAllowContentAccess(boolean))

Relevant files:
```
com/microsoft/aad/adal/AuthenticationActivity.java
com/microsoft/aad/adal/AuthenticationDialog.java 
```

iambmelt:
[ADO: Tracking (corpnet reqd)](https://identitydivision.visualstudio.com/Engineering/_boards/board/t/DevEx%20-%20Android/Backlog%20items/?workitem=1063911)

PR #1542, opened 2020-07-01
Set WebSettings#setAllowFileAccess(false)

### LINKED ISSUE #1074: Set WebSettings#setAllowUniversalAccessFromFileURLs(false)....
Reported by [`linkedin/qark`](https://github.com/linkedin/qark)

>JavaScript running in a file scheme context can access content from any origin. This is an insecure default value for minSdkVersion < 16 or may have been overridden (setAllowUniversalAccessFromFileURLs) in later versions.<br /><br />To validate this vulnerability, load the following url in this WebView: http://www.secbro.com/qark/poc/UNIV_FILE_WARNING.html.

AFAIK there isn't any js loaded from disk in either or CP/Authenticator. That said, as we support `minSdkVersion 14` we should err on the side of caution and disable this functionality by explicitly setting `WebSettings#setAllowUniversalAccessFromFileURLs(false)`

iambmelt:
~Closes #1074~

[Related security guidance](https://wiki.sei.cmu.edu/confluence/display/android/DRD02-J.+Do+not+allow+WebView+to+access+sensitive+local+resource+through+file+scheme)

Repurposing this PR - [see documentation from `setAllowFileAccess()`](https://developer.android.com/reference/android/webkit/WebSettings#setAllowFileAccess(boolean))
>To prevent possible security issues targeting Build.VERSION_CODES.Q and earlier, you should explicitly set this value to false.

### DIFF
--- a/adal/src/main/java/com/microsoft/aad/adal/AuthenticationActivity.java
+++ b/adal/src/main/java/com/microsoft/aad/adal/AuthenticationActivity.java
@@ -436,6 +436,8 @@ public class AuthenticationActivity extends DualScreenActivity {
     private void setupWebView() {
         final WebSettings webSettings = mWebView.getSettings();
         webSettings.setJavaScriptEnabled(true);
+        // Explicitly set false, may be true for older devices (ICS MR1 & earlier)
+        webSettings.setAllowUniversalAccessFromFileURLs(false);
         webSettings.setAllowContentAccess(false);
         mWebView.requestFocus(View.FOCUS_DOWN);
 
--- a/adal/src/main/java/com/microsoft/aad/adal/AuthenticationActivity.java
+++ b/adal/src/main/java/com/microsoft/aad/adal/AuthenticationActivity.java
@@ -434,6 +434,7 @@ public class AuthenticationActivity extends DualScreenActivity {
 
     private void setupWebView() {
         mWebView.getSettings().setJavaScriptEnabled(true);
+        mWebView.getSettings().setAllowContentAccess(false);
         mWebView.requestFocus(View.FOCUS_DOWN);
 
         // Set focus to the view for touch event
--- a/adal/src/main/java/com/microsoft/aad/adal/AuthenticationActivity.java
+++ b/adal/src/main/java/com/microsoft/aad/adal/AuthenticationActivity.java
@@ -436,8 +436,7 @@ public class AuthenticationActivity extends DualScreenActivity {
     private void setupWebView() {
         final WebSettings webSettings = mWebView.getSettings();
         webSettings.setJavaScriptEnabled(true);
-        // Explicitly set false, may be true for older devices (ICS MR1 & earlier)
-        webSettings.setAllowUniversalAccessFromFileURLs(false);
+        webSettings.setAllowFileAccess(false);
         webSettings.setAllowContentAccess(false);
         mWebView.requestFocus(View.FOCUS_DOWN);
 
--- a/adal/src/main/java/com/microsoft/aad/adal/AuthenticationActivity.java
+++ b/adal/src/main/java/com/microsoft/aad/adal/AuthenticationActivity.java
@@ -43,6 +43,7 @@ import android.view.inputmethod.InputMethodManager;
 import android.webkit.ClientCertRequest;
 import android.webkit.CookieManager;
 import android.webkit.CookieSyncManager;
+import android.webkit.WebSettings;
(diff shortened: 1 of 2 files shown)


iambmelt:
This PR has failing CI builds due to busted tests that are removed in #1073 - merging that PR and merging those changes into this branch will trigger a successful build

adal/src/main/java/com/microsoft/aad/adal/AuthenticationActivity.java
```diff
@@ -436,6 +436,8 @@ protected void onRestoreInstanceState(final Bundle savedInstanceState) {
     private void setupWebView() {
         final WebSettings webSettings = mWebView.getSettings();
         webSettings.setJavaScriptEnabled(true);
+        // Explicitly set false, may be true for older devices (ICS MR1 & earlier)
+        webSettings.setAllowUniversalAccessFromFileURLs(false);
```
AdamBJohnsonx, line 440: Are there other WebSettings that we should manipulate?

There's allowFileAccess, allowFileAccessFromFileUrls, etc. which have similar statements about their usability and functionality on different versions, but I don't see a statement about what the defaults are... 

https://developer.android.com/reference/android/webkit/WebSettings

Also, is it possible to test this?

iambmelt: Let me dig deeper into that and see what I can find

iambmelt: | API                                     | Default Behavior                                                                                                                        | Enabled in ADAL | Desired state in ADAL |
|-----------------------------------------|-----------------------------------------------------------------------------------------------------------------------------------------|-----------------|-----------------------|
| [`setAllowContentAccess()`](https://developer.android.com/reference/android/webkit/WebSettings#setAllowContentAccess(boolean))               | Enabled                                                                                                                                 | Disabled        | Disabled              |
| [`setAllowFileAccess()`](https://developer.android.com/reference/android/webkit/WebSettings#setAllowFileAccess(boolean))                  | API Level Dependent <br/> Enabled by default for API <=29 <br/> Disabled by default for API 30+                                         | Enabled         | Disabled              |
| [`setAllowFileAccessFromFileURLs()`](https://developer.android.com/reference/android/webkit/WebSettings#setAllowFileAccessFromFileURLs(boolean))      | Ignored if `getAllowUniversalAccessFromFileURLs()` is `true` <br/> Enabled by default for API <=15 <br/> Disabled by default for API 16+ | Disabled        | Disabled              |
| [`setAllowUniversalAccessFromFileURLs()`](https://developer.android.com/reference/android/webkit/WebSettings#setAllowUniversalAccessFromFileURLs(boolean)) | Enabled by default for API <=15 <br/> Disabled by default for API 16+                                                                   | Disabled        | Disabled              |

iambmelt: @AdamBJohnsonx See above table, since we're now targeting API 27 in our build, we can actually no-op all of these _except_ `setAllowFileAccess()` which is still enabled by default.

From the docs:
```
To prevent possible security issues targeting Build.VERSION_CODES.Q and earlier, you should explicitly set this value to false.
```

So I am going to repurpose this PR to instead disable file access, and will close the _original_ issue as it no longer applies.

iambmelt: @AdamBJohnsonx ADAL `minSdk` is currently 16, with `targetSdk` at 27. So these are noops for everything _except_ `setAllowFileAccess()`.

From the docs:
> To prevent possible security issues targeting Build.VERSION_CODES.Q and earlier, you should explicitly set this value to false.

So I am going to close the original issue, and repurpose this PR to address `setAllowFileAccess()`

iambmelt: (I am not thinking I will implement a specific test for this, given it's a built-in)

iambmelt: See 02f83275bc60159ceb97fb72ada5a60b66bac529

iambmelt:
> Could you add a backlog item to add a test to insure that the actual bug doesn't manifest itself in the software when exercised?

Added:
https://identitydivision.visualstudio.com/Engineering/_workitems/edit/1069892

PR #1552, opened 2020-08-12
Clear the MSAL cache replica is the ADAL cache is cleared

iambmelt:
Clear the MSAL replica cache if the ADAL cache is cleared.

Changes in this PR:
- Updates the submodule
- Wrap the provided `ITokenCacheStore` interface with our own implementation (if provided instance is `!= null`)
- Delegate `ITokenCacheStore` method calls to provided instance
- If `removeAll()` is called, get an instance of the MSAL replica cache and clear it
- Adds nullability annotations to `AuthenticationContext` constructor params
- Fix formatting issues in `AuthenticationContext`
- Adds a new unit test to verify cache replica is cleared
- Add import statement to test, remove FQN usage to shorten lines
- Removes a doze mode test that's no longer applicable

### DIFF
--- a/adal/src/androidTest/java/com/microsoft/aad/adal/DiscoveryTests.java
+++ b/adal/src/androidTest/java/com/microsoft/aad/adal/DiscoveryTests.java
@@ -384,23 +384,6 @@ public class DiscoveryTests extends AndroidTestHelper {
         }
     }
 
-    @Test
-    @SdkSuppress(minSdkVersion = Build.VERSION_CODES.M)
-    public void testValidateAuthorityFailedInDozeMode() throws IOException {
-        final FileMockContext context = new FileMockContext(androidx.test.platform.app.InstrumentationRegistry.getInstrumentation().getContext());
-        final Discovery discovery = new Discovery(context);
-        context.setDeviceInIdleMode();
-        final URL endpointFull = new URL("https://login.invalidlogin.net/common/oauth2/authorize");
-
-        try {
-            discovery.validateAuthority(endpointFull);
-            fail();
-        } catch (AuthenticationException e) {
-            assertNotNull(e);
-            assertTrue(e.getCode().equals(ADALError.NO_NETWORK_CONNECTION_POWER_OPTIMIZATION));
-        }
-    }
-
     static String getDiscoveryResponse() {
         final Map<String, String> discoveryResponse = AuthorityValidationMetadataCacheTest.getDiscoveryResponse();
         final JSONObject discoveryResponseJsonObject = new JSONObject(discoveryResponse);
--- a/adal/src/main/java/com/microsoft/aad/adal/AuthenticationContext.java
+++ b/adal/src/main/java/com/microsoft/aad/adal/AuthenticationContext.java
@@ -1250,40 +1250,40 @@ public class AuthenticationContext {
      * @throws JSONException if input claims is an invalid JSON
      *
      * <pre>
-     Sample input claim :
-        {
-            "userinfo": {
-                "given_name": {"essential": true},
-                "email": {"essential": true},
-            },
-            "id_token": {
-                "auth_time": {"essential": true},
-            }
-        }
-
-    Sample capabilities list : [CP1, CP2 CP3]
-
-    Output merged claims :
-        {
-            "userinfo": {
-                "given_name": {
-                    "essential": true
-                },
-                "email": {
-                    "essential": true
-                }
-            },
-            "id_token": {
-                "auth_time": {
-                    "essential": true
-                }
-            },
-            "access_token": {
-                "xms_cc": {
-                    "values": ["CP1", "CP2"]
-                }
-            }
-        }
+     * Sample input claim :
+     *   {
+     *       "userinfo": {
+     *           "given_name": {"essential": true},
+     *           "email": {"essential": true},
+     *       },
+     *       "id_token": {
+     *           "auth_time": {"essential": true},
+     *       }
+     *   }
+     *
+     * Sample capabilities list : [CP1, CP2 CP3]
+     *
+     * Output merged claims :
+     *   {
+     *       "userinfo": {
+     *           "given_name": {
+     *               "essential": true
+     *           },
+     *           "email": {
+     *               "essential": true
+     *           }
+     *       },
+     *       "id_token": {
+     *           "auth_time": {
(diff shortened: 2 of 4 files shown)


adal/src/androidTest/java/com/microsoft/aad/adal/DiscoveryTests.java
```diff
@@ -384,23 +384,6 @@ public void testValidateAuthorityFailedWithoutNetwork() throws IOException {
         }
     }
 
-    @Test
```
iambmelt, line 387: This test is now invalid, due to doze mode detection changes that now use the internet connection if active/available irrespective of if the current app thinks its in doze mode

adal/src/main/java/com/microsoft/aad/adal/AuthenticationContext.java
```diff
@@ -141,28 +145,77 @@ public AuthenticationContext(Context appContext, String authority, boolean valid
...
         mAuthority = extractAuthority(authority);
         mValidateAuthority = validateAuthority;
-        mTokenCacheStore = tokenCacheStore;
+
+        if (null != tokenCacheStore) {
+            mTokenCacheStore = wrapCache(tokenCacheStore);
+        }
+    }
+
+    private ITokenCacheStore wrapCache(@NonNull final ITokenCacheStore originalCache) {
```
shahzaibj, line 183: Do we need to do this? only `removeAll` seems to be doing something special. may be I'm missing something here

adal/src/main/java/com/microsoft/aad/adal/AuthenticationContext.java
```diff
@@ -141,28 +145,77 @@ public AuthenticationContext(Context appContext, String authority, boolean valid
...
         initialize(appContext, authority, tokenCacheStore, true, false);
     }
 
-    private void initialize(Context appContext, String authority, ITokenCacheStore tokenCacheStore,
-                            boolean validateAuthority, boolean defaultCache) {
+    private void initialize(@NonNull final Context appContext,
+                            @NonNull final String authority,
+                            @Nullable final ITokenCacheStore tokenCacheStore,
+                            boolean validateAuthority,
+                            boolean defaultCache) {
```
AdamBJohnsonx, line 158: final?

adal/src/main/java/com/microsoft/aad/adal/AuthenticationContext.java
iambmelt: ADAL supports a custom cache implementation. This is to ensure that this function is still called even if the developer is using their own cache implementation (see constructor where `ITokenCacheStore` can be supplied)

adal/src/main/java/com/microsoft/aad/adal/AuthenticationContext.java
iambmelt: Will add

adal/src/main/java/com/microsoft/aad/adal/AuthenticationContext.java
```diff
@@ -141,28 +145,77 @@ public AuthenticationContext(Context appContext, String authority, boolean valid
...
+            }
+
+            @Override
+            public void removeAll() {
+                // Clear our original cache
+                originalCache.removeAll();
+
+                // clear our replica cache
+                getMsalOAuth2TokenCache(mContext).clearAll();
+            }
```
AdamBJohnsonx, line 217: do we want to synchronize here, and, indeed, for all of these?

adal/src/main/java/com/microsoft/aad/adal/AuthenticationContext.java
iambmelt: f0c622ae20f42f6888f34032fabb1583c03e544f

adal/src/main/java/com/microsoft/aad/adal/AuthenticationContext.java
```diff
@@ -141,28 +145,77 @@ public AuthenticationContext(Context appContext, String authority, boolean valid
...
+
+        if (null != tokenCacheStore) {
+            mTokenCacheStore = wrapCache(tokenCacheStore);
+        }
+    }
+
+    private ITokenCacheStore wrapCache(@NonNull final ITokenCacheStore originalCache) {
+        return new ITokenCacheStore() {
+            @Override
+            public TokenCacheItem getItem(final String key) {
```
iambmelt, line 186: @AdamBJohnsonx This is the suggestion? For each of these overridden methods? (Just confirming I understood your feedback)

```suggestion
            public synchronized TokenCacheItem getItem(final String key) {
```

adal/src/main/java/com/microsoft/aad/adal/AuthenticationContext.java
iambmelt: Like so?
https://github.com/AzureAD/azure-activedirectory-library-for-android/pull/1552/files#r469607985

adal/src/main/java/com/microsoft/aad/adal/AuthenticationContext.java
AdamBJohnsonx: Yeah.  The main goal would be that we want removeAll and setItem to have a defined behavior.

iambmelt: ACK. Will push

adal/src/main/java/com/microsoft/aad/adal/AuthenticationContext.java
AdamBJohnsonx: That's the thought.  Not having setItem have undefined behavior with respect to removeAll.

adal/src/main/java/com/microsoft/aad/adal/AuthenticationContext.java
iambmelt: Applied in 7a26dbee681f6ab9fd5fe8a7d10ff13d586414fe

adal/src/main/java/com/microsoft/aad/adal/AuthenticationContext.java
iambmelt: Applied in 7a26dbee681f6ab9fd5fe8a7d10ff13d586414fe

PR #1556, opened 2020-08-28
Release/3.1.1 

shahzaibj:
Prep for ADAL release 3.1.1 (Depends on common 2.1.1 - awaiting deployment)


userappwithbroker/build.gradle
```diff
@@ -70,7 +70,7 @@ dependencies {
     implementation "com.google.android.material:material:$rootProject.ext.materialVersion"
 
     localImplementation(project(':adal'))
-    distImplementation("com.microsoft.aad:adal:3.0.2") {
+    distImplementation("com.microsoft.aad:adal:3.1.0") {
```
AdamBJohnsonx, line 73: Did we want 3.1.0 here or 3.1.1?

iambmelt: Should be 3.1.1 -- thanks, will update

iambmelt: Actually, let's hold off on this until we get these bits deployed -- this artifact isn't going to resolve until after we deploy.

PR #1560, opened 2020-09-15
Updated gradle to latest plugin and gradle versions

shoatman:

### DIFF
--- a/adal/build.gradle
+++ b/adal/build.gradle
@@ -1,5 +1,5 @@
 apply plugin: 'com.android.library'
-apply plugin: 'findbugs'
+//apply plugin: 'findbugs'
 apply plugin: 'pmd'
 apply plugin: 'checkstyle'
 apply plugin: 'maven-publish'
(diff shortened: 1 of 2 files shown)


adal/build.gradle
```diff
@@ -298,6 +297,7 @@ artifacts {
     archives sourcesJar
 }
 
+/*
```
iambmelt, line 300: Do we plan to readd? Or should we remove outright?

iambmelt: From offline mail, sounds like plan is remove this entirely and swap it out for SpotBugs -- in that case, I'm fine with deleting these blocks entirely, since there's no plan to reenable later

shoatman: I removed all references to findbugs.  I added the workitem to track introducing spotbugs to our projects.  https://identitydivision.visualstudio.com/Engineering/_boards/board/t/Auth%20Client%20-%20Android/Backlog%20items/?workitem=1151494

iambmelt (approved):

PR #1568, opened 2020-11-02
Adjoh/adal update common

AdamBJohnsonx:

* @mentions of the person required for reviewing proposed changes (optional)


userappwithbroker/userapp-rules.pro
```diff
@@ -0,0 +1,66 @@
+# Add project specific ProGuard rules here.
```
rpdome, line 1: Do we need this? (given that user app is our internal test app... not having this means one less thing to maintain.)

adal/adal-rules.pro
```diff
@@ -0,0 +1,65 @@
...
+# class:
+#-keepclassmembers class fqcn.of.javascript.interface.for.webview {
+#   public *;
+#}
+
+-dontskipnonpubliclibraryclasses
+
+##---------------Begin: proguard configuration for Common  --------
+-keep,includedescriptorclasses class com.microsoft.identity.common.exception.** { *; }
+-keep,includedescriptorclasses class com.microsoft.identity.common.internal.broker.** { *; }
```
rpdome, line 23: I'm thinking if we could just do 

-keep,includedescriptorclasses class com.microsoft.identity.common.**

This way, we don't have to maintain this file (and the one in MSAL) much when we introduce new stuffs in common.

adal/adal-rules.pro
```diff
@@ -0,0 +1,65 @@
...
+-keep,includedescriptorclasses class com.microsoft.identity.common.internal.cache.** { *; }
+-keep,includedescriptorclasses class com.microsoft.identity.logging.internal.dto.** { *; }
```
rpdome, line 25: missing .common

adal/adal-rules.pro
```diff
@@ -0,0 +1,65 @@
...
+-keep,includedescriptorclasses class com.microsoft.identity.internal.logging.** { *; }
```
rpdome, line 26: missing .common

rpdome (approved):
left a few comments.

userappwithbroker/userapp-rules.pro
AdamBJohnsonx: I think we should dog food the ProGuard rules.

adal/adal-rules.pro
AdamBJohnsonx: The idea here is that we don't use most of common in ADAL, so we should drop as much of it as we can.

PR #1573, opened 2020-12-01
Update common submodule

AdamBJohnsonx:

* @mentions of the person required for reviewing proposed changes (optional)

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -1,11 +1,12 @@
 buildscript {
     apply from: rootProject.file("gradle/versions.gradle")
+    println rootProject
 
     repositories {
         google()
         jcenter()
     }
-
+    println "com.android.tools.build:gradle:$rootProject.ext.gradleVersion"
     dependencies {
         classpath "com.android.tools.build:gradle:$rootProject.ext.gradleVersion"
         classpath "net.serenity-bdd:serenity-gradle-plugin:1.9.6"
--- a/gradle/versions.gradle
+++ b/gradle/versions.gradle
@@ -8,7 +8,7 @@ ext {
     buildToolsVersion = "28.0.3"
 
     // Plugins
-    gradleVersion = '3.3.3'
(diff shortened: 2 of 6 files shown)


build.gradle
```diff
@@ -1,11 +1,12 @@
 buildscript {
     apply from: rootProject.file("gradle/versions.gradle")
+    println rootProject
```
AdamBJohnsonx, line 3:
```suggestion
```

build.gradle
```diff
@@ -1,11 +1,12 @@
...
 
     repositories {
         google()
         jcenter()
     }
-
+    println "com.android.tools.build:gradle:$rootProject.ext.gradleVersion"
```
AdamBJohnsonx, line 9:
```suggestion

```

build.gradle
```diff
@@ -5,7 +5,6 @@ buildscript {
         google()
         jcenter()
     }
-
     dependencies {
```
AdamBJohnsonx, line 8:
```suggestion

    dependencies {
```

gradle/versions.gradle
```diff
@@ -8,7 +8,7 @@ ext {
     buildToolsVersion = "28.0.3"
```
gitaumoses4, line 8: I see the build tools were updated to `29.0.3` on `travis.yml`. How comes we are not doing the same here?

AdamBJohnsonx: The build tools version is ignored, in any event.

gitaumoses4: Ah. I see. Thank you for the clarification.

AdamBJohnsonx: Eh - we should really clean up the Gradle files.  If this works, I may clean out some of these warnings before it merges.

AdamBJohnsonx:
Think this Is now moot.

PR #1579, opened 2021-01-20
Resolving Authenticator MFA flow for Webview

tanmaymanolkar1:
This PR contains changes needed in ADAL.
Related  PR : https://github.com/AzureAD/microsoft-authentication-library-common-for-android/pull/1195

### DIFF
--- a/adal/src/androidTest/java/com/microsoft/aad/adal/BasicWebViewClientTests.java
+++ b/adal/src/androidTest/java/com/microsoft/aad/adal/BasicWebViewClientTests.java
@@ -51,9 +51,13 @@ import java.security.cert.X509Certificate;
 import java.util.concurrent.CountDownLatch;
 import java.util.concurrent.TimeUnit;
 
+import static com.microsoft.identity.common.adal.internal.AuthenticationConstants.Broker.AUTHENTICATOR_MFA_LINKING_PREFIX;
+import static com.microsoft.identity.common.adal.internal.AuthenticationConstants.Broker.COMPANY_PORTAL_PROD_APP_PACKAGE_NAME;
+import static com.microsoft.identity.common.adal.internal.AuthenticationConstants.Broker.PLAY_STORE_INSTALL_PREFIX;
 import static com.microsoft.identity.common.adal.internal.AuthenticationConstants.Browser.RESPONSE_ERROR_CODE;
 import static com.microsoft.identity.common.adal.internal.AuthenticationConstants.Browser.RESPONSE_ERROR_MESSAGE;
 import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertFalse;
 import static org.junit.Assert.assertTrue;
 import static org.junit.Assert.fail;
 
@@ -62,6 +66,7 @@ public class BasicWebViewClientTests {
 
     private static final String NONCE = "123123-123213-123";
     private static final String CONTEXT = "ABcdeded";
+    private static final String TEST_REDIRECT_URI = "abc12";
     private static final String TEST_PKEY_AUTH_URL = AuthenticationConstants.Broker.PKEYAUTH_REDIRECT
             + "?Nonce="
             + NONCE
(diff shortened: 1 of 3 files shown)


adal/src/main/java/com/microsoft/aad/adal/BasicWebViewClient.java
```diff
@@ -458,6 +460,39 @@ public void run() {
 
             openLinkInBrowser(parameters.get(INSTALL_URL_KEY));
             view.stopLoading();
+            return true;
+        } else if (url.toLowerCase(Locale.US).startsWith(PLAY_STORE_INSTALL_PREFIX)) {
+
+            view.stopLoading();
+            if (!(url.startsWith(PLAY_STORE_INSTALL_PREFIX + COMPANY_PORTAL_PROD_APP_PACKAGE_NAME))
+                    && !(url.startsWith(PLAY_STORE_INSTALL_PREFIX + AuthenticationConstants.Broker.AZURE_AUTHENTICATOR_APP_PACKAGE_NAME))) {
+                return false;
```
AdamBJohnsonx, line 469: Can we log something here as to why we're doing this?

adal/src/main/java/com/microsoft/aad/adal/BasicWebViewClient.java
```diff
@@ -458,6 +460,39 @@ public void run() {
...
+            }
+            final String appPackageName = (url.contains(COMPANY_PORTAL_PROD_APP_PACKAGE_NAME) ?
+                    COMPANY_PORTAL_PROD_APP_PACKAGE_NAME : AuthenticationConstants.Broker.AZURE_AUTHENTICATOR_APP_PACKAGE_NAME);
+            Logger.verbose(TAG + methodName, "Request to open PlayStore.");
```
AdamBJohnsonx, line 473: ...to install which application?

AdamBJohnsonx:
Again, if we're doing this, I'd like to see some unit tests.

adal/src/main/java/com/microsoft/aad/adal/BasicWebViewClient.java
```diff
@@ -474,6 +476,40 @@ public void run() {
 
             openLinkInBrowser(parameters.get(INSTALL_URL_KEY));
             view.stopLoading();
+            return true;
+        } else if (url.toLowerCase(Locale.US).startsWith(PLAY_STORE_INSTALL_PREFIX)) {
```
iamgusain, line 480: >else if [](start = 10, length = 7)

It looks like we have enough if-else cases here.
it might be a good idea to refactor this code to create separate UrlHandler instance based on the Url type with each handling the url in its own handleUrl method.

UrlType urlType = getUrlType(url);
switch (urlType)
{
case PLAYSTORE_INSTALL_URL:
return PlayStoreUrlHandler.HandleUrl(url);
case AUTHENTICATOR_MFA_LINKING:
return AuthenticatorMFALinkingUrlHandler.HandlerUrl(url);
...
}

This way the code can be shared between MSAL and ADAL, rather than duplicating, right?

tanmaymanolkar1: The reason why this extensive refactoring isnt done is because, unlike MSAL, ADAL does not depend on common module when it comes to handling Overriden Url loading, also ADAL is on the path to be deprecated, and would be having far less releases, so at this point it wouldnt be worth refactoring this entire logic for ADAL, and there might be few usecases involved here which would not have been exactly mirrored as MSAL. Hence for quick resolution, we have added the capability of supporting 1st party brokers in the existing code of ADAL.

adal/src/androidTest/java/com/microsoft/aad/adal/BasicWebViewClientTests.java
```diff
@@ -75,6 +80,14 @@
             AuthenticationConstants.Broker.BROWSER_EXT_PREFIX + "https://graph.microsoft.io";
     private static final String TEST_INSTALL_REQUEST_URL =
             AuthenticationConstants.Broker.BROWSER_EXT_INSTALL_PREFIX + "url";
+    private static final String TEST_PLAY_STORE_INSTALL_AUTH_APP_URL =
+            PLAY_STORE_INSTALL_PREFIX + AuthenticationConstants.Broker.AZURE_AUTHENTICATOR_APP_PACKAGE_NAME;
+    private static final String TEST_PLAY_STORE_INSTALL_CP_URL =
+            PLAY_STORE_INSTALL_PREFIX + COMPANY_PORTAL_PROD_APP_PACKAGE_NAME;
+    private static final String TEST_PLAY_STORE_INSTALL_INVALID_APP =
+            PLAY_STORE_INSTALL_PREFIX + "com.azure.xyz";
```
AdamBJohnsonx, line 88: Can we add a test to make sure we're blocking a custom scheme?

adal/src/main/java/com/microsoft/aad/adal/BasicWebViewClient.java
```diff
@@ -474,6 +476,40 @@ public void run() {
...
+                intent.addFlags(Intent.FLAG_ACTIVITY_NEW_TASK | Intent.FLAG_ACTIVITY_CLEAR_TASK);
+                mCallingContext.startActivity(intent);
+            } catch (android.content.ActivityNotFoundException e) {
+                //if GooglePlay is not present on the device.
+                Logger.error(TAG + methodName, "Failed to launch the PlayStore.", e);
+            }
+
+            return true;
+        } else if (url.toLowerCase(Locale.US).startsWith(AuthenticationConstants.Broker.AUTHENTICATOR_MFA_LINKING_PREFIX)) {
+
```
AdamBJohnsonx, line 503: extra line

AdamBJohnsonx (approved):

iamgusain (dismissed):
:shipit:

iamgusain (approved):
:shipit:

PR #1591, opened 2021-03-25
Update common dependency to 3.2.0

p3dr0rv:
Update common dependency to 3.2.0 in order to run gradle assembleDist

### DIFF
--- a/changelog.txt
+++ b/changelog.txt
@@ -1,10 +1,9 @@
 v.Next
 ----------
-- [PATCH] Update common dependency to 3.2.0 (#1591)
+- Picks up common@3.2.0
 - [PATCH] Adds API23 WebViewClient#onReceivedError overload (#1580)
 - [PATCH] Fixes for MFA setup using Authenticator app.(#1579)
 - [MINOR] Changes to Broker Validation to allow setting whether to trust debug brokers (prod brokers are always trusted).
-- [PATCH] Update common dependency to 3.1.0 (#1583)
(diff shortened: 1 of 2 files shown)


changelog.txt
```diff
@@ -1,5 +1,6 @@
 v.Next
 ----------
+- [PATCH] Update common dependency to 3.2.0 (#1591)
```
shahzaibj, line 3: Seems there is already a line for down below on line 7....could just update there....but with that said, I prefer not documenting common change in this way because we don't really know whether it is MAJOR, MINOR or PATCH...or may be we do if we look at common's changelog, however, I think we usually document common dependency change as follows

- Picks up common@3.2.0

so it is neither of major, minor or patch

p3dr0rv: thanks @shahzaibj  for the feedback :D, so in conclusion the preferred way to approach this is to remove line 7, and leave only
* Picks up common@3.2.0
right?

shahzaibj: Yes my recommendation would be to remove both line 3 and 7, and add a line for `Picks up common@3.2.0` - though this value will probably change whenever we actually plan to release another version of ADAL...so we can probably just leave this line out but doesn't harm to add it there either.

cc/ @iambmelt

changelog.txt
```diff
@@ -1,9 +1,9 @@
 v.Next
 ----------
+- Picks up common@3.2.0
```
shahzaibj, line 3: Probably put it at the bottom

p3dr0rv: sure

shahzaibj (approved):

PR #1587, opened 2021-03-04
Replaced deprecated PackageInfo.signatures with PackageInfo.signingInfo

paulkagiri:
Replaced the deprecated `PackageInfo.signatures` with `PackageInfo.signingInfo`. Created a utility method across code and tests for compatibility reasons.

Work Item: https://identitydivision.visualstudio.com/Engineering/_workitems/edit/1226178
Issue: https://github.com/AzureAD/microsoft-authentication-library-common-for-android/issues/851

PR #1601, opened 2021-05-04
Fix ADAL local variant build after consuming common4j

shahzaibj:
We need to consume common as transitive true now because it needs to resolve classes from common4j on the classpath

If we don't do this then ADAL won't build because classes from common4j are not on the Runtime classpath

Similar to: AzureAD/microsoft-authentication-library-for-android#1380

Related common PR: AzureAD/microsoft-authentication-library-common-for-android#1338

Tests are failing per changes made in https://github.com/AzureAD/azure-activedirectory-library-for-android/pull/1587

### DIFF
--- a/adal/src/main/java/com/microsoft/aad/adal/BasicWebViewClient.java
+++ b/adal/src/main/java/com/microsoft/aad/adal/BasicWebViewClient.java
@@ -169,7 +169,15 @@ abstract class BasicWebViewClient extends WebViewClient {
     public void onReceivedError(@NonNull final WebView view,
                                 @NonNull final WebResourceRequest request,
                                 @NonNull WebResourceError error) {
-        sendErrorResponse(error.getErrorCode(), error.getDescription().toString());
+        final String methodName = "onReceivedError (23)";
+        final boolean isForMainFrame = request.isForMainFrame();
+
+        com.microsoft.identity.common.logging.Logger.warn(TAG + methodName, "WebResourceError - isForMainFrame? " + isForMainFrame);
+        com.microsoft.identity.common.logging.Logger.warnPII(TAG + methodName, "Failing url: " + request.getUrl());
+
+        if (isForMainFrame) {
+            sendErrorResponse(error.getErrorCode(), error.getDescription().toString());
+        }
     }
 
     private void sendErrorResponse(final int errorCode,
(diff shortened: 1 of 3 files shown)


adal/build.gradle
```diff
@@ -141,7 +141,7 @@ dependencies {
 
     // 'local' flavor dependencies
     localApi(project(":common")) {
```
iamgusain, line 143: >localApi [](http://example.com/codeflow?start=4&length=8)

Why this is only needed for local variant?

shahzaibj: because dist pulls from Maven and that doesn't have common4j yet...we will have to do it for dist as well when we publish common4j...added todo as well: https://github.com/AzureAD/azure-activedirectory-library-for-android/pull/1601/commits/2a2db290221a8533307e47091b3ec7fea3217e6c

adal/build.gradle
```diff
@@ -141,12 +141,13 @@ dependencies {
...
-        transitive = false
+        transitive = true
     }
 
     snapshotApi(group: 'com.microsoft.identity', name: 'common', version: '3.2.0', changing: true)
 
     // 'dist' flavor dependencies
+    //TODO: we will have to change transitive to true once common4j is published
```
rpdome, line 150: Would this work when common4j is not yet published - but is consumed by common?

shahzaibj: I don't know yet - will need to publish a common with the changes made here https://github.com/AzureAD/microsoft-authentication-library-common-for-android/pull/1338 and try it out

rpdome: Let's try it out - if we need to change it here, it's a breaking change (meaning that we probably wouldn't want to do this now).

shahzaibj: I guess it makes sense to just make it true right now

adal/build.gradle
iamgusain: I am still not clear on, why this only matters for local variant. as long as the verison of :common is same, shouldnt it be varian agnostic. unless we are doing something in common project to be variant specific. are we?

adal/build.gradle
```diff
@@ -146,12 +146,13 @@ dependencies {
...
```
iambmelt, line 155: Should we log a warning from gradle here?

iambmelt (approved):

adal/build.gradle
shahzaibj: We could, but I'm going to leave as is for now..I don't think we will be in this state for too long (two weeks probably at most)

shahzaibj:
@iambmelt Your review got dismissed after I updated with dev...want to give this one another approval?

rpdome (approved):

PR #1605, opened 2021-05-06
update instrumented test to use templates in common

p3dr0rv:
Instrumented tests have been failing intermittently
the error presented is `Access to the path [...] is denied`  which is a File or folder in use error type

in this PR we intent to use the new [common template](https://github.com/AzureAD/microsoft-authentication-library-common-for-android/blob/pedroro/pipelines-check/azure-pipelines/templates/steps/auth-client/docker-tasks.yml) with:

* clean true in the git checkout task  (git clean -ffdx && git reset --hard HEAD` before fetching)
* add [PostBuildCleanup task](https://github.com/MicrosoftPremier/VstsExtensions/blob/master/PostBuildCleanup/en-US/overview.md)
* add `--volumes` to [`docker system prune`](https://docs.docker.com/engine/reference/commandline/system_prune/)

https://identitydivision.visualstudio.com/Engineering/_build/results?buildId=772242&view=results

[common](https://github.com/AzureAD/microsoft-authentication-library-common-for-android/pull/1342)
[msal](https://github.com/AzureAD/microsoft-authentication-library-for-android/pull/1387)


iamgusain (approved):
:shipit:

PR #1606, opened 2021-05-10
Adding exception handling.

aherciya:
https://portal.microsofticm.com/imp/v3/incidents/details/237872008/home

Adding exception handling for com.google.gson.JsonSyntaxException 

* @mentions of the person required for reviewing proposed changes (optional) @rpdome


iambmelt (changes requested):
At a glance, we're going to want similar logic in `DefaultTokenCacheStore#getItem`

rpdome (approved):
I assume this is not a fix for the given icm, but is added so that we know why the issue occurs.

aherciya:
@rpdome The crash is happening because of an unhandled error according to the latest logs, which this PR fixes, so there shouldn't be anymore crashes, but there there still needs to be investigation on why bad input to Gson.fromJson was being given.

PR #1642, opened 2021-09-11
Aherciya/enhance logging

aherciya:
### Overview
Enhancing logging for TokenCacheItem deserialization for more insight during ICM investigations. Prompted by ICM item [259533326](https://portal.microsofticm.com/imp/v3/incidents/details/259533326/similarincidents).

### Changes 
Adds a string validator to check expected attributes about the input string for a serialized TokenCacheItem and logs bad input if exists. 

### Things to consider for reviews

- Since null check was previously being ignored, it's now logged as a warning. 
- Rest of checks are logged as error since they would have thrown error previously.


changelog.txt
```diff
@@ -1,5 +1,6 @@
 v.Next
 ----------
+- [MINOR] Enhancing Logging (#1642)
```
shoatman: You're doing more than just enhancing the logging.  You are validating the value stored in the cache looks like JSON.  The Json Syntax exception we receive is telling us it's not JSON.  I'm wondering whether rather than changing how the code will potentially behave and hardcoding rules for what the JSON should look like.  ... Instead you can run your heuristics after the failure.

adal/src/main/java/com/microsoft/aad/adal/DefaultTokenCacheStore.java
```diff
@@ -169,7 +169,7 @@ public TokenCacheItem getItem(String key) {
             String json = mPrefs.getString(key);
             json = null != json ? json : "";
             String decrypted = decrypt(key, json);
-            if (decrypted != null) {
+            if (validTokenCacheItem(decrypted)) {
```
shoatman: Not just a logging enhancement.  You've added custom validation around whether this is JSON.  I think perhaps instead of validating yourself, we capture these additional hueristics after we receive the exception.

shoatman (changes requested):

adal/src/main/java/com/microsoft/aad/adal/DefaultTokenCacheStore.java
```diff
@@ -382,6 +382,30 @@ private static Calendar getTokenValidityTime() {
...
+        if(item.isEmpty()) {
+            Logger.e(TAG, "Bad input, was empty string. ");
+            return false;
+        } else if(!item.contains(":")) {
+            Logger.e(TAG, "Bad input, doesn't contain key pairs. ");
+            return false;
+        } else if(item.charAt(0) != '{') {
+            Logger.e(TAG, "Bad input, start of input is invalid. Expected opening bracket '{'. ");
+            return false;
+        } else if(item.charAt(item.length()-1) != '}') {
```
tanmaymanolkar1, line 395: Are these checks enough to validate the json

adal/src/main/java/com/microsoft/aad/adal/DefaultTokenCacheStore.java
aherciya: Yes, after after throwing the error might be better. What seemed like is probably happening is the JSON is malformed, and so I wanted to know in which way it's malformed particularly. I did this to avoid logging the the tokenCacheItem itself since it'll validate PII. Do you think logging the first+last 5/10 chars would be safe?

adal/src/main/java/com/microsoft/aad/adal/DefaultTokenCacheStore.java
aherciya: It's trying to validate the tokenCacheItem for being malformed JSON rather than 'validating JSON'. This seemed all we can do for now to avoid PII risk.

changelog.txt
aherciya: Ok I can update the description. After the failure does make sense, I was putting these conditions to gain more insight as to what specific paring failure there might be to avoid putting PII.

PR #1653, opened 2021-10-13
Add Claims to test app

rpdome:
Also add buttons to fill out extraQP (instance_aware=true) and deviceid claims.

### DIFF
--- a/userappwithbroker/src/main/java/com/microsoft/aad/adal/example/userappwithbroker/AcquireTokenFragment.java
+++ b/userappwithbroker/src/main/java/com/microsoft/aad/adal/example/userappwithbroker/AcquireTokenFragment.java
@@ -47,6 +47,9 @@ public class AcquireTokenFragment extends Fragment {
     private EditText mOtherAuthority;
     private EditText mLoginhint;
     private EditText mExtraQp;
+    private Button mExtraQpInstanceAware;
+    private EditText mClaims;
+    private Button mClaimsDeviceId;
     private Spinner mResource;
     private Spinner mPromptBehavior;
     private Spinner mClientId;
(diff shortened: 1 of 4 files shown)


userappwithbroker/src/main/java/com/microsoft/aad/adal/example/userappwithbroker/AcquireTokenFragment.java
```diff
@@ -92,6 +98,22 @@ public View onCreateView(final LayoutInflater inflater, final ViewGroup containe
         bindSpinnerChoice(mRedirectUri, Constants.RedirectUri.class);
         bindSpinnerChoice(mAssertionType, Constants.AssertionVersion.class);
 
+        mExtraQpInstanceAware.setOnClickListener(new View.OnClickListener() {
+            @Override
+            public void onClick(View v) {
+                final String str = "instance_aware=true";
+                mExtraQp.setText(str);
```
AdamBJohnsonx, line 105: Can we read-modify-write this text box?

userappwithbroker/src/main/java/com/microsoft/aad/adal/example/userappwithbroker/AcquireTokenFragment.java
```diff
@@ -92,6 +98,22 @@ public View onCreateView(final LayoutInflater inflater, final ViewGroup containe
...
+            }
+        });
+
+        mClaimsDeviceId.setOnClickListener(new View.OnClickListener() {
+            @Override
+            public void onClick(View v) {
+                final String str = "{\"access_token\":{\"deviceid\":{\"essential\":true}}}";
```
AdamBJohnsonx: Similar, to read, modify write this box, so that we add the text rather than overwrite it.  Less important for the QPs, but more important for these nasty claims objects.

AdamBJohnsonx (approved):

shahzaibj (approved):

PR #1662, opened 2021-11-04
Fix failing test by changing key type

AdamBJohnsonx:
This test started to fail at some point.  A submodule update I was involved in seemed to trigger it, and so I went looking at it because the cache seemed to be implicated.  This is true, but not complete.  The FRT changed at some point to be a multi-resource token, which has a different key.

This key is present.

However, I have no idea if this is supposed to be updated on the other key, which remains in cache.  If it is... we have work to do.

### DIFF
--- a/adal/src/androidTest/java/com/microsoft/aad/adal/AuthenticationContextTest.java
+++ b/adal/src/androidTest/java/com/microsoft/aad/adal/AuthenticationContextTest.java
@@ -23,6 +23,29 @@
 
 package com.microsoft.aad.adal;
 
+import static androidx.test.platform.app.InstrumentationRegistry.getInstrumentation;
+import static com.microsoft.aad.adal.AuthenticationConstants.UIResponse.BROWSER_CODE_AUTHENTICATION_EXCEPTION;
+import static com.microsoft.aad.adal.AuthenticationConstants.UIResponse.BROWSER_CODE_CANCEL;
+import static com.microsoft.aad.adal.AuthenticationConstants.UIResponse.BROWSER_CODE_COMPLETE;
+import static com.microsoft.aad.adal.AuthenticationConstants.UIResponse.BROWSER_CODE_ERROR;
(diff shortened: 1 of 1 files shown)


AdamBJohnsonx:
@shoatman Do you have any clarity on whether the contents of the cache are sane in this case?

shoatman:
> @shoatman Do you have any clarity on whether the contents of the cache are sane in this case?

I think we should be cautious here.  Everything from AAD is an MRRT; however, ADAL also supported an older version of ADFS as an authorization server.  ADFS as an authorization server does not understand MRRT... each RT is for a specific resource.

shoatman:
> > @shoatman Do you have any clarity on whether the contents of the cache are sane in this case?
> 
> I think we should be cautious here. Everything from AAD is an MRRT; however, ADAL also supported an older version of ADFS as an authorization server. ADFS as an authorization server does not understand MRRT... each RT is for a specific resource.

Weird... this is an FRT test though... so I'm not sure I understand what the test was doing originally... let me look closer.

shoatman:
> > > @shoatman Do you have any clarity on whether the contents of the cache are sane in this case?
> > 
> > 
> > I think we should be cautious here. Everything from AAD is an MRRT; however, ADAL also supported an older version of ADFS as an authorization server. ADFS as an authorization server does not understand MRRT... each RT is for a specific resource.
> 
> Weird... this is an FRT test though... so I'm not sure I understand what the test was doing originally... let me look closer.

Ok.  I'm guessing that originally it was not important to the test whether to look a regular RT or an MRRT because it was mocked whether the request was to ADFS or to AAD.  Whatever changed to break this test... modified the ADAL cache key to ensure that it was saved as an MRRT (which is correct... since only AAD will issue an FRT).  If you can confirm that something changed relative to creating the key... then the case will be closed.

shoatman (approved):
ADAL now uses common / ADAL-specific code for managing the cache.  Not sure why this code was moved to common as I don't think that MSAL duals-writes to ADAL, but perhaps this was enabled at some point.  @iambmelt - Do you know why the ADAL cache logic was moved to common?

PR #1674, opened 2022-01-25
Update common to latest(4.0.1)

iamgusain:
Updating common to latest version 4.0.1

### DIFF
--- a/adal/build.gradle
+++ b/adal/build.gradle
@@ -147,10 +147,10 @@ dependencies {
         transitive = true
     }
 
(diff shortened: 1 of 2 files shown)


shahzaibj:
Do we need RC and additional test passes or are we good with the validation?

shahzaibj (approved):

shahzaibj:
> Do we need RC and additional test passes or are we good with the validation?

Synced offline. We shouldn't need them.

PR #1682, opened 2022-03-23
Update maven release parameters to point to correct folders

iamgusain:
### What 
Update maven release parameters to point to correct folders

### Why
We no longer publish the artifacts in C:\temp directory.
Instead, the build artifacts go to default project directory.


p3dr0rv (approved):
LGTM

PR #1683, opened 2022-03-23
Update pipeline to point to correct path for build outputs

iamgusain:
Updating pipeline parameters to point to correct path for generated build outputs (aar, jar and pom file)


tanmaymanolkar1 (approved):
LGTM!

PR #1685, opened 2022-04-04
Remove PKeyAuth support from ADAL

rpdome:
ADAL will still declare itself as PKeyAuth capable, but it will just return a **_valid_** empty response back to the server (it will not try to load WPJ cert/key, and will always respond as such)

see https://github.com/AzureAD/microsoft-authentication-library-common-for-android/pull/1711


fadidurah (approved):
LGTM

mohitc1 (approved):

somalaya (approved):

shahzaibj (approved):

PR #1711, opened 2023-01-18
Fadi/test 1610899

fadidurah:
This PR includes an added test in ADAL to allow removal of this manual E2E test case by replacing it with a component test https://identitydivision.visualstudio.com/Engineering/_workitems/edit/1610899.

Note that the component test is not meant to cover this entire E2E test case, just the adal client side request/response handling, as the Broker behavior for this flow is already tested in msal UI automation.

### DIFF
--- a/adal/src/androidTest/java/com/microsoft/aad/adal/AuthenticationContextTest.java
+++ b/adal/src/androidTest/java/com/microsoft/aad/adal/AuthenticationContextTest.java
@@ -627,7 +627,11 @@ public final class AuthenticationContextTest {
         signal.await(CONTEXT_REQUEST_TIME_OUT, TimeUnit.MILLISECONDS);
 
         final JWT decodedToken = JWTParser.parse(callback.getAuthenticationResult().getAccessToken());
-        Assert.assertNotNull(decodedToken.getJWTClaimsSet().getClaim("deviceid"));
+        final Object deviceIdClaim = decodedToken.getJWTClaimsSet().getClaim("deviceid");
+        Assert.assertNotNull(deviceIdClaim);
+
+        // Adal gets device id from token
+        Assert.assertEquals("761b45e9-2080-4573-bebb-b43d435ce1af", (String) deviceIdClaim);
     }
 
     @Test
--- a/adal/src/androidTest/java/com/microsoft/aad/adal/AuthenticationContextTest.java
+++ b/adal/src/androidTest/java/com/microsoft/aad/adal/AuthenticationContextTest.java
@@ -52,6 +52,8 @@ import android.content.Context;
 import android.content.Intent;
 import android.content.SharedPreferences;
 import android.content.pm.PackageInfo;
+import android.content.pm.PackageManager;
+import android.content.pm.ResolveInfo;
 import android.content.pm.Signature;
 import android.os.Build;
 import android.os.Bundle;
@@ -72,6 +74,8 @@ import com.microsoft.identity.common.java.cache.SharedPreferencesAccountCredenti
 import com.microsoft.identity.common.java.dto.AccountRecord;
 import com.microsoft.identity.common.java.providers.microsoft.azureactivedirectory.AzureActiveDirectory;
 import com.microsoft.identity.common.java.providers.microsoft.azureactivedirectory.AzureActiveDirectoryCloud;
+import com.nimbusds.jwt.JWT;
+import com.nimbusds.jwt.JWTParser;
 
 import org.json.JSONException;
 import org.json.JSONObject;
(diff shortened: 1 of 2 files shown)


adal/src/androidTest/java/com/microsoft/aad/adal/AuthenticationContextTest.java
```diff
@@ -583,6 +590,46 @@ public void testPrompt() throws InterruptedException {
...
+        Mockito.when(mockedPackageManager.getPackageInfo(Mockito.anyString(), Mockito.anyInt())).thenReturn(mockPackageInfo);
+        mockContext.setMockedPackageManager(mockedPackageManager);
+
+        // Call acquire token which will try refresh token based on cache
+        context.acquireToken(testActivity.getTestActivity(), "resource", "clientid", "redirectUri",
+                TEST_IDTOKEN_UPN, PromptBehavior.Auto, null, "{\"access_token\":{\"deviceid\":{\"essential\":true}}}", callback);
+        signal.await(CONTEXT_REQUEST_TIME_OUT, TimeUnit.MILLISECONDS);
+
+        final JWT decodedToken = JWTParser.parse(callback.getAuthenticationResult().getAccessToken());
+        Assert.assertNotNull(decodedToken.getJWTClaimsSet().getClaim("deviceid"));
```
shahzaibj: Let's also add an assert for the value of the deviceid.

adal/src/androidTest/java/com/microsoft/aad/adal/AuthenticationContextTest.java
```diff
@@ -583,6 +590,50 @@ public void testPrompt() throws InterruptedException {
...
+        mockContext.setMockedPackageManager(mockedPackageManager);
+
+        // Call acquire token which will try refresh token based on cache
+        context.acquireToken(testActivity.getTestActivity(), "resource", "clientid", "redirectUri",
+                TEST_IDTOKEN_UPN, PromptBehavior.Auto, null, "{\"access_token\":{\"deviceid\":{\"essential\":true}}}", callback);
+        signal.await(CONTEXT_REQUEST_TIME_OUT, TimeUnit.MILLISECONDS);
+
+        final JWT decodedToken = JWTParser.parse(callback.getAuthenticationResult().getAccessToken());
+        final Object deviceIdClaim = decodedToken.getJWTClaimsSet().getClaim("deviceid");
+        Assert.assertNotNull(deviceIdClaim);
```
iamgusain, line 634: nit: we can compare the full token with "TEST_ACCESS_TOKEN_WITH_DEVICE_ID "

adal/src/androidTest/java/com/microsoft/aad/adal/AuthenticationContextTest.java
```diff
@@ -583,6 +590,50 @@ public void testPrompt() throws InterruptedException {
...
+
+        final PackageManager mockedPackageManager = Mockito.mock(PackageManager.class);
+        final PackageInfo mockPackageInfo = new PackageInfo();
+        mockPackageInfo.setLongVersionCode(Long.MAX_VALUE);
+        Mockito.when(mockedPackageManager.getPackageInfo(Mockito.anyString(), Mockito.anyInt())).thenReturn(mockPackageInfo);
+        mockContext.setMockedPackageManager(mockedPackageManager);
+
+        // Call acquire token which will try refresh token based on cache
+        context.acquireToken(testActivity.getTestActivity(), "resource", "clientid", "redirectUri",
+                TEST_IDTOKEN_UPN, PromptBehavior.Auto, null, "{\"access_token\":{\"deviceid\":{\"essential\":true}}}", callback);
```
iamgusain, line 626: does it matter what claim is passed, based on the mocking it is going to return the same token right, irrespective of what claims are passed.

iamgusain (approved):
LGTM

adal/src/androidTest/java/com/microsoft/aad/adal/AuthenticationContextTest.java
fadidurah: Yes, the same token will be returned. This is moreso checking that we can pass in the claim and actually complete an acquiretoken regardless of what the result we actually get is.

adal/src/androidTest/java/com/microsoft/aad/adal/MockDeviceCertProxy.java
```diff
@@ -77,12 +72,22 @@ public String getThumbPrint() {
         return sThumbPrint;
     }
 
+    @Override
+    public byte[] sign(String algorithm, byte[] dataToBeSigned) throws ClientException {
```
shahzaibj, line 76: I'm wondering why this change is being needed now

shahzaibj (approved):

adal/src/androidTest/java/com/microsoft/aad/adal/MockDeviceCertProxy.java
fadidurah: Looks like the super class was updated in common, but this class that extends it was not updated. I was getting a complitation error.

PR #1724, opened 2023-04-14
Migrate to component factory

rpdome:
Moving the usage of the old class to the new class (which we have shipped for a while)..... so that i can kill the old one.


mohitc1 (approved):

adal/src/androidTest/java/com/microsoft/aad/adal/JwsBuilderTests.java
```diff
@@ -1,102 +0,0 @@
...
-import java.security.NoSuchAlgorithmException;
-import java.security.PrivateKey;
-import java.security.PublicKey;
-import java.security.cert.CertificateEncodingException;
-import java.security.cert.X509Certificate;
-
-import static org.junit.Assert.assertTrue;
-
-@RunWith(AndroidJUnit4.class)
-public class JwsBuilderTests extends AndroidTestHelper {
```
shahzaibj, line 45: Reason for deleting this class?

shahzaibj (approved):

fadidurah (approved):

adal/src/androidTest/java/com/microsoft/aad/adal/JwsBuilderTests.java
rpdome: The original class/method is gone (IIRC due to the refactoring for PRTv3).

PR #1726, opened 2023-04-18
Mapping ECDSA to EC in ClientCertRequest's keyTypes array

melissaahn:
### Summary
Please see the summary in this common PR: https://github.com/AzureAD/microsoft-authentication-library-common-for-android/pull/2015.
The change in common needs to also be put into ADAL (which has the same on-device CBA logic in its own class).

### Testing
I initially tried to reproduce the issue with the ADAL sample app and Pixel, and I got the expected behavior. I then added the change and was able to successfully get the cert picker to pop up again. Confirmed that the cert picker was appearing for both my Pixel and Samsung devices.


adal/src/main/java/com/microsoft/aad/adal/AuthenticationActivity.java
```diff
@@ -888,13 +891,37 @@ public void alias(final String alias) {
...
         }
+
+        /**
+         * Map instances of key types without a literal reference in {@link KeyProperties} to corresponding constants in KeyProperties.
+         * @param keyTypes array of key types.
+         * @return array of mapped key types.
+         */
+        @RequiresApi(api = Build.VERSION_CODES.M)
+        @Nullable
+        public String[] mapKeyTypes(@Nullable final String[] keyTypes) {
```
rpdome: maybe you can put this code in a static method in common and use it in both places.

rpdome (approved):

adal/src/main/java/com/microsoft/aad/adal/AuthenticationActivity.java
melissaahn: Made mapKeyTypes in common static: https://github.com/AzureAD/microsoft-authentication-library-common-for-android/pull/2015/commits/19f0d0f81c9cc2f2c0218451e862e3dbcbe8c6b5

fadidurah (approved):

PR #1728, opened 2023-05-02
Revert "Consolidate IStorageSupplier (in common) (#1727)"

rpdome:


melissaahn (approved):

iamgusain:
Was there a changelog entry before, that should be removed?

iamgusain (approved):

rpdome:
> Was there a changelog entry before, that should be removed?

no changelog (since this is not really an ADAL level change.)

PR #1755, opened 2023-08-15
Target, compile SDK, AGP and gradle versions increased

somalaya:
- Target sdk version changed to 33
- Compile sdk version changed to 34
- AGP version changed to 7.4.2
- Gradle distribution url changed to 7.5

After changing above versions, I had to update other dependencies for compatibility

- Powerlift version changed to 1.0.3
- Bouncy castle version changed to 1.69
- robolectric version to 4.9
- Also, with new gradle versions, some of the properties used in gradle were deprecated (naming conventions changed). So, made those changes
- With gradle version changes, we should not be having a package name in AndroidManifest.xml.. instead, should add a namespace property in build.gradle. https://developer.android.com/build/publish-library/prep-lib-release#choose-namespace
- For apps targeting Android 14, Android restricts apps from sending implicit intents to internal app components. So , added exported attribute.  https://developer.android.com/about/versions/14/behavior-changes-14#safer-intents
- Renamed the assemble dist release aar name to contain the version number. Starting gradle version 7, gradle generates adal-dist-release.aar. But we want to publish adal-5.0.0.aar

### DIFF
--- a/adal/build.gradle
+++ b/adal/build.gradle
@@ -54,7 +54,7 @@ android {
         targetCompatibility JavaVersion.VERSION_1_8
     }
 
-    compileSdkVersion rootProject.ext.compileSdkVersion
+    compileSdk rootProject.ext.compileSdkVersion
 
     defaultConfig {
         multiDexEnabled true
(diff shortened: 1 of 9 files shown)


melissaahn (approved):

adal/build.gradle
```diff
@@ -81,7 +81,20 @@ android {
...
+            def buildFile = file("$buildDir/outputs/aar/${archivesBaseName}-dist-release.aar")
+            print "Build file $buildFile"
+            doLast {
+                println "Renaming build file $buildFile to '$buildDir${File.separator}outputs${File.separator}aar${File.separator}${archivesBaseName}-${version}.aar'"
+                if (!buildFile.renameTo("$buildDir${File.separator}outputs${File.separator}aar${File.separator}${archivesBaseName}-${version}.aar")) {
+                    println "Rename failed!"
+                }
+            }
+        }
+    }
```
iamgusain: Check if we can do this in the maven release pipeline itselft, instead of the build.gradle

iamgusain (approved):

PR #1756, opened 2023-08-18
Bump daily library versions to 1.0.X

fadidurah:
Bump dependencies that use 0.0.+ for our library versions to now use 1.0.+

### DIFF
--- a/adal/build.gradle
+++ b/adal/build.gradle
@@ -131,6 +131,7 @@ android {
     useLibrary 'android.test.mock'
 }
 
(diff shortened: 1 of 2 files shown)


adal/build.gradle
```diff
@@ -131,7 +131,7 @@ android {
     useLibrary 'android.test.mock'
 }
```
iamgusain, line 133: can you add same comment here (dev vs release branch), as in the other libs..

iamgusain (approved):

adal/build.gradle
fadidurah: Done

shahzaibj (approved):

PR #1772, opened 2023-12-29
Adding CredMan version

melissaahn:
### Summary
Please see the main common PR for details regarding Credential Manager and the passkey feature: https://github.com/AzureAD/microsoft-authentication-library-common-for-android/pull/2267
Note that the passkey feature is not going to be available for ADAL, but this PR is necessary to avoid any compilation errors when taking in the new version of common. (So the library will include CredMan by consuming common, but it has it's own set of classes for WebView and configurations, so it will not get the passkey feature from common).

### DIFF
--- a/changelog.txt
+++ b/changelog.txt
@@ -1,5 +1,6 @@
 Version Next
 -------------
+- [PATCH] Adding Credential Manager version (for compilation only) (#1772)
 
 Version 4.8.3
 -------------
--- a/changelog.txt
+++ b/changelog.txt
@@ -1,6 +1,6 @@
 Version Next
 -------------
-- [PATCH] Adding Credential Manager version (for compilation only) (#1772)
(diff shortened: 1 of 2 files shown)


shahzaibj (approved):

changelog.txt
```diff
@@ -1,5 +1,6 @@
 Version Next
 -------------
+- [PATCH] Adding Credential Manager version (for compilation only) (#1772)
```
shahzaibj: `for compilation only`

So it doesn't ship? IIUC, if common is using this then this should still make it into the final APK?

melissaahn: It does ship, but I didn't want it to seem that ADAL is actually getting this feature (ADAL uses a different set of classes for WebView). I'll change this changelog to cut the compilation part and just mention that ADAL isn't getting the feature.

iamgusain (approved):

PR #1782, opened 2024-03-14
Changes for March 2024 release

somalaya:


adal/build.gradle
```diff
@@ -132,7 +132,7 @@ android {
 }
 
 // In dev, we want to keep the dependencies(common4j, broker4j, common) to 1.0.+ to be able to be consumed by daily dev pipeline.
-String commonVersion = project.hasProperty("distCommonVersion") ? project.distCommonVersion : "1.0.+"
+String commonVersion = project.hasProperty("distCommonVersion") ? project.distCommonVersion : "17.2.0"
```
rpdome, line 135: This should remain unchanged.

rpdome (approved):

adal/build.gradle
rpdome, line 135: Should this remain unchanged in dev?

PR #1829, opened 2025-10-23
Add webkitVersion 1.14.0 to project dependencies

p3dr0rv:
Required to build adal in assemble consumers pipeline
https://github.com/AzureAD/microsoft-authentication-library-common-for-android/pull/2769

### DIFF
--- a/adal/src/main/java/com/microsoft/aad/adal/AuthenticationActivity.java
+++ b/adal/src/main/java/com/microsoft/aad/adal/AuthenticationActivity.java
@@ -660,6 +660,7 @@ public class AuthenticationActivity extends DualScreenActivity {
 
     @Override
     public void onBackPressed() {
+        super.onBackPressed();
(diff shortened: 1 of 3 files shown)


melissaahn (approved):

adal/src/main/java/com/microsoft/aad/adal/AuthenticationActivity.java
```diff
@@ -660,6 +660,7 @@ protected void onRestart() {
 
     @Override
     public void onBackPressed() {
+        super.onBackPressed();
```
mohitc1, line 663: >super.onBackPressed(); [](http://example.com/codeflow?start=8&length=22)

is this intentional?

p3dr0rv: yes, it caused lint to fail, that's why I added it

fadidurah (approved):

